PR #5, opened 2013-08-03
added urdf and meshes for the sia20d

josiahtaylor:
I realized after pushing that saying I added the meshes was rather inaccurate, it's actually an update, but it wouldn't let me change the comment.


gavanderhoorn:
If you really want to change the commit message, you can always do an interactive rebase, then a force push to your branch. This pull request will be automatically updated with the newly pushed commit.

shaun-edwards:
@josiahtaylor The dae files you submitted seem to require image files.  I think there is a better/cleaner way to handle this. The universal does not require them and it uses dae files ( https://github.com/ros-industrial/universal_robot/tree/master/ur_description/meshes/ur5/visual ).  @kphawkins has worked on the Universal robots...he may provide the solution.

Also, if the dae files will replace stl's then you will need to delete them as part of the pull requests.

shaun-edwards:
The new URDFs are different than the previous URDFs (names have changed).  All that should have changed was the file names of the solid models (from STLs to DAEs).  Too many issues to fix.

Issue #10, opened 2013-08-24
Missing install targets


gavanderhoorn:
The `CMakeLists.txt` of the `dx100` and `fs100` packages are missing install targets. This makes it impossible to use them in a install space.

gavanderhoorn:
After pull request ros-industrial/motoman#14, just the `motoman_driver` package is missing install targets.

PR #11, opened 2013-08-26
Fix issue #10: add install targets to dx100 and fs100 packages

gavanderhoorn:
Not too sure about which targets should be installed (`dx100`: should `dx100_utest` also be installed?).

Note: `fs100` package has some directory names with capitals in them.


shaun-edwards:
Waiting on this pull request until the new dx100 driver changes are made.  Once this is done, this pull request will be modified for the new install targets.

gavanderhoorn:
Just rebased (and re-did) this ontop of 734602ad. Should be good to go.

PR #16, opened 2013-09-25
Direct streaming

ghost:
These are the changes that were discussed in https://groups.google.com/forum/#!topic/swri-ros-pkg-dev/rRL9vwQ8yK0
The commit tree is a bit messy (ie "bug fix" and "typo fix") but it might be too late to rebase it...


shaun-edwards:
#88 updates this PR for `indigo`, closing

Issue #19, opened 2013-12-28
driver: hard coded Python path


gavanderhoorn:
See ros-industrial/ros_industrial_issues#9.

PR #20, opened 2013-12-28
Fix issue #19: hard coded python path in move_to_joint script

gavanderhoorn:
As per subject. Minor change. Perhaps we should explicitly specify `python2` in the shebang.


shaun-edwards:
+1

Issue #29, opened 2014-02-03
sia5d urdf doesn't have move-able joints


shaun-edwards:
The sia5d urdf doesn't have movable joints when viewed in the urdf_tutorial:

`roslaunch urdf_tutorial display.launch model:=sia5d.urdf gui:=true`

This is due to safety controller limits in the urdf (these should be removed)

```
<safety_controller k_position="0" k_velocity="0" soft_lower_limit="0" soft_upper_limit="0"/>
```

PR #30, opened 2014-02-04
Fix issue #29: remove safety controller limits in SIA5D xacro

gavanderhoorn:
Small fix.


shaun-edwards:
+1 (keeping this on the TODO list, since this needs to be merged into hydro and hyrdo-devel).

Issue #32, opened 2014-02-10
moveit_cfgs: missing run_depend on industrial_robot_simulator


gavanderhoorn:
See ros-industrial/ros_industrial_issues#23.

PR #35, opened 2014-03-16
Fix issue #32: add missing run_depend on ind_robot_simulator

gavanderhoorn:
As per subject.


gavanderhoorn:
@shaun-edwards: ping?

shaun-edwards:
+1

PR #41, opened 2014-08-15
Added error-return from streamer to action server.

hersh:
This adds the sending of an error via RobotStatus from the joint_trajectory_streamer to the action server.  This allows the action server to respond with an error immediately when there is a problem with the trajectory message.  Previously, the action server did not hear about problems with the trajectory messages, so the error would not be discovered until the action _client_ would time out, which could be several seconds.

I'm not sure how my changes fit with your idea of how this code should look.  I'm happy to modify as you like, as long as I can get that quick feedback aboud failed trajectories.

### DIFF
--- a/motoman_driver/include/motoman_driver/industrial_robot_client/joint_trajectory_streamer.h
+++ b/motoman_driver/include/motoman_driver/industrial_robot_client/joint_trajectory_streamer.h
@@ -105,6 +105,7 @@ public:
 
 protected:
 
+  void publishInvalidStart();
   void trajectoryStop();
 
   boost::thread* streaming_thread_;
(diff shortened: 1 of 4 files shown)


shaun-edwards:
@hersh,   you are way to efficient...I just sent you an email.  Let's discuss this there and then we circle back to the PR.

hersh:
Oops, I pushed a new change to the same branch.

This change ("epsilon") just adds .000001 of leeway to the speed-limit checks.  There was a problem where it was rejecting my trajectories because the limits were violated.  But they were violated by an amount too small to see with 6 digits of precision in the error message.

shaun-edwards:
Closing as this hasn't been updated in a while.  If you believe this is still needed, please feel free to resubmit.  Thanks for taking the time to contribute.

PR #43, opened 2014-08-18
Multi-group support for MotoPlus motion driver

ted-miller:
See changelog in ted-miller/motoman branch.


motoman_driver/MotoPlus/SimpleMessage.h
```diff
@@ -94,10 +102,10 @@ typedef struct					// ROS_MSG_ROBOT_STATUS = 13
 	int in_motion;				// Is currently executing a motion command:  -1=Unknown, 1=True, 0=False 
 	int mode;  					// Controller/Pendant mode: -1=Unknown, 1=Manual(TEACH), 2=Auto(PLAY)
 	int motion_possible;		// Is the controller ready to receive motion: -1=Unknown, 1=ENABLED, 0=DISABLED 
-} SmBodyRobotStatus;  	
+}  __attribute__((__packed__));
```
shaun-edwards, line 105: I haven't seen this in a long time ;).  Why was it needed?

motoman_driver/CHANGELOG.rst
```diff
@@ -2,6 +2,42 @@
 Changelog for package motoman_driver
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+1.2.3 (2014-07-02)
```
shaun-edwards, line 11: Please check that I set the appropriate version when I do a ROS release (or at least remind me to check).

motoman_driver/MotoPlus/SimpleMessage.h
```diff
@@ -195,7 +203,62 @@ typedef struct					// ROS_MSG_MOTO_MOTION_REPLY = 2012
...
+	float vel[ROS_MAX_JOINT];	// Desired joint velocities in radian/sec.  
+	float acc[ROS_MAX_JOINT];	// Desired joint accelerations in radian/sec^2.
+} __attribute__((__packed__));  	
+typedef struct _SmBodyJointTrajPtExData	SmBodyJointTrajPtExData;
+
+struct _SmBodyJointTrajPtFullEx
+{
+	int numberOfValidGroups;
+	int sequence;
+	SmBodyJointTrajPtExData	jointTrajPtData[MOT_MAX_GR];
```
shaun-edwards, line 224: Does this mean that only `MOT_MAX_GR` (i.e. 4) robots can be controlled with a single controller?  Is this a controller limitation or did you set this to limit the communications traffic?

The ROS-I REP describes these messages as being dynamic (i.e. the number of robot groups could change).  Is there a reason that you decided not to do this?  There was some concern about communications overload.  Did you see this?

shaun-edwards:
@ted-miller, Thanks for the contribution.  

There are some changes that are being marked due to differences in spacing (tabs vs spaces).  I can set the diff to ignore spacing, but the view is still messed up with some code running of the right hand side.  Is this easy for you to fix/adjust.  If so, can you do another commit (it will automatically get added to the PR).

See other comments/questions inline.

motoman_driver/MotoPlus/SimpleMessage.h
ted-miller, line 105: This guarantees consistent structure alignment on all three supported robot controllers.  The three different compilers apply padding slightly differently.  During development, I found that at least one of these structs (don't remember which) behaved differently between the FS and DX.  So, I went through and ensured everyone of them is packed without any extra padding between struct members.

motoman_driver/MotoPlus/SimpleMessage.h
ted-miller, line 224: This is a limitation of the robot controller.  The motion API can control 4 groups (of 8 axes each) on the FS100 and DX200.  This can be any combination of robots or external axes.
The DX100 is actually limited to only 2 groups.  I opted not to put a compiler switch around this definition so as to keep the message size consistent.

ted-miller:
Can you point to a specific example of a file that is showing these white-space differences?  I should be able to convert the tabs to spaces (even though it goes against every fiber of my being to not use tabs... that's almost as bad as putting the curly bracket on the same line as an IF statement)

ted-miller:
I just found some examples in Controller.h and Controller.c.

However, I would argue that the tabs should stay.  I put them there to fix some of the display issues in the MotoPlus environment (either Visual Studio or M+IDE).  Since MotoPlus doesn't allow much flexibiilty with your choice of IDE's, I think it works out well.  
Agree or disagree?

shaun-edwards:
Tabs vs spaces....wars have been started over smaller disagreements.  I agree with you.

shaun-edwards:
One additional question.  Are you concerned that the IO calls could add latency to motion streaming (since they share the same communications channel)?  This is more for my info.

ted-miller:
Yes.  I have personally seen this become a problem when exchanging the amount of I/O that is typical for an industrial work-cell.
I added a basic message in there for triggering a single I/O point like opening a gripper (I'm not concerned about that).  But for a production system, I would much rather see the ROS node use a standard I/O fieldbus technology (Ethernet/IP, Profinet, Modbus TCP) on a dedicated channel, rather than implementing new ROS-I messages just for I/O.

shaun-edwards:
+1, Waiting to merge until PR ros-industrial/industrial_core#82 is approved.

thiagodefreitas:
@ted-miller, as requested by @shaun-edwards on my PR, please change the JOINT_TRAJ_PT_FULL_EX and JOINT_FEEDBACK_EX to the MOTOMAN range.

In my opinion, the following values could be used, as you have already suggested in the past:

ROS_MSG_MOTO_JOINT_TRAJ_PT_FULL_EX with a value of 2016.
The new feedback message is ROS_MSG_MOTO_JOINT_FEEDBACK_EX with a value of 2017.

Thanks.

ted-miller:
@thiagodefreitas 
Done.  See updates in v1.2.4

PR #49, opened 2014-09-09
Executable Naming issues

thiagodefreitas:
Fixes an issue similar to https://github.com/ros-industrial/ros_industrial_issues/issues/8
Changes from version_0.x to version0 as the other naming convention gives some issues.

Tested with sda10f with normal config and with config from bmda3


shaun-edwards:
+1

motoman_driver/CMakeLists.txt
```diff
@@ -127,12 +127,12 @@ set_target_properties(motoman_motion_streaming_interface_bswap
   PROPERTIES OUTPUT_NAME motion_streaming_interface_bswap
   PREFIX "")
 
-add_executable(joint_trajectory_action
+add_executable(${PROJECT_NAME}_joint_trajectory_action
```
gavanderhoorn, line 130: All other targets in this file are prefixed with a hardcoded `motoman_`. Using `${PROJECT_NAME}_` for this target alone seems inconsistent, style-wise.

motoman_driver/CMakeLists.txt
```diff
@@ -127,12 +127,12 @@ set_target_properties(motoman_motion_streaming_interface_bswap
...
   src/joint_trajectory_action_node.cpp
   src/industrial_robot_client/joint_trajectory_action.cpp)
-target_link_libraries(joint_trajectory_action
+target_link_libraries(${PROJECT_NAME}_joint_trajectory_action
   industrial_robot_client ${catkin_LIBRARIES})
-add_dependencies(joint_trajectory_action industrial_robot_client_gencpp)
+add_dependencies(${PROJECT_NAME}_joint_trajectory_action industrial_robot_client_gencpp)
```
gavanderhoorn, line 135: The `joint_trajectory_action` target is missing a `set_target_properties(..)`, which would remove the prefix again (the recommended pattern for this). See the other targets.

This would've made the related change to `motoman_driver/launch/robot_interface_streaming.launch` unnecessary.

PR #55, opened 2014-09-18
Lower-cased sia10f.urdf stl file references, to match stl file names.

hersh:
On Linux, case matters for file-names, so this fixes a bug for me.


shaun-edwards:
+1..Thanks for the PR.

PR #58, opened 2015-03-19
v1.2.5 - Fix disconnect logic in the State Server

ted-miller:
- Fixed issue with re-connecting to the State Server after closing
  multiple concurrent connections.


gavanderhoorn:
This was already reviewed internally?

ted-miller:
Yes, this change was reviewed internally at Motoman and requires no additional changes to the ROS-I interface.

I saw Shaun was out for a while, so I went ahead and merged it since it only affects the robot side.

-Ted

gavanderhoorn:
Not sure, but I think this should then have gone into the `indigo-devel` branch?

gavanderhoorn:
Hm, looks like that is missing some commits from `hydro-devel`, so it can probably be synced some other time.

ted-miller:
I don’t know.  I forked my branch originally from the hydro-devel.  I didn’t realize there was a newer branch to work off.

Is there any easy way to merge just the MotoPlus folder into the indigo branch?  Or do I need to do a new fork and make my changes?

-Ted

gavanderhoorn:
`indigo-devel` is only missing a few commits. See [here](https://github.com/ros-industrial/motoman/compare/indigo-devel...hydro-devel) for the comparison. Syncing that is easy enough, but I'm wondering why the changes to the SIA10F urdf weren't merged into `indigo-devel` in the first place.

I'll leave this one to @shaun-edwards. Just noticed the lagging of `indigo-devel`.

shaun-edwards:
I'll make certain indigo-devel is synced.  Thanks for taking care of this
Ted.  FYI...I'm only out until the 20th of March...I fat fingered the
phone.

PR #59, opened 2015-03-30
Fixed bug in motoman_driver: external joint state messages.

hersh:
If you use any other joints with the same ROS master and publish their values on /joint_states (as one is supposed to do), the old version of the joint_trajectory_interface.cpp would erroneously record the whole joint_state message as the "current state" of the arm.  We added a gripper to our arm and published its joint states, and the arm thought its current state involved gripper joints.

This fix is not very elegant... a complete fix might check all the joints instead of just the first, or use a CurrentStateMonitor object, or even just a hash table of the joint names and values.

### DIFF
--- a/motoman_driver/src/industrial_robot_client/joint_trajectory_interface.cpp
+++ b/motoman_driver/src/industrial_robot_client/joint_trajectory_interface.cpp
@@ -85,7 +85,6 @@ bool JointTrajectoryInterface::init(SmplMsgConnection* connection)
 {
   if (this->version_0_)
   {
-    ROS_ERROR("motoman driver version 0");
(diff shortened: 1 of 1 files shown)


shaun-edwards:
@hersh, is it true that all joint states should be published on the `joint_states` topic?  This would require every node that listens on `joint_states` to perform the matching.  Seems like it adds complexity to parsing out the joint states.

We have solved this problem in the past by namespacing joint states.  If we needed a combined joint state topic on the global `joint_states`, we would use the [joint_state_publisher](http://wiki.ros.org/joint_state_publisher#Parameters) with the `source_list` parameter set appropriately.

shaun-edwards:
Won't be merged.

Issue #62, opened 2015-04-13
build fails: motoman_driver missing motoman_msgs dependencies


JeremyZoss:
Clone `motoman` repo to a bare catkin workspace, with no pre-installed `motoman` packages.  Run `catkin_make`, and build fails with the following error:

```
In file included from /home/ros-industrial/ros/src/motoman/motoman_driver/src/industrial_robot_client/joint_relay_handler.cpp:37:0:
/home/ros-industrial/ros/src/motoman/motoman_driver/include/motoman_driver/industrial_robot_client/joint_relay_handler.h:47:45: fatal error: motoman_msgs/DynamicJointsGroup.h: No such file or directory
 #include "motoman_msgs/DynamicJointsGroup.h"
                                             ^
```

Looks like the `CMakeLists.txt` in `motoman_driver` is missing some dependencies on `motoman_msgs` targets.  I confirmed that `motoman_msgs` is not being built prior to the error, and that the message header files are not being generated.

I have a fix with the required dependencies added, and will submit a pull-request shortly.

gavanderhoorn:
Fixed in #63.

PR #63, opened 2015-04-13
Fix Issue #62

JeremyZoss:
Add motoman_driver dependencies on motoman_msgs, for proper build order.

### DIFF
--- a/motoman_driver/CMakeLists.txt
+++ b/motoman_driver/CMakeLists.txt
@@ -62,6 +62,7 @@ add_library(motoman_industrial_robot_client ${CLIENT_SRC_FILES})
 target_link_libraries(motoman_industrial_robot_client
   industrial_robot_client
   industrial_utils)
+add_dependencies(motoman_industrial_robot_client ${catkin_EXPORTED_TARGETS})
 
 # Robot state node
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
Looks good.

I have noticed people preferring to add a dependency on the `_gencpp` target (of the specific msg project) if they don't really need any of the other msg generation targets to have run first.

`_EXPORTED_TARGETS` is the documented way though, so +1.

JeremyZoss:
If I understand correctly, `_gencpp` is [deprecated](http://answers.ros.org/question/60994/pkg_name_generate_messages_cpp-vs-pkg_name_gencpp/), but `_generate_messages_cpp` is still a valid alternative.  Since `_EXPORTED_TARGETS` seemed to be [recommended](https://github.com/ros/catkin/issues/547#issuecomment-29740592), that's what I used.

gavanderhoorn:
Yes it is deprecated. See also [C++ message or service dependencies](http://docs.ros.org/indigo/api/catkin/html/howto/format1/cpp_msg_dependencies.html) in the catkin docs.

shaun-edwards:
+1.

PR #64, opened 2015-05-07
Fixes a bug that was incorrectly zeroing the values for unused groups

thiagodefreitas:
Tested at UT Austin
Thanks to Andrew Sharp and Alex von Stenberg et al.


shaun-edwards:
+1...thanks for the contribution.

PR #66, opened 2015-08-03
correcting comment to match with launch files: dx100 does not use bsw…

sachinchitta:
…ap, FS100 does


shaun-edwards:
Thanks for the contribution!

shaun-edwards:
+1

PR #67, opened 2015-08-13
Reduced joint_u limits to values measured from an MH5F.

hersh:
The joint limits in the MH5 URDF are based on someone converting the joint limits stated in an [mh5 brochure](http://www.motoman.com/datasheets/MH5F.pdf) from degrees to radians, but don't relate to reality at all (that I can see).  Joint_u is the elbow and can clearly rotate less than 360 degrees, but the joint angle range specified by the joint limits in the brochure is greater than 360.  The upshot is that sometimes MoveIt tries to send the joint to impossible values and the robot stops with an error.

The values here are measured by turning joint_r (the next one) to about 45 degrees (the worst-case for interference) and moving joint_u until the robot refused to go any further.

I suppose I should do a pull request to motoman for their brochure, but I don't know where their repo is. :)

### DIFF
--- a/motoman_mh5_support/urdf/mh5_macro.xacro
+++ b/motoman_mh5_support/urdf/mh5_macro.xacro
@@ -137,6 +137,11 @@
 			<child link="${prefix}link_u"/>
 			<origin xyz="0 0 0.3100" rpy="0 0 0"/>
 			<axis xyz="0 -1 0" />
+      <!-- Old limits: lower="-2.3736" upper="4.4506" The new limits
+           below were measured from an MH5F.  They ensure that link_r
+           won't hit link_l even when joint_r is at 45 degrees.  The
+           old limits were converted directly from the MH5 data sheet,
+           which was clearly way off. -->
 			<limit lower="-1.0122" upper="3.2218" effort="100" velocity="2.96" />
 		</joint>
 		<joint name="${prefix}joint_r" type="revolute">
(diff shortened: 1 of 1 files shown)


shaun-edwards:
@hersh, sorry for the delay, I was out on vacation.  Do you know if the robot joint limits are dependent?  If you set `joint_r` to zero degrees, does the robot allow `joint_u` to further?.

hersh:
Yes, if you set joint_r to zero degrees, joint_u can go a few degrees further on the low side.  For some reason, it lets you destroy the robot on the high side.

For my purposes it is better to have something reliable than have a slightly larger workspace and sometimes hang because of an error.

The problem is that the FS100 has a different lower joint limit for joint_u depending on the value of joint_r.  The collision model loaded from the URDF, and the real robot, are not super close to colliding when the FS100 gives the error of "special interference" between U and R.  Any time we get one of those errors we have to go to the teach pendant, clear it, move the robot away from the "danger" zone, and then restart our moveit program.  So, for me, it's better to just set one limit that is far enough out to be independent of joint_r.

Like I said though, the FS100 doesn't do the same "special interference" thing on the high side of joint_u.  I went far enough to chip the paint and it didn't give any error.  I didn't test if it would go further. :)

shaun-edwards:
Ok...this makes sense to me.  Can we leave the original values in there (commented out) with a note explaining why the updated values are different?

hersh:
Sure.

shaun-edwards:
@hersh are you going to be able to make the change to comment out the original values?

hersh:
Sorry it took me a while.  There you go.

shaun-edwards:
+1... @hersh, thanks for the contribution.

PR #71, opened 2015-12-26
add support for sda5f

qqfly:
Configure the description package and moveit_cnofig based on my urdf file.


shaun-edwards:
@qqfly, thanks for the contribution.  Could you please issue this PR against the [Motoman Experimental](https://github.com/ros-industrial/motoman_experimental) repository instead.  The experimental repo is where we bring in new support/moveit config?

Also, in order to avoid redundant 3D model data (because it's large), is it possible to reuse the [SIA5](https://github.com/ros-industrial/motoman/tree/indigo-devel/motoman_sia5d_support)?

shaun-edwards:
Friendly ping?

qqfly:
@shaun-edwards , sorry for the late reply. I'll improve my packages next mounth, and contribute to the experimental repo. Besides, thanks for your work.

PR #75, opened 2016-01-10
Added error message with useful information for transition to Indigo

shaun-edwards:
This is a stop gap until documentation can be updated on the wiki.  Single arm (hydro) packages will also need to be updated in subsequent commits.

### DIFF
--- a/motoman_driver/include/motoman_driver/industrial_robot_client/motoman_utils.h
+++ b/motoman_driver/include/motoman_driver/industrial_robot_client/motoman_utils.h
@@ -1,0 +1,59 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ * Copyright (c) 2012, Southwest Research Institute
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *       * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *       * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *       * Neither the name of the Southwest Research Institute, nor the names
+ *       of its contributors may be used to endorse or promote products derived
+ *       from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef MOTOMAN_DRIVER_INDUSTRIAL_ROBOT_CLIENT_MOTOMAN_UTILS_H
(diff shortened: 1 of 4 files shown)


gavanderhoorn:
Would it be an idea to reduce duplication (and ease updating) by defining the error message text in a shared header?

shaun-edwards:
@gavanderhoorn, I could do that.  Turns out a good chuck of that code is copy and pasted (even the TODO comment :/ ).  I'll ignore the rest of the code for now, but pull out the message text.

shaun-edwards:
@gavanderhoorn, error message pulled out.  FYI, this change is temporary as I think I have figured out how to parse the Hydro style parameters in a backwards compatible way.

motoman_driver/include/motoman_driver/industrial_robot_client/motoman_utils.h
```diff
@@ -0,0 +1,59 @@
...
+/**
+  @brief Common error message displayed when node fails to find 'topic_list' parameter
+  */
+
+const std::string TOPIC_LIST_ERROR_MSG(
+    "Failed to find topic_list parameter (new in Indigo)\r\n"
+    "  \tNew parameter documentaion can be found here:\r\n"
+    "  \thttp://wiki.ros.org/motoman_driver/Tutorials/Creating%20a%20Dual-Arm%20System\r\n"
+    "If still using the Hydro server version on the controller, then set the 'version0' parameter to FALSE\r\n"
+    "  \tThe driver will assume a single arm with joint names in order of the URDF (bast to tip) OR\r\n"
```
gavanderhoorn: `basT`?

motoman_driver/include/motoman_driver/industrial_robot_client/motoman_utils.h
```diff
@@ -0,0 +1,59 @@
...
+{
+namespace motoman_utils
+{
+
+/**
+  @brief Common error message displayed when node fails to find 'topic_list' parameter
+  */
+
+const std::string TOPIC_LIST_ERROR_MSG(
+    "Failed to find topic_list parameter (new in Indigo)\r\n"
```
gavanderhoorn: Maybe `'topic_list'`.

motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
```diff
@@ -62,7 +63,12 @@ JointTrajectoryAction::JointTrajectoryAction() :
   // A parameter named topics_list contains the respective information for each of the groups
 
   std::string value;
-  ros::param::search("topics_list", value);
+  if(!ros::param::search("topics_list", value))
+  {
+
```
gavanderhoorn: Empty line.

motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
```diff
@@ -62,7 +63,12 @@ JointTrajectoryAction::JointTrajectoryAction() :
...
+    ROS_ERROR_STREAM(industrial_robot_client::motoman_utils::TOPIC_LIST_ERROR_MSG);
+    return;
```
gavanderhoorn: Do you want to `return` or actually fail the ctor? If the latter, I think throwing an exception would be more appropriate.

motoman_driver/src/industrial_robot_client/joint_trajectory_interface.cpp
```diff
@@ -97,9 +98,14 @@ bool JointTrajectoryInterface::init(SmplMsgConnection* connection)
...
     std::string value;
-    ros::param::search("topics_list", value);
+    if(!ros::param::search("topics_list", value))
+    {
+      ROS_ERROR_STREAM(industrial_robot_client::motoman_utils::TOPIC_LIST_ERROR_MSG);
+      return false;
+    }
 
     XmlRpc::XmlRpcValue topics_list_rpc;
+
```
gavanderhoorn: Why the extra empty line?

shaun-edwards:
@gavanderhoorn, I've addressed your comments...OK to merge?

motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
```diff
@@ -62,7 +63,11 @@ JointTrajectoryAction::JointTrajectoryAction() :
   // A parameter named topics_list contains the respective information for each of the groups
 
   std::string value;
-  ros::param::search("topics_list", value);
+  if(!ros::param::search("topics_list", value))
+  {
+    ROS_ERROR_STREAM(industrial_robot_client::motoman_utils::TOPIC_LIST_ERROR_MSG);
+    throw;
```
gavanderhoorn, line 69: Afaik, `throw` by itself is only allowed when re-`throw`ing an exception. That is not the case here, is it?

shaun-edwards, line 69: That is not the case.  It does work as expected, causing the node to crash with an unhandled exception.  For now, I'd like to accept this PR.  I have one in the works that pulls all this logic out and replaces it with a backwards compatible version.

gavanderhoorn, line 69: It probably works, but only because (from [here, syntax 2](http://en.cppreference.com/w/cpp/language/throw)):

> This form is only allowed when an exception is presently being handled (it calls [std::terminate](http://en.cppreference.com/w/cpp/error/terminate) if used otherwise).

I'm probably being anal here, but even though the `if` will only be taken when the parameter is not found, the exception that is then thrown (most likely a `std::terminate`) does not convey/encode that error (ie: it's actually thrown because of that `throw` without an active exception, not because of the missing parameter).

gavanderhoorn:
I'll leave the issue with the `throw` up to you. If you feel it isn't important enough, go ahead and merge.

Issue #60, opened 2015-04-03
mh5: Binary STL meshes begin with "solid" (they shouldn't)


jbohren:
Head's up, ASCII STL files are supposed to be designated by the first 5 lines reading "solid" in ASCII. The following binary STL files start with "solid" and they shouldn't. I'm pretty sure this is something that the SolidWorks STL exporter does. It should be sufficient to replace "solid" with some other five characters like "hello".

```
> grep -r -A 1 ^solid *                     (hydro-devel) 
Binary file mh5/visual/MH5_T_AXIS.stl matches
Binary file mh5/visual/MH5_L_AXIS.stl matches
Binary file mh5/visual/MH5_R_AXIS.stl matches
Binary file mh5/visual/MH5_BASE_AXIS.stl matches
Binary file mh5/visual/MH5_S_AXIS.stl matches
Binary file mh5/visual/MH5_U_AXIS.stl matches
Binary file mh5/visual/MH5_B_AXIS.stl matches
```

ref: https://en.wikipedia.org/wiki/STL_(file_format)#Binary_STL

gavanderhoorn:
Nice catch.

The other pkgs should probably also be checked.

I've used something like `sed -i 's/solid/dilos/g'` in the past for this.

PR #77, opened 2016-01-12
Fix for issue #60: remove 'solid' from binary STL headers

gavanderhoorn:
As per subject.

Having 'solid' anywhere in an STL header can thoroughly confuse loaders.


shaun-edwards:
+1...thanks for the contribution.

Issue #65, opened 2015-06-30
sia10f vs sia10d: identical robots?


gavanderhoorn:
Aren't the SIA10D and the SIA10F the same robot, just using a different controller?

Should they be merged into a `motoman_sia10_support` package, with variant specific launch files?

Additional suggestion: should all `sia` robots be merged into a `motoman_sia_support` package, which would include the `5(d|f)`, `10(d|f)` and the `20(d|f)`?

gavanderhoorn:
According to [this](http://www.motoman.se/uploads/tx_catalogrobot/Flyer_Robot_SIA_Serie_E_05.13_09.pdf) flyer, `SIA` is the _series name_, so a `motoman_sia_support` pkg would seem to make sense.

gavanderhoorn:
Same goes for the [MH series](http://docs.yaskawa.eu/Related%20robots%20pallet%20repair%20EN/files/assets/common/downloads/publication.pdf) (and [HP and UP](http://yaskawaindia.in/wp-content/uploads/2012/10/Motoman-MH-HP-UP-Series.pdf)) and the [SDA series](http://www.motoman.se/uploads/tx_catalogrobot/Flyer_Robot_SDA_Serie_E_05.2013_09.pdf).

gavanderhoorn:
This will be addressed for the SIA series by introducing a `motoman_sia_support` package (candidate pkg [here](https://github.com/gavanderhoorn/motoman_experimental/tree/refactor_sia_support/motoman_sia_support)).

PR #82, opened 2016-02-15
Refactor all SIA support pkgs into single pkg

gavanderhoorn:
This is a discussion PR. I want to gather some feedback on this before putting in too much work.

Based on the observation in #65, the proposed package layout more closely follows the idea of the [support pkg conventions](http://wiki.ros.org/Industrial/Tutorials/WorkingWithRosIndustrialRobotSupportPackages) by merging all support for robots in the SIA family into a single `motoman_sia_support` package. This gets rid of the 4 pkgs that are currently used for those robots.

Besides changes to the file and directory layout, I've also removed all `f` and `d` suffixes on all files that had them: the robots themselves do not appear to be different, and the launch files required specifying a `controller` arg anyway (which resulted in the rather confusing situation where you could `roslaunch ..._sia20d.launch controller:=fs100 ..`).

The changes shown in the diff aren't too interesting, it's probably easier to look at the package contents after a PR such as this one would've been merged: [here](https://github.com/gavanderhoorn/motoman/tree/f7d3359e42d6b0624c6d3677b6c90611bf2cf212/motoman_sia_support).

Question: would it make sense to add back the `..._siaXd` and `..._siaXf` launch files? They could just wrap the already present `..._siaX` launch file as a convenience, and provide a value for the `controller` arg based on their suffix. It would mean adding additional files, but it would also remove one extra argument that has to be propagated in 'calling' launch files.

### DIFF
--- a/motoman_sia10d_support/package.xml
+++ b/motoman_sia10d_support/package.xml
@@ -43,5 +43,10 @@
  <run_depend>joint_state_publisher</run_depend>
  <export>
   <architecture_independent/>
+  <deprecated>
+    This package will be removed in ROS Kinetic. The configuration data and
+    models included in this package can now be found in the motoman_sia_support
+    package in ROS Jade.
+  </deprecated>
  </export>
 </package>
--- a/motoman_sia10f_support/package.xml
+++ b/motoman_sia10f_support/package.xml
@@ -43,5 +43,10 @@
  <run_depend>joint_state_publisher</run_depend>
  <export>
   <architecture_independent/>
+  <deprecated>
+    This package will be removed in ROS Kinetic. The configuration data and
+    models included in this package can now be found in the motoman_sia_support
+    package in ROS Jade.
+  </deprecated>
  </export>
 </package>
--- a/motoman_sia20d_support/package.xml
+++ b/motoman_sia20d_support/package.xml
@@ -43,5 +43,10 @@
  <run_depend>joint_state_publisher</run_depend>
  <export>
   <architecture_independent/>
+  <deprecated>
+    This package will be removed in ROS Kinetic. The configuration data and
+    models included in this package can now be found in the motoman_sia_support
+    package in ROS Jade.
+  </deprecated>
  </export>
 </package>
--- a/motoman_sia5d_support/package.xml
+++ b/motoman_sia5d_support/package.xml
@@ -43,5 +43,10 @@
  <run_depend>joint_state_publisher</run_depend>
  <export>
   <architecture_independent/>
+  <deprecated>
+    This package will be removed in ROS Kinetic. The configuration data and
+    models included in this package can now be found in the motoman_sia_support
+    package in ROS Jade.
+  </deprecated>
  </export>
 </package>


gavanderhoorn:
Wrt bw compatibility: this would have to follow the _tick-tock_ approach we've been using with other packages already.

shaun-edwards:
Thanks for getting this started.  I was considering doing this as well.  My only concern is that there are several SIA models (see [here](http://www.motoman.se/uploads/tx_catalogrobot/Flyer_Robot_SIA_Serie_E_05.13_09.pdf)) and that might be a "large" package.  I had considered splitting things into `sia5`, `sia10`, etc...

Regarding your question, I think keeping `siaXf`, `siaXd` is good idea.

gavanderhoorn:
@shaun-edwards wrote:

> My only concern is that there are several SIA models (see [here](http://www.motoman.se/uploads/tx_catalogrobot/Flyer_Robot_SIA_Serie_E_05.13_09.pdf)) and that might be a "large" package. I had considered splitting things into `sia5`, `sia10`, etc...

According to that pdf there are 'only' 4 models: 5, 10, 20 & 50. The `d` and `f` are really all the same robots but with different controllers. I think we can manage 4 models in one package ([fanuc_lrmate200ic_support](https://github.com/ros-industrial/fanuc/tree/indigo-devel/fanuc_lrmate200ic_support) fi has 7).

Splitting it out would not make sense to me: that is essentially how it is now, and would just mean we'd put some launch files in the same package.

> Regarding your question, I think keeping `siaXf`, `siaXd` is good idea.

Ok. That would mean including controller-specific launch files that wrap the `siaX` files and then provide a value for the `controller` arg.

gavanderhoorn:
@shaun-edwards: some progress: this PR now only marks the various `motoman_siaX(d|f)_support` packages as deprecated in Indigo. `motoman_sia_support` is now located in `motoman_experimental` ([here](https://github.com/gavanderhoorn/motoman_experimental/tree/refactor_sia_support/motoman_sia_support) in my fork).

> > Regarding your question, I think keeping `siaXf`, `siaXd` is good idea.
> 
> Ok. That would mean including controller-specific launch files that wrap the `siaX` files and then provide a value for the `controller` arg.

I've added those, but it results in a lot of files, so I'm not sure if I like it very much.

TODO:
1. looks like some launch files are missing the `version0` argument?
2. see if any of the new `xacro` features in Indigo/Jade could be exploited to reduce the nr of launch files / xacros.

shaun-edwards:
Regarding the `version0` argument, I'm about to commit a change that removes this parameter (at least at the lower level nodes).

gavanderhoorn:
I've updated the meshes for the SIA20 in [my fork](https://github.com/gavanderhoorn/motoman_experimental) of `motoman_experimental`. This is the SIA20 now (thanks @ted-miller for the CAD of the `base_link`):

sia20_detailed

I'm not entirely happy with the blue, but it's hard to get right in RViz.

@shaun-edwards wrote:

> My only concern is that there are several SIA models (see [here](http://www.motoman.se/uploads/tx_catalogrobot/Flyer_Robot_SIA_Serie_E_05.13_09.pdf)) and that might be a "large" package. I had considered splitting things into `sia5`, `sia10`, etc...

and I think I understand your comment about package size a bit better now. Even with not too high quality detailed meshes the package is now 45MB (with the `motoman_experimental` repository at 100+MB total). The smooth geometry of the links just requires a certain nr of faces to not look like a checkerboard unfortunately. It would be nice if ROS supported `.zae` everywhere (compressed form of Collada): that would save us 80%. I expect the Debian pkgs to be around that size as well, as they use a similar form of compression (`gzip`).

gavanderhoorn:
@shaun-edwards wrote:

> Regarding the `version0` argument, I'm about to commit a change that removes this parameter (at least at the lower level nodes).

can you give me an update on this?

shaun-edwards:
@gavanderhoorn, the `version0` change is still in the pipeline, but nearly done.  I'll try and get a PR up.  I will still need help testing with dual arm hardware before I will feel comfortable merging my change.

I really like the updated models.  Thanks for taking the time to make this better.  The look of the models is really important.

shaun-edwards:
@gavanderhoorn the `version0` has been proposed in #89

shaun-edwards:
@gavanderhoorn, do we want to merge this change for Indigo without the alternative files available in Indigo?  I'm leaning towards merging in this change on the `jade` development and pulling the alternative files from the experimental repo.  Thoughts?

gavanderhoorn:
I'd rather this go into indigo - the deprecation at least. That way we can switch things in Kinetic. Otherwise we'd have to keep the deprecated packages around all of Kinetic's life (which is 2016-2021).

gavanderhoorn:
I won't have time to remodel the SIA 5 and 10, so we could just merge the current state (which does have all the models in one support package, but only has the SIA20 with the new meshes). 

Alternatively we could only merge the 'merged' pkg, leave out the mesh updates. That is probably better.

gavanderhoorn:
@gavanderhoorn wrote:

> Alternatively we could only merge the 'merged' pkg, leave out the mesh updates. That is probably better.

I've just pushed a version of the refactored pkg but without the new - more detailed - model to [refactor_sia_support_old_meshes](https://github.com/gavanderhoorn/motoman_experimental/tree/refactor_sia_support_old_meshes).

Would that - in combination with the deprecation notice in this PR - be something we could merge, @shaun-edwards?

gavanderhoorn:
I've also changed the `motoman_sia_support` launch files to work with the new (non-version0) launch files.

gavanderhoorn:
Update: I've been looking at [Nishida-Lab/motoman_project](https://github.com/Nishida-Lab/motoman_project) to see whether we could import their SIA5 model:

sia5.

shaun-edwards:
> I've just pushed a version of the refactored pkg but without the new - more detailed - model to refactor_sia_support_old_meshes.
> Would that - in combination with the deprecation notice in this PR - be something we could merge

Yes.  Is there a PR from the `refactor_sia_support_old_meshes` branch?

gavanderhoorn:
Well, this PR adds the deprecation notices.

I wanted to open a PR against `motoman_experimental` for the `motoman_sia_support` package, seeing as it's a new package.

Would you rather I submit `motoman_sia_support` here?

shaun-edwards:
From a user perspective, it's hard to switch to `motoman_sia_support` because it isn't obvious where it is.  I think the `motoman_sia_support` has to exist in the `motoman` meta-package to make it easier for users to switch.  I think this means `motoman_sia_support` must be committed to this repo.  Can you think of a different way?

gavanderhoorn:
@shaun-edwards wrote:

> From a user perspective, it's hard to switch to `motoman_sia_support` because it isn't obvious where it is.

Well, true.

> I think the `motoman_sia_support` has to exist in the `motoman` meta-package to make it easier for users to switch.  I think this means `motoman_sia_support` must be committed to this repo.  Can you think of a different way?

Well we could add some more text to the deprecation warning, something along the lines of:

"_This package will be removed in ROS Jade. The configuration data included in this package can now be found in the motoman_sia_support package, which is part of the motoman_experimental repository. Check the ROS wiki for that package for more information._"

We could optionally include a direct link (URL) to the wiki page or repository as well.

Not including `motoman_sia_support` in `motoman` does make it more work to work with the new package, but if we do include it it would mean breaking our own rules (ie: new pkgs get to soak in `_experimental`). Now `motoman_sia_support` is essentially a refactored and combined version of three packages that have been part of `motoman` for a long time, but I did have to make changes to them, so some testing is warranted imo.

shaun-edwards:
I got an idea!  We could deprecate them now (accept the PR) and remove them in `kinetic`.  We can pull the `motoman_sia_support` package over for the `jade` release (which I expect to happen very quickly after `indigo`)

gavanderhoorn:
Ok. Sounds good.

I'm a little confused, as I thought that was the plan all along, but I'm ok with your suggestion :).

gavanderhoorn:
But should we then not update the deprecation notice? If you only want to introduce `motoman_sia_support` in Jade, it'd be confusing to refer to it in the Indigo deprecation message, would it not?

shaun-edwards:
Ok...so we do need to update the notice:
`This package will be removed in ROS Kinetic. The configuration data included in this
 package will be replaced with data in the motoman_sia_support package in Jade.`
or something similar.

gavanderhoorn:
I've updated the message and rebased on-top of `indigo-devel`.

shaun-edwards:
+1...thanks for contributing.

gavanderhoorn:
So just to be clear: do we now want to merge `motoman_sia_support` into a `jade-devel` branch in this repository, or in `motoman_experimental`?

shaun-edwards:
Since it is just a refactor, I'd be OK with merging into `jade-devel`

PR #86, opened 2016-03-25
Bug fix: JointTrajectoryAction (Rebased bug fix from Feb5)

psarakisz89:
JointTrajectoryAction can now handle correctly cases where a single trajectory containing all robots (2 arms 2 torsos) of motoman is used.

Before this fix the state of the robot changed at "LOST" after some seconds and the trajectory was aborted.

### DIFF
--- a/motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
+++ b/motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
@@ -32,7 +32,6 @@
  */
 
 #include <motoman_driver/industrial_robot_client/joint_trajectory_action.h>
-#include "motoman_driver/industrial_robot_client/motoman_utils.h"
 #include <industrial_robot_client/utils.h>
 #include <industrial_utils/param_utils.h>
 #include <industrial_utils/utils.h>
@@ -55,6 +54,8 @@ JointTrajectoryAction::JointTrajectoryAction() :
 {
   ros::NodeHandle pn("~");
 
+  has_active_goal_ = false;
+
   pn.param("constraints/goal_threshold", goal_threshold_, DEFAULT_GOAL_THRESHOLD_);
 
   std::map<int, RobotGroup> robot_groups;
@@ -63,11 +64,7 @@ JointTrajectoryAction::JointTrajectoryAction() :
   // A parameter named topics_list contains the respective information for each of the groups
 
   std::string value;
-  if(!ros::param::search("topics_list", value))
-  {
-    ROS_ERROR_STREAM(industrial_robot_client::motoman_utils::TOPIC_LIST_ERROR_MSG);
-    throw;
-  }
+  ros::param::search("topics_list", value);
 
   XmlRpc::XmlRpcValue topics_list_rpc;
   ros::param::get(value, topics_list_rpc);
@@ -142,6 +139,10 @@ JointTrajectoryAction::JointTrajectoryAction() :
       boost::bind(&JointTrajectoryAction::cancelCB,
                   this, _1, group_number_int));
 
+    JointTractoryActionClient* actionClient_ = new JointTractoryActionClient(
+      node_, joint_path_action_name + "/joint_trajectory_action", false );
+
+
     pub_trajectory_command_ = node_.advertise<motoman_msgs::DynamicJointTrajectory>(
                                 joint_path_action_name + "/joint_path_command", 1);
     sub_trajectory_state_  = this->node_.subscribe<control_msgs::FollowJointTrajectoryFeedback>(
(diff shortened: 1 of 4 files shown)


motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
```diff
@@ -63,11 +64,7 @@ JointTrajectoryAction::JointTrajectoryAction() :
   // A parameter named topics_list contains the respective information for each of the groups
 
   std::string value;
-  if(!ros::param::search("topics_list", value))
```
shaun-edwards, line 66: This error message helps people when the "topics_list" parameter cannot be found.  Was it removed for a reason?

motoman_driver/src/industrial_robot_client/joint_trajectory_interface.cpp
```diff
@@ -98,11 +97,7 @@ bool JointTrajectoryInterface::init(SmplMsgConnection* connection)
     std::map<int, RobotGroup> robot_groups;
 
     std::string value;
-    if(!ros::param::search("topics_list", value))
```
shaun-edwards, line 101: See previous comment.

motoman_driver/src/industrial_robot_client/robot_state_interface.cpp
```diff
@@ -99,11 +98,7 @@ bool RobotStateInterface::init(SmplMsgConnection* connection)
     std::map<int, RobotGroup> robot_groups;
 
     std::string value;
-    if(!ros::param::search("topics_list", value))
```
shaun-edwards, line 102: See previous comment

motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
```diff
@@ -754,7 +683,187 @@ bool JointTrajectoryAction::withinGoalConstraints(
   return rtn;
 }
 
+void JointTrajectoryAction::goalCB(JointTractoryActionServer::GoalHandle & gh)
```
shaun-edwards, line 686: Was this method changed or simply moved from above?

motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
```diff
@@ -159,10 +160,20 @@ JointTrajectoryAction::JointTrajectoryAction() :
 
     this->act_servers_[group_number_int]->start();
 
+    this->act_clients_[group_number_int] = actionClient_;
```
shaun-edwards, line 163: Can you explain why action clients are needed in this node?  Originally, this node contained just servers that sent commands (via topics) to the robot driver node.  What do the client connect to?

motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
```diff
@@ -676,9 +591,23 @@ bool JointTrajectoryAction::withinGoalConstraints(
   {
     int last_point = traj.points.size() - 1;
 
+
+// bool isWithinRange(const std::vector<std::string> & lhs_keys, const std::vector<double> & lhs_values,
```
shaun-edwards, line 595: Can this commented out code be removed?

shaun-edwards:
@psarakisz89, sorry it took me so long to review.  I had a few comments (see above).  I'm also hoping you could walk me through this change.  I do not have access to a dual arm system, so it is hard for me to review these changes.  I also don't believe it's possible to "simulate" the robot (let me know if I am incorrect).

motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
psarakisz89, line 686: This method was changed. The previous code was not keeping track of the goals that were received. This was only the case when a full robot trajectory was sent (included all robot groups in one trajectory). This resulted in lost goals, meaning that successful executions were not reported and general misbehavior would occur. The code that was added is just some checking to cancel a previous goal if a new one is requested. In addition, upon completion of a goal, it is set to completed.

motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
psarakisz89, line 66: It might have been removed for debugging reasons. It is safe to bring it back.

motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
psarakisz89, line 163: This was something we were trying when debugging it is safe for it to be removed.

motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
psarakisz89, line 595: Yes it can.

psarakisz89:
@shaun-edwards, I will help you as much as I can with the change and reviewing the changes! We have a dual arm manipulator available and if needed I can test something on it. 

I guess what you are asking about the "simulation"  of the robot is, if it is possible to simulate the control box and the real robot's behavior. I guess you can try it but there is a limit on how much accuracy a simulation will have. Even with a "perfect" simulator you may miss something!

shaun-edwards:
@psarakisz89, my question regarding simulation was whether there is a way simulate the multi-group nodes below the `joint_trajectory_action`.  With the old single group nodes, this was possible.  As near as I can tell, this capability does not exist for multiple groups.

motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
shaun-edwards, line 66: Can you do this?  I think you might need to rebase this PR on the current version of `indigo-devel`.

shaun-edwards:
@psarakisz89, are you able to address the comments above?  Also, could you rebase your code.  Some recent changes to the `motoman` libraries are causing conflicts.

psarakisz89:
Hello Shaun.

I am currently away from the lab. I will address the issues as soon as possible. Sorry for the delay.

gavanderhoorn:
We're going to merge this in, but as part of #259.

There were outstanding issues with this PR, which @shaun-edwards identified, but were never fixed.

So closing this one.

Thanks @psarakisz89 for the initial PR.

PR #87, opened 2016-04-03
Add initial mh6_support.

de-vri-es:
This PR adds support for the MH6 [1]. It used the package for the MH5 as template and updated the kinematic parameters and meshes. Please take the time to scrutinize it all properly.

A few points to note:
- I set myself as the maintainer in the package.xml for now, but I wouldn't mind at all to set it to @shaun-edwards again if that is preferred (it would be for me ^^). I did leave Shaun as author in addition to myself since I copied the MH5 package as template.
- I set the version to 0.0.1 for now. If there is some policy regarding version numbers, I can amend the PR. Or you can change it later, whichever is preferred.
- I used generated convex hulls (sometimes multiple per link) for the collision meshes. I saw that the MH5 package simply uses further decimated visuals. The convex hulls are about half the size of the MH5 collision meshes, but they will give somewhat more false positives, though only in pretty risky areas. Anyway, this can be discussed.
- I used a xacro property for pi in the URDF so that the numbers can easily be checked against the datasheet (which uses degrees). I kinda prefer it this way, but this can also be discussed. Not sure on the scoping rules for these parameters with xacro, so maybe it's a bad idea.
- If I decimated the visuals too far, I still got the undecimated meshesh lying around.
- I don't have any hardware, so the only tests I ran was launching `test_mh6.launch` and playing around with the joint values.
- Maybe I completely missed/forgot something. I dunno.

[1] http://www.motoman.co.uk/en/products/robots/product-view/?tx_catalogrobot_pi1[uid]=444


de-vri-es:
Oh, and I forgot to mention:
- I changed the license from BSD to Apache v2.0. I'm fine with BSD too though.

gavanderhoorn:
Could you please resubmit this against [ros-industrial/motoman_experimental](https://github.com/ros-industrial/motoman_experimental)? Thanks.

de-vri-es:
That's a different repository. Migrating a package from one to the other will destroy the commit history? That seems a bit strange. How is that normally done?

gavanderhoorn:
This is a completely new package. What history are you referring to?

gavanderhoorn:
All new packages which have not seen significant use are first put in `_experimental` repositories. After a certain amount of time, with enough trust and / or verification, they will be moved to the non-experimental repositories, like this one.

The rationale is to give those packages a bit of soak time, and to avoid having to go through a review process for every little change before it is accepted. Also: packages in the main repository are released through the buildfarm, leading to backwards-compatibility requirements for any changes that may be necessary. Marking things as experimental avoids all that, hence the described policy.

de-vri-es:
I mean the history if it is migrated from experimental to main. I understand the rationale for having some soak time, but the commit history is valuable. Even if just to see who to talk to about a certain value.

It's also not very good etiquette to turn the work of others into a single commit "Import package from experimental repository" under the wrong name...

gavanderhoorn:
> [..] but the commit history is valuable. Even if just to see who to talk to about a certain value.

Certainly, but that is not lost. A reference is made to the originating repository, and the commit at which something is removed from the `_experimental` one can always be found. It'll take a little more effort, but is certainly doable.

> It's also not very good etiquette to turn the work of others into a single commit "Import package from experimental repository" under the wrong name...

I assume that with "under the wrong name" you mean the original author of the commits? If someone feels strongly about that, we can always set `GIT_COMMIT_AUTHOR` so that information is preserved.

And to be honest, you make it sound like we're trying to steel people's work, which is not the case: `author` tags are always kept, as well as copyright notices if any, and the commits in the `_experimental` repository will also always still mention the original author.

I'm happy to discuss how to do this in a another way (ie: preserve commits), but I don't know of a way without having to rewrite repository history, which is not something we can do.

de-vri-es:
> And to be honest, you make it sound like we're trying to steel people's work, which is not the case: author tags are always kept, as well as copyright notices if any, and the commits in the _experimental repository will also always still mention the original author.

I don't mean to imply anyone here is out to steal work here, I know that is not the case. However, it does feel unrewarding when the commit in the real repository ends up squashed under someone else's name.

> I'm happy to discuss how to do this in a another way (ie: preserve commits), but I don't know of a way without having to rewrite repository history, which is not something we can do.

Well, there is a relatively easy way. Commits from one repository can be applied to another either by manual cherry picking or using git format-patch and git-am. A new branch for the release repository can be formed from those commits, which can then be merged. I'd be willing to put in that effort if/when it would be time for a merge.

/edit: I do believe its worth the effort, even besides my otherwise bruised ego.

de-vri-es:
I'll just go an submit this PR to the experimental repository and put a note there.

gavanderhoorn:
> However, it does feel unrewarding when the commit in the real repository ends squashed under someone else's name.

Github, `gitk`, `git log` and most other tools I've worked with show the name of the original author first, not the committer?

> Well, there is a relatively easy way. Commits from one repository can be applied to another either by manual cherry picking or using git format-patch and git-am. A new branch on for the release repository can be formed from those commits, which can then be merged. I'd be willing to put in that effort if/when it would be time for a merge.

Yeah, I figured that was going to be your suggestion. It's not too much work, but does require some additional effort for setting up the migration PR. We can do that though. I'll be sure to ping you when this gets migrated.

de-vri-es:
> Github, gitk, git log and most other tools I've worked with show the name of the original author first, not the committer?

That is true, but if there are commits from others as well before it is merged, it will be pretty much impossible to set a proper committer for the squashed commit.

> Yeah, I figured that was going to be your suggestion. It's not too much work, but does require some additional effort for setting up the migration PR. We can do that though. I'll be sure to ping you when this gets migrated.

Thanks, the PR will be at motoman_experimental shortly.

PR #37, opened 2014-04-09
On-the-fly point streaming and Cartesian jog

boneil:
I have implemented on-the-fly point streaming to the controller via the joint_command topic. The approach very closely emulates the way that trajectories are currently sent to the controller. I have also included a Cartesian jogging package that uses this new functionality. There is a rogue source file called keyboard_teleop_node.cpp that is not part of any package. It is example code that does Cartesian jogging from the keyboard.

Some code might be rough around the edges with extraneous comments and the like, but I thought I'd offer it up to anyone who might be able to use it.

PR #88, opened 2016-04-22
merge branch boneil_point_streaming manually

jim-rees:
original author: Brian O'Neil

This is an update of pull request #37, "On-the-fly point streaming and Cartesian jog". I've merged it to apply to indigo-devel, and have tested it and am using it. This does not include the cartesian jog part.

### DIFF
--- a/motoman_driver/src/industrial_robot_client/joint_trajectory_interface.cpp
+++ b/motoman_driver/src/industrial_robot_client/joint_trajectory_interface.cpp
@@ -194,6 +194,8 @@ bool JointTrajectoryInterface::init(SmplMsgConnection* connection, const std::ve
   this->sub_cur_pos_ = this->node_.subscribe(
                          "joint_states", 1, &JointTrajectoryInterface::jointStateCB, this);
 
+  this->sub_joint_command_ = this->node_.subscribe("joint_command", 0, &JointTrajectoryInterface::jointCommandCB, this);
+
   return true;
 }
 
@@ -335,6 +337,12 @@ void JointTrajectoryInterface::jointTrajectoryCB(
 
   // send command messages to robot
   send_to_robot(robot_msgs);
+}
+
+void JointTrajectoryInterface::jointCommandCB(
+  const trajectory_msgs::JointTrajectoryConstPtr &msg)
+{
+  //ROS_INFO("Yessir?");
 }
 
 bool JointTrajectoryInterface::trajectory_to_msgs(
--- a/motoman_driver/src/industrial_robot_client/joint_trajectory_streamer.cpp
+++ b/motoman_driver/src/industrial_robot_client/joint_trajectory_streamer.cpp
@@ -159,6 +159,73 @@ void JointTrajectoryStreamer::jointTrajectoryCB(const trajectory_msgs::JointTraj
   send_to_robot(new_traj_msgs);
 }
 
+void JointTrajectoryStreamer::jointCommandCB(const trajectory_msgs::JointTrajectoryConstPtr &msg)
+{
+  // read current state value (should be atomic)
+  int state = this->state_;
+
+  ROS_DEBUG("Current state is: %d", state);
+
+  //If current state is idle, set to POINT_STREAMING
+  if (TransferStates::IDLE == state)
+  {
+    this->mutex_.lock();
+    this->state_ = TransferStates::POINT_STREAMING;
+    this->current_point_ = 0;
+    this->streaming_sequence = 0;
+    this->streaming_start_ = ros::Time::now();
+    this->mutex_.unlock();
+    state = TransferStates::POINT_STREAMING;
+    ROS_INFO("First joint point received. Starting on-the-fly streaming.");
+  }
+
+  //if current state is POINT_STREAMING, process incoming point.
+  if (TransferStates::POINT_STREAMING == state)
+  {
+    if (msg->points.empty())
+    {
+      ROS_INFO("Empty point received, cancelling current trajectory");
+      return;
+    }
+
+    //Else, Push point into queue
+    SimpleMessage message;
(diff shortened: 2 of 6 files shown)


shaun-edwards:
@jim-rees, thanks for this.  I'm working my way through PRs on the Motoman repo over the next few weeks.

One quick questions, does this PR support multiple joint groups (i.e. for a dual arm set up)?

jim-rees:
I have not tested with dual arm, but I don't see any reason why it shouldn't.

motoman_driver/src/industrial_robot_client/joint_trajectory_interface.cpp
```diff
@@ -337,6 +339,12 @@ void JointTrajectoryInterface::jointTrajectoryCB(
   send_to_robot(robot_msgs);
 }
 
+void JointTrajectoryInterface::jointCommandCB(
+  const trajectory_msgs::JointTrajectoryConstPtr &msg)
+{
+  //ROS_INFO("Yessir?");
```
shaun-edwards, line 345: Can this be deleted?

motoman_driver/src/joint_trajectory_streamer.cpp
```diff
@@ -401,6 +401,74 @@ void MotomanJointTrajectoryStreamer::streamingThread()
...
+
+    case TransferStates::POINT_STREAMING:
+      // if no points in queue, streaming complete, set to idle.
+      if (this->streaming_queue_.empty())
+      {
+        if (this->time_since_last < this->point_streaming_timeout)
+        {
+          time_since_last = ros::Time::now().toSec() - time_of_last;
+          ros::Duration(0.005).sleep();
+          //ROS_INFO("Time since last point: %f", time_since_last);
```
shaun-edwards, line 413: Can this log message be deleted? Or should it be of type DEBUG?

motoman_driver/src/joint_trajectory_streamer.cpp
```diff
@@ -401,6 +401,74 @@ void MotomanJointTrajectoryStreamer::streamingThread()
...
       break;
+
+    case TransferStates::POINT_STREAMING:
+      // if no points in queue, streaming complete, set to idle.
+      if (this->streaming_queue_.empty())
+      {
+        if (this->time_since_last < this->point_streaming_timeout)
+        {
+          time_since_last = ros::Time::now().toSec() - time_of_last;
+          ros::Duration(0.005).sleep();
```
shaun-edwards, line 412: Can you make the `0.005` a constant with a descriptive name?

motoman_driver/src/joint_trajectory_streamer.cpp
```diff
@@ -401,6 +401,74 @@ void MotomanJointTrajectoryStreamer::streamingThread()
...
+        }
+        else
+        {
+          ROS_INFO("Point streaming complete,  setting state to IDLE");
+          this->state_ = TransferStates::IDLE;
+          break;
+        }
+      }
+      // if not connected, reconnect.
+      if (!this->connection_->isConnected())
```
shaun-edwards, line 424: Do you know why this reconnect logic is needed here?  Why doesn't the reconnect [here](https://github.com/ros-industrial/motoman/pull/88/files#diff-a99f5449c5369b017cba549fc88643c2R329) work?

motoman_driver/src/industrial_robot_client/joint_trajectory_streamer.cpp
```diff
@@ -276,8 +343,43 @@ void JointTrajectoryStreamer::streamingThread()
...
+      ROS_INFO("I'm streaming, sir.");
+      // if no points in queue, streaming complete, set to idle.
+      if (this->streaming_queue_.empty())
+      {
+        ROS_INFO("Point streaming complete, setting state to IDLE");
+        this->state_ = TransferStates::IDLE;
+        break;
+      }
+      // if not connected, reconnect.
+      if (!this->connection_->isConnected())
```
shaun-edwards, line 357: Is this reconnection logic required?  Why doesn't [this](https://github.com/ros-industrial/motoman/pull/88/files#diff-f4bf6d1e82f3b9e996488d2548de00fcR324) logic handle it?

shaun-edwards:
@jim-rees, sorry this has taken a little while.  I'm working through a bunch of PRs on the motorman repo (in the hopes of clearing them out).  I had a few comments as well as one general question:

It's been a while since I have looked at the `industrial_robot_client` classes, but why does the state `case TransferStates::POINT_STREAMING:` come up multiple times?  Should the state machine be handled in one spot?

jim-rees:
I've had a report that this does not work with dual arm robots, because they require motoman_msgs/DynamicJointTrajectory messages which this doesn't implement.

jim-rees:
Regarding the state machines, there are two, one in MotomanJointTrajectoryStreamer::streamingThread() and one in JointTrajectoryStreamer::streamingThread(). I don't know why there are two. As far as I can tell, only the one in MotomanJointTrajectoryStreamer is actually used. I do not know why there are two classes, one overriding the other. I was hoping someone could tell me, because I find it overly confusing.

shaun-edwards:
@jim-rees, thanks for the info.  I'm working on merging some other PRs before this one.  Can you address some of my comments above?  If we can get everything resolved, I think we should merge this into a `jade` development branch (thoughts?)

curranw:
This seems like a useful feature, and based on the conversation, it seems implemented. However, the joint_command subscription in the JointTrajectoryInterface seems to be a placeholder. Is this fully implemented?

jim-rees:
JointTrajectoryInterface is not fully implemented because it is not fully used. See my comments above about the two classes. The Yessir? code is never called. I did not remove this code because I don't know why it's there.

I have not worked on this code in over a year so my memory may be a bit hazy but I will try to answer questions if anyone wants to resurrect this.

PR #90, opened 2016-05-11
Fix compilation with g++ 6

de-vri-es:
Since the motoman driver seems to contain a copy of the `industrial_robot_client`, I had to do the same here as with ros-industrial/industrial_core#143. It seems weird that this copy exists, but I'm sure there are reasons.

Additionally, g++ 6 no longer accepts initialization of non-integral static const members (such as double) in class definitions. This was never legal C++ although g++ and clang++ used to accept it. It seems g++ 6 dropped support for this in favour of `constexpr` members, which are much easier for the compiler to deal with. However, `constexpr` is a C++11 feature, so instead I moved the constants to the source file where they are actually used (in an anonymous namespace to prevent possible linking problems).

### DIFF
--- a/motoman_driver/include/motoman_driver/industrial_robot_client/joint_trajectory_action.h
+++ b/motoman_driver/include/motoman_driver/industrial_robot_client/joint_trajectory_action.h
@@ -209,7 +209,7 @@ private:
    *
    */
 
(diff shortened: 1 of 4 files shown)


shaun-edwards:
> It seems weird that this copy exists, but I'm sure there are reasons.

Agreed.

+1 to this change.  Thanks for the contribution.

Issue #92, opened 2016-05-19
driver: no tests for motoman-specific simple_message and IRC extensions


gavanderhoorn:
The `simple_message` and `industrial_robot_client` packages have a number of tests to make sure the (de)serialisation and data structure classes work as they are supposed to. While the `motoman_driver` package has introduced a number of motoman-specific extensions to that infrastructure, it did not also introduce tests for them.

Based on our experience with the tests in `simple_message`, issues like #91 are examples of problems that would most likely have been uncovered by a solid test set.

This situation should be rectified and tests for _at least_ the set of simple messages implemented in this package should be added.

PR #93, opened 2016-05-21
Add Simple Message data & msg structures for Motoman 'Single IO' extensions

gavanderhoorn:
As per subject.

C++ classes for the (Motoman) `READ_SINGLE_IO`, `READ_SINGLE_IO_REPLY`, `WRITE_SINGLE_IO` and `WRITE_SINGLE_IO_REPLY` message types.

Note that this PR contributes only the data structures; there is no way to interact with the IO services through the `motoman_driver` nodes yet.


gavanderhoorn:
I know I'm not helping #92 by not immediately including tests, but those can/will be contributed later.

shaun-edwards:
@gavanderhoorn, thanks for this.  I'm going to get some critical PRs out of the way (hopefully today) and then do a release.  After the release, I'll come back around to this PR.

gavanderhoorn:
Just a note: these classes have been extensively used the past couple of months. So although no real unit-tests are included in this PR, the large amount of runtime testing we've performed on them seems to indicate that they are alright.

Not to say that real unit-tests are sorely needed, but I already pointed that out in #92.

gavanderhoorn:
Rebased to trigger Travis.

motoman_driver/include/motoman_driver/simple_message/messages/motoman_read_single_io_message.h
```diff
@@ -0,0 +1,126 @@
...
+  // Overrides - SimpleSerialize
+  bool load(industrial::byte_array::ByteArray *buffer);
+  bool unload(industrial::byte_array::ByteArray *buffer);
+
+  unsigned int byteLength()
+  {
+    return this->cmd_.byteLength();
+  }
+
+  motoman::simple_message::io_ctrl::ReadSingleIO cmd_;
```
mmoerdijk, line 118: Should this be private?

gavanderhoorn: Yes, that might make sense.

None of the other message classes have the wrapped data structure as a private member though, so if we want to maintain consistency we should probably leave it public.

I would welcome a PR that makes all wrapped structures private though.

shaun-edwards:

PR #94, opened 2016-06-15
sia5d: use meshes in support pkg. Fix #81.

### LINKED ISSUE #81: sia5d_support: xacro macro refers to files in deprecated 'motoman_config' package
Current version of `sia5d_macro.xacro` still refers to mesh files in the `motoman_config` package ([here](https://github.com/ros-industrial/motoman/blob/6c89ebfa999aa750df9e215097563ce29f5e34db/motoman_sia5d_support/urdf/sia5d_macro.xacro#L13) fi).

This would make `sia5d_macro.xacro` stop working as soon as #76 is fixed.

gavanderhoorn:
As per subject.

This was the last xacro (at least in `motoman` itself) that used files from `motoman_config`.

### DIFF
--- a/motoman_sia5d_support/urdf/sia5d.urdf
+++ b/motoman_sia5d_support/urdf/sia5d.urdf
@@ -5,17 +5,18 @@
 <!-- =================================================================================== -->
 <!--Generates a urdf from the macro in sia5_macro.xacro -->
 <robot name="motoman_sia5d" xmlns:xacro="http://ros.org/wiki/xacro">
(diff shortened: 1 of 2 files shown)


gavanderhoorn:
I think we should merge this before an Indigo release (the `motoman_config` pkg has been deprecated since jan-2014).

gavanderhoorn:
Ping?

shaun-edwards:
+1...thanks for the contribution.  I'm pulling this in before the indigo release.

Issue #76, opened 2016-01-12
Remove motoman_config pkg before releasing into Indigo


gavanderhoorn:
has been deprecated for quite some time now already.

PR #95, opened 2016-06-15
Fix for issue #76: remove deprecated config pkg

gavanderhoorn:
This was deprecated by @shaun-edwards in 1257a94b (2014).

All the models have since been migrated to separate robot support packages.

This PR depends on changes in #94, so should only be merged after that is merged.


gavanderhoorn:
Ping?

shaun-edwards:
+1..thanks for the contributing...merging...

Issue #104, opened 2016-07-13
sia10(d|f)_support: incorrect joint limits (s, r & t)


gavanderhoorn:
The position limits for `joint_s`, `joint_r` and `joint_t` are specified as `(-)3.1416` for the SIA10 in [sia10d_macro.xacro](https://github.com/ros-industrial/motoman/blob/4f7b4c1839987b495f2395d1dbdd5557ad537fb5/motoman_sia10d_support/urdf/sia10d_macro.xacro), which is incorrect ([here](https://github.com/ros-industrial/motoman/blob/4f7b4c1839987b495f2395d1dbdd5557ad537fb5/motoman_sia10d_support/urdf/sia10d_macro.xacro#L285) for `joint_t` fi). The [spec sheet](http://www.motoman.co.uk/index.php?eID=tx_nawsecuredl&u=0&file=uploads/tx_catalogrobot/Flyer_Robot_SIA_Series_E_06.2015_43.pdf&t=1468483513&hash=74c7c80310b166b7ee5b5ce4bbc918d7d208a945) has these at (-)180 degrees, which would be exactly `pi`.

It would appear the values in the urdf are _rounded up_, while they should'be been either _floored_ or written out with more digits.

gavanderhoorn:
This is the same problem as reported in #102, but tracking this separately here for the SIA10.

gavanderhoorn:
The xacro in `sia10f_support` has the same problem ([here](https://github.com/ros-industrial/motoman/blob/6f2a85142bd7442b85a83533409f502b6f6530fd/motoman_sia10f_support/urdf/sia10f_macro.xacro)).

gavanderhoorn:
Note that this package will be removed in ROS Kinetic (and replaced with `motoman_sia_support`), but should still be fixed for Indigo.

gavanderhoorn:
Fixed in c0e5031.

PR #108, opened 2016-07-19
Fix for issue #104: fix incorrect joint limits in SIA10 urdf/xacro

gavanderhoorn:
As per subject.

d300760 fixes the rounding up of `pi`.

8702ebc makes all joint limits consistent with values given in [this spec sheet](http://www.motoman.co.uk/index.php?eID=tx_nawsecuredl&u=0&file=uploads/tx_catalogrobot/Flyer_Robot_SIA_Series_E_06.2015_43.pdf&t=1468483513&hash=74c7c80310b166b7ee5b5ce4bbc918d7d208a945). Degrees converted to radians and floats truncated after the fourth decimal digit.


shaun-edwards:
+1...thanks for the contribution @gavanderhoorn, merging.

Issue #106, opened 2016-07-18
sia20: incorrect T->tool0 rotation (pi != 3.1416)


gavanderhoorn:
The `link_t-tool0` transform defining the orientation of the `tool0` frame currently specifies only a rotation over Z of `-pi`, but it uses the value `3.1416` for it ([here](https://github.com/ros-industrial/motoman/blob/e2384f02d926da5b9716fdf915783af31962138c/motoman_sia20d_support/urdf/sia20d_macro.xacro#L188)).

The difference is enough to cause the `base->tool0` transform to be off by as much as 3 degrees wrt to the values reported by the robot.

gavanderhoorn:
Similar to #102.

gavanderhoorn:
The SIA10F (and I assume the D) has the same problem ([here](https://github.com/ros-industrial/motoman/blob/6f2a85142bd7442b85a83533409f502b6f6530fd/motoman_sia10f_support/urdf/sia10f_macro.xacro#L186)).

gavanderhoorn:
SIA10D urdf doesn't have this problem, as it doesn't have a `tool0` link.

gavanderhoorn:
Fixed in 9bf25ea.

PR #110, opened 2016-07-19
Fix for issue #106: use full-precision PI for tool0 transform

gavanderhoorn:
As per subject.


shaun-edwards:
+1 merging...thanks for the contribution @gavanderhoorn.

PR #114, opened 2016-08-04
v1.3.5 MotoROS Driver

ted-miller:
Update prevents error that would occur if the user's robot controller
has more control-groups than what is supported by MotoROS and ROS-I.


shaun-edwards:
@ted-miller, what is the behavior if the max number of control groups is violated (does it fail gracefully)?

gavanderhoorn:
Connected to @shaun-edwards' question: the `printf(..)` is only shown on the telnet console, right?

Is it possible to somehow popup a dialogue / put the controller in a (custom) error state to give more visibility to it?

ted-miller:
@shaun-edwards  Before the update, it would cause an array out of bounds error and crash the application.  Now, it simply ignores the extra groups, but continues normal operation on the first 4 groups.  I think this is the most graceful option.

@gavanderhoorn  I could throw up an alarm on the pendant during initialization to warn the user.  Maybe "WARNING: Too many groups for ROS" (limited to 32 characters).
But, if someone is not actually using a pendant (which does happen sometimes) then they will never see the alarm.  Currently, there is no method to clear an active alarm from ROS without the pendant.  And the motion commands won't work with an active alarm.

Your thoughts?

gavanderhoorn:
@ted-miller wrote:

> @gavanderhoorn  I could throw up an alarm on the pendant during initialization to warn the user.  Maybe "WARNING: Too many groups for ROS" (limited to 32 characters).
> But, if someone is not actually using a pendant (which does happen sometimes) then they will never see the alarm.  Currently, there is no method to clear an active alarm from ROS without the pendant.  And the motion commands won't work with an active alarm.

Would we consider moving just the first four groups as graceful degradation?

A motion planner may plan a (coordinated) motion of all groups that should result in collision-free motion. Executing that same motion with just the first four groups invalidates the plan completely (as the assumption that the planner controls all groups it knows the state of no longer holds), but the controller will still execute it. Worst case, links in groups that are moved collide with ignored groups. As a user, I'd rather have the controller refusing to execute the entire trajectory than running the risk of damaging my equipment.

Would it be an option to implement the following:
1. add the dialogue / custom error: users with a TP will see that, and can take appropriate action (or at least know that their configuration isn't going to work)
2. extend the `SmResultType`, `SmInvalidSubCode` and / or `SmNotReadySubcode` enums ([here](https://github.com/ros-industrial/motoman/blob/f5c7a0ddbdb967cdf18d8e78b2c5c27c9960c96f/motoman_driver/MotoPlus/SimpleMessage.h#L151-L195)) to be able to reflect the "too many groups" / "invalid controller configuration" state
3. make the controller report the error in the `error_code` field of the `STATUS` message, and set appropriate values for `in_error` and `motion_possible`
4. return `REPLY_FAILURE` for all incoming service requests as long as the controller is not correctly configured

Afaict, 3 & 4 should already automatically be handled by the current code, leaving just 1 & 2.

I think this would give the error message the highest visibility in both situations, with and without TPs.

ted-miller:
@gavanderhoorn 
You make a valid point.  I added an alarm that will display on the pendant when it truncates the number of groups.  This alarm will prevent all motion until the user goes to the pendant and reads the alarm.  At that point, they can continue operation with the first four groups.

If they attempt a motion command while this alarm is active, they will get a ROS_RESULT_NOT_READY code with subcode ROS_RESULT_NOT_READY_ALARM.  This forces them to connect a pendant and find out what is happening.  (All users will have at least one pendant for initial setup of their controllers... it just might not be connected to this particular robot during normal operation.)

gavanderhoorn:
@ted-miller wrote:

> At that point, they can continue operation with the first four groups.

I assume after they clear the error, which would mean they've hopefully understood the possible consequences.

gavanderhoorn:
Personally, I'd be in favour of an even stronger / more visible warning/error, but this seems like a good compromise.

:+1: from me.

shaun-edwards:
Thanks for the contribution....merging.

PR #118, opened 2016-08-23
driver: update error message displayed when 'port' parameter is not found

gavanderhoorn:
As per subject.

Technically 'IP ports' don't exist, but it was a strange line anyway.

It still is. These messages could use some attention to make them more user friendly.


gavanderhoorn:
Sorry. Meant to open this against `indigo-devel`.

PR #120, opened 2016-08-24
driver: fix whitespace inconsistencies in launch files.

gavanderhoorn:
As per subject.

Really minor, and probably me being OCD, but these bother me every time I see them :)

No functional changes.


gavanderhoorn:
And note: following our own internal style-guide, launch files should be indented with two spaces per indentation level, but I chose to leave that for another PR (so keeping the tabs in this one).

shaun-edwards:

Issue #105, opened 2016-07-18
sia20: incorrect B-T translation?


gavanderhoorn:
According to the [spec sheet](https://github.com/ros-industrial/motoman/files/2244663/Flyer_Robot_SIA_Series_E_08.2017.pdf) ([old link](http://www.motoman.co.uk/index.php?eID=tx_nawsecuredl&u=0&file=uploads/tx_catalogrobot/Flyer_Robot_SIA_Series_E_06.2015_43.pdf&t=1468483513&hash=74c7c80310b166b7ee5b5ce4bbc918d7d208a945)), the translation between `joint_b` and `joint_t` should be 208 mm. The `motoman_sia20d` macro currently has this at 180 mm ([here](https://github.com/ros-industrial/motoman/blob/e2384f02d926da5b9716fdf915783af31962138c/motoman_sia20d_support/urdf/sia20d_macro.xacro#L183)).

This discrepancy of 28 mm is also not fixed-up in the `link_t-tool0` transform (which is a zero-transform in the macro).

gavanderhoorn:
This will cause the transformation between a (to-be-added) `base` and `tool0` frame to be off by the same amount.

gavanderhoorn:
@shaun-edwards: you introduced the SIA20D support package in d19d6cce4fb8b1d56b6d208d818718da47db4d32, any idea where this might be coming from?

gavanderhoorn:
Same issue with the SIA10(d|f) ([here](https://github.com/ros-industrial/motoman/blob/4f7b4c1839987b495f2395d1dbdd5557ad537fb5/motoman_sia10d_support/urdf/sia10d_macro.xacro#L283)), and the SIA5D ([here](https://github.com/ros-industrial/motoman/blob/6a7a506290af111727ececa3baaee9f7afed58a7/motoman_sia5d_support/urdf/sia5d_macro.xacro#L178), although that gets an additional 11 mm in the `link_t-tool0` transform ([here](https://github.com/ros-industrial/motoman/blob/6a7a506290af111727ececa3baaee9f7afed58a7/motoman_sia5d_support/urdf/sia5d_macro.xacro#L188)), which still doesn't make it correct (134 + 11 == 145 != 148)).

gavanderhoorn:
@shaun-edwards: were all these support packages created using the SW2URDF plugin perhaps?

Issue #45, opened 2014-08-27
support: add 'base' coordinate system link to URDFs


gavanderhoorn:
As discussed in ros-industrial/ros_industrial_issues#24.

gavanderhoorn:
Related #21.

PR #121, opened 2016-08-24
Partial fix for 45: add 'base' frames to SIA series robots

gavanderhoorn:
This adds the ROS-I standard `base` frame to all robots in the SIA series we support.

Note: I've only been able to test this on the SIA20, but according to information from Motoman, this should result in the correct `base`->`tool0` transforms (but only after #105 is fixed).

### DIFF
--- a/motoman_sia10d_support/urdf/sia10d.urdf
+++ b/motoman_sia10d_support/urdf/sia10d.urdf
@@ -272,6 +272,12 @@
     <axis xyz="0 0 -1"/>
     <limit effort="100" lower="-3.1415" upper="3.1415" velocity="6.9813"/>
   </joint>
-  <!-- end of joint list -->
+
+  <link name="base"/>
+  <joint name="link_l-base" type="fixed">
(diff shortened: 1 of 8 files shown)


gavanderhoorn:
This is a partial fix, as #45 asks for `base` frames to be added to all supported models & variants, not just the SIA series.

shaun-edwards:
@gavanderhoorn, please fix merge issues.  Also, can you please re-issue this PR against `jade-devel`?

gavanderhoorn:
> @gavanderhoorn, please fix merge issues.

I'll rebase the commits.

> Also, can you please re-issue this PR against `jade-devel`?

I'd rather actually have these in Indigo. It's a pure addition, orthogonal to everything that was already there. Indigo users that don't want to use that are not impacted in any way, and neither are existing applications.

Unless existing applications have added a `base` frame themselves, of course.

Reason for getting this released (eventually) into Indigo is that it makes depending on it much easier (binary installs would also have it, so no need to do source installs).

gavanderhoorn:
@shaun-edwards: rebased.

Issue #123, opened 2016-08-24
sda10f_support: malformed STL warnings


gavanderhoorn:
Loading the SDA10F model results in the following warnings being printed on the console:

```
[ros.rviz]: The STL file 'package://motoman_sda10f_support/meshes/sda10f/visual/motoman_base.stl' is malformed. It starts with the word 'solid', indicating that it's an ASCII STL file, but it does not contain the word 'endsolid' so it is either a malformed ASCII STL file or it is actually a binary STL file. Trying to interpret it as a binary STL file instead.
[ros.rviz]: The STL file 'package://motoman_sda10f_support/meshes/sda10f/visual/motoman_axis_b1.stl' is malformed. It starts with the word 'solid', indicating that it's an ASCII STL file, but it does not contain the word 'endsolid' so it is either a malformed ASCII STL file or it is actually a binary STL file. Trying to interpret it as a binary STL file instead.
```

This is easily fixed, by changing the `solid` to a `dilos` fi using a hex editor. Alternatively loading and saving the meshes (binary encoded) with Meshlab will also work.

PR #126, opened 2016-08-29
Fix for issue #123: avoid binary<->ascii stl loader issues

gavanderhoorn:
The two updated meshes contained the string `solid` in their STL headers, which triggered a warning from the ROS stl loader.


shaun-edwards:

Issue #129, opened 2016-09-06
move_to_joint.py not included in debians


shaun-edwards:
The `move_to_joint.py` (https://github.com/ros-industrial/motoman/blob/jade-devel/motoman_driver/src/move_to_joint.py) script is not accessible from the debian build of indigo.

The install of the `CMakeLists.txt` probably needs to be updated.

shaun-edwards:
Thanks @pbeeson for reporting.

gavanderhoorn:
Would we want to include that? It's really a debug script, and those are often not included in released versions.

shaun-edwards:
I think so.  Universal has a similar script that I believe is included in the install.  

@pbeeson is using it to verify motion, and I think others would find it useful to move the robot without bringing up moveit (or having to install MoveIt for that matter).

gavanderhoorn:
Well, last time I checked, the script doesn't work (properly) in case multiple motion groups are configured. So it would seem nice if that is fixed before distributing it.

gavanderhoorn:
Can't test it right now, and it probably doesn't make sense to delay fixing this issue. Will need to investigate multi-group compatibility later.

gavanderhoorn:
I've labelled this _bug_, even though I think this was actually purposefully left out of the binary distribution.

gavanderhoorn:
Closed in 259b468ea17fa680e68a0b91f57a2b58fa99b9bb.

gavanderhoorn:
Note that this has not made it to `jade-devel` yet.

PR #130, opened 2016-09-06
Fix for issue #129: install test_move script as well

gavanderhoorn:
As per subject.


shaun-edwards:

PR #132, opened 2016-10-07
v1.3.7 MotoROS Driver

ted-miller:
Based on updates from Rosen Diankov (Mujin):

> Detection of "energy savings" mode.
> Fix support for R3 and R4 robots.
> Add function to read torque values. (Not implemented in messaging
> yet.)
> Improve detection of robot's starting position.
> Allow for multiple packets to be received in single Recv command.
> Add command for servo power.
> Add command to clear alarms/errors.


shaun-edwards:
Sorry for the long delay everyone.  I'm finally back to things.  I'm going to merge the previous PRs into "indigo" based on @gavanderhoorn's suggestion.  After those PRs, I will merge this one last.  I'm guessing we will need rebase this one after all that since it already has conflicts.  Shouldn't take more than a couple of days.  Thanks again for the help.

shaun-edwards:
@ted-miller, I attempted to resolve the conflicts in #133.  It appears the version you were working from didn't include the latest (indigo-devel), plus there were some changes (mostly white space) from @gavanderhoorn that conflicted.  If the diff looks good to you, could you please regenerate the *.out files from the version in #133?

If you want to avoid fancy "git" stuff, then you can simply download my version [here](https://github.com/shaun-edwards/motoman/archive/groovy-devel.zip) and copy over everything in the MotoPlus directory.  Commit it back to your fork and the updates should be reflected in the PR.

gavanderhoorn:
@shaun-edwards wrote:

> If you want to avoid fancy "git" stuff, then you can simply download my version [here](https://github.com/shaun-edwards/motoman/archive/groovy-devel.zip) and copy over everything in the MotoPlus directory.

The link points to a file `groovy-devel.zip`. Is that correct?

ted-miller:
Although I appreciate the attempt at dumbing it down to my level... the zip file is not correct.  It's really old.

I'll attempt the fancy "git stuff".

shaun-edwards:
Addressed by #136

PR #135, opened 2016-11-20
Bug fix: JointTrajectoryAction for Dual Arm FS100

qqfly:
JointTrajectoryAction can now handle correctly cases where a single trajectory containing all robots (2 arms 2 torsos) of motoman is used. This is only vertified on Motoman SDA5F.

Before this fix the state of the robot changed at "LOST"(with error code: Transitioning goal to LOST) after some seconds and the trajectory was aborted.
This issue was discussed in [mail-lists](https://groups.google.com/forum/#!topic/swri-ros-pkg-dev/svtmTaMwqx0) and PR #79 .

### DIFF
--- a/motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
+++ b/motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
@@ -57,6 +57,8 @@ JointTrajectoryAction::JointTrajectoryAction() :
 {
   ros::NodeHandle pn("~");
 
+  has_active_goal_ = false;
+
   pn.param("constraints/goal_threshold", goal_threshold_, DEFAULT_GOAL_THRESHOLD_);
 
   std::map<int, RobotGroup> robot_groups;
@@ -86,6 +88,7 @@ JointTrajectoryAction::JointTrajectoryAction() :
                                joint_path_action_name + "/feedback_states", 1,
                                boost::bind(&JointTrajectoryAction::controllerStateCB,
                                            this, _1, group_number_int));
+
     sub_robot_status_ = node_.subscribe(
                           "robot_status", 1, &JointTrajectoryAction::robotStatusCB, this);
 
@@ -104,6 +107,9 @@ JointTrajectoryAction::JointTrajectoryAction() :
 
   pub_trajectory_command_ = node_.advertise<motoman_msgs::DynamicJointTrajectory>(
                               "joint_path_command", 1);
+
+  sub_trajectory_state_all_  = this->node_.subscribe<control_msgs::FollowJointTrajectoryFeedback>("/feedback_states", 1,
+                               boost::bind(&JointTrajectoryAction::controllerStateCB, this, _1));
 
   this->robot_groups_ = robot_groups;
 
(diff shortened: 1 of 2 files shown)


shaun-edwards:
@qqfly thanks for the pr. I'll take a look in the next  couple of days.

qqfly:
friendly ping.

motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
```diff
@@ -532,18 +554,19 @@ void JointTrajectoryAction::controllerStateCB(
     return;
   }
 
-  if (!industrial_utils::isSimilar(all_joint_names_, msg->joint_names))
+  // joint torso_joint_b2 is not in msg->joint_names, so this code is commented out.
```
shaun-edwards, line 557: What do you mean by this comment?

motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
```diff
@@ -614,9 +637,19 @@ bool JointTrajectoryAction::withinGoalConstraints(
   {
     int last_point = traj.points.size() - 1;
 
+    std::vector<std::string> state_names;
+    std::vector<double> state_values;
+
+    // joint torso_joint_b2 is not in traj.joint_names, so the last group is ignored here.
```
shaun-edwards, line 643: Is this comment related to the issue above?

motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
```diff
@@ -105,6 +108,9 @@ JointTrajectoryAction::JointTrajectoryAction() :
   pub_trajectory_command_ = node_.advertise<motoman_msgs::DynamicJointTrajectory>(
                               "joint_path_command", 1);
 
+  sub_trajectory_state_all_  = this->node_.subscribe<control_msgs::FollowJointTrajectoryFeedback>("/feedback_states", 1,
```
shaun-edwards, line 111: Is this subscriber different than the one [here](https://github.com/ros-industrial/motoman/pull/135/files#diff-dbbdafd73ea4190550a2bc504ba53f24R87)

shaun-edwards (changes requested):
@qqfly, i've got a few questions on the change you proposed (see review comments).

motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
qqfly: subscriber **_sub_trajectory_state__** listen to single group feedback, for example **_*/arm_left/feedback_states_** and **_*/arm_right/feedback_states_**;
but when we execute a multi-groups motion, such as group "arms" ( = "arm_left" + "arm_right"), we should listen to  **_/feedback_states_** for all joints.
The origin code ignores feedback of multi-groups motion, and the robot state "LOST".
In this PR, I create a new subscriber **_sub_trajectory_state_all__** to fix this problem.

motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
qqfly: There are four groups (not MoveGroup) for SDA dual arm robot (arm_left, arm_right, torso_1, torso_2), and torso_1 and torso_2 share a same joint (torso_joint).
for **_all_joint_names__**, there are 16 joints (7 left + 7 right + torso_joint_b1 + torso_joint_b2)
while **_msg->joint_names_** only have 15 joints (7 left + 7 right + torso_joint_b1), torso_joint_b2 is missing.
so **_industrial_utils::isSimilar()_** always return false in this scene.

motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
qqfly: Yes, this is the same with the issue above.
we only use the first 3 groups value, and ignore the last torso_joint_b2 (**last_trajectory_state_map_.size()** ).

motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
gavanderhoorn: @qqfly: the driver is also used for non-dual-arm robots. This logic will not work there, as the 'shared joint' is a peculiarity of how dual-arm robots are configured.

motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
gavanderhoorn: See my earlier comment: ignoring the last group is not a good idea on systems that actually do have 4 groups, but do not have the special joint<->group mapping that the SDA series has.

motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
qqfly: @gavanderhoorn , for single-arm robots, this function _JointTrajectoryAction::controllerStateCB(
  const control_msgs::FollowJointTrajectoryFeedbackConstPtr &msg)_ will not be called, it will turn to _JointTrajectoryAction::controllerStateCB(
  const control_msgs::FollowJointTrajectoryFeedbackConstPtr &msg, **int robot_id**)_
I have cout this two kind of joint_names when connecting with my SDA5F, there is no torso_joint_b2 in msg->joint_names. That's why I comment out this joint name checking.

gavanderhoorn: Single arm, but multi-group robots are also targeted by this driver.

qqfly: Got it.
But for SDA-series robot (4-groups), this problem is still there, (all_joint_names_ isn't consistent with _msg->joint_names), and the joint_name checking or goal-constraint checking fails. The same with the issue below.
Maybe we can only do this for SDA-series robots?

qqfly: Got it.
But for SDA-series robot (4-groups), this problem is still there, (all_joint_names_ isn't consistent with _msg->joint_names), and the joint_name checking or goal-constraint checking fails. The same with the issue below.
Maybe we can only do this for SDA-series robots?

shaun-edwards: Sorry for the delay on my part, but I'm in agreement with @gavanderhoorn.  Any solution should be independent of the robot type.  I can't think of a bulletproof way of knowing the robot type.  I don't fully understand why the message doesn't have `torso_joint_b2`.  Is the answer to just fix the `msg->joint_names` publisher?

qqfly: yes, it would be much better. I will firstly vertify this bug (publisher for dual-arm robots), after a couple of days.

shaun-edwards:
Closing this issue...if this is still an active PR, please feel free to comment with more information and we will reopen it.

Issue #146, opened 2017-01-09
driver: document potential problems with line-endings in Inform jobs


gavanderhoorn:
As commented by @ted-miller in https://github.com/ros-industrial/motoman/issues/142#issuecomment-271280846:

> However, the most common reason for the error: *ERROR 3200 : NOP or END instruction not found* 
is because of the line-endings.
> 
> The robot controller requires Windows-style line endings: <carriage return><line feed>
> 
> If you open the job in a Linux text editor (like gedit), it can change the line endings to eliminate the <carriage return>.  This will cause the Error 3200.

This should be documented somewhere, probably the installation tutorial, or a *Troubleshooting* page (to be created, #144).

gavanderhoorn:
Perhaps we should provide MotoROS distributions on the wiki, packaging the application binaries with the various Inform jobs and have users download and install those instead of cloning the github repository.

gavanderhoorn:
Documenting is nice, but it's probably easier if we let the VCS deal with this. See #147.

ted-miller:
I agree with automating it.  But, I've had some people try to modify the job in GEDIT before.  So I plan to also put this information on the wiki.

gavanderhoorn:
#147 just tries to make sure that Inform jobs don't get checked out with their line-endings already converted to Unix. If that does happen, editing with Gedit or not doesn't matter any more.

Good thing to document it though. [This page](https://ubuntugenius.wordpress.com/2010/10/26/creating-windows-text-files-in-ubuntu-how-to-save-as-ms-dos-end-of-line-in-gedit/) seems to show a way to keep Windows line-endings in the file when saving.

gavanderhoorn:
The newly added *Troubleshooting* page contains a sub section that covers the issue with line-endings ([here](http://wiki.ros.org/motoman_driver/Troubleshooting#ERROR_3200:_NOP_or_END_instruction_not_found)). Additionally, #146 was merged, which forces git to always use Windows line-endings when cloning the repository.

I feel this sufficiently addresses the reported issue.

Issue #142, opened 2017-01-08
installation: INFORM Code in DX200 ERROR!


BlazingForests:
Hi, guys

I am following this wiki http://wiki.ros.org/motoman_driver/Tutorials/indigo/InstallServer

I already installed the MotoROS in DX200, looks no problem.

When i installing the INFORM Code in DX200, ERROR!

DX200 show this ERROR

```
L:****) ERROR 3200 : NOP or END instruction not found (J:INIT_ROS
```

My DX200's version is 

```
DN2.05.00A-00
```

Does anyone know how to solve it?

THX

gavanderhoorn:
Just to make sure: you did use the file from [motoman_driver/Inform/DX200/single_arm](https://github.com/ros-industrial/motoman/tree/jade-devel/motoman_driver/Inform/DX200/single_arm), correct?

BlazingForests:
Hi, @gavanderhoorn Yes, I am sure.

BlazingForests:
Even i modify the INIT_ROS, still ERROR

```
NOP
DOUT OT#(890) OFF
DOUT OT#(889) OFF
TIMER T=0.05
DOUT OT#(889) ON
WAIT OT#(890)=ON
DOUT OT#(890) OFF
END
```

gavanderhoorn:
I'm pretty sure you can't just remove the header of jobs and expect things to work.

Perhaps an incompatibility was introduced recently, as `INIT_ROS.JBI` has been like that for a long time.

@ted-miller ?

BlazingForests:
I think i fixed it.

I directly create INIT_ROS on pendant, It's OK, no problem.
Then i save this job to USB drive. As follows

```
/JOB
//NAME INIT_ROS
//POS
///NPOS 0,0,0,0,0,0
//INST
///DATE 2017/01/09 13:10
///ATTR SC,RW
///GROUP1 RB1
NOP
DOUT OT#(890) OFF
DOUT OT#(889) OFF
TIMER T=0.05
DOUT OT#(889) ON
WAIT OT#(890)=ON
DOUT OT#(890) OFF
END
```

THX

ted-miller:
gavanderhoorn is correct that you cannot just remove the job header.  This is required when loading a job into the robot controller.

However, the most common reason for the error:
"_ERROR 3200 : NOP or END instruction not found_" is because of the line-endings.

The robot controller requires Windows-style line endings: <carriage return><line feed>

[...]

PR #147, opened 2017-01-09
Fix for #146: force CRLF line-endings for Inform Jobs

gavanderhoorn:
This should avoid confusion such as reported in #142, by forcing git to always checkout the `*.jbi` (and `*.dat`) files with `<CR><LF>` line-endings (ie: Windows), even on platforms which normally use Unix line-endings.


ted-miller (approved):

gavanderhoorn:
Thanks for the review.

PR #150, opened 2017-03-01
v1.4.1 - MotoROS Driver (jade-devel)

ted-miller:
Removed extra task creation when connecting to the motion server which already has reached the maximum number of connection.
Removed unneeded parameters from setup validation library.


motoman_driver/MotoPlus/MotionServer.c
```diff
@@ -944,9 +977,41 @@ BOOL Ros_MotionServer_StartTrajMode(Controller* controller)
...
+			}
+			else
+			{
+				char errMsg[ERROR_MSG_MAX_SIZE];
+				memset(errMsg, 0x00, ERROR_MSG_MAX_SIZE);
+				Ros_Controller_ErrNo_ToString(rData.err_no, errMsg, ERROR_MSG_MAX_SIZE);
+				printf("Can't disable energy-savings mode because: %s\r\n", errMsg);
+				goto updateStatus;
+			}
+		}
```
gavanderhoorn: Could this perhaps be refactored into a function?

I'm not sure, but it looks like a copy of the same lines earlier in the commit?

ted-miller: ok

ted-miller:
A lot of small things hit all at once.  This should be the last of my updates.

motoman_driver/MotoPlus/MotionServer.c
gavanderhoorn: :)

I actually meant the entire piece of code starting with 

```c++
if (servoOnOff == ON && Ros_Controller_IsEcoMode(controller) == TRUE)
```

basically the "try to get servos out of ECO mode" sequence (with the retries). Not just the error printing bit :)

but that is also nice to put into a separate function.

ted-miller: That part actually does have slightly different logic.  It's look at a different signal and also checking for a different state of that signal.  So... I think a function would just over complicate it.

ted-miller: Oh wait... ignore my last comment.  You are correct.  Give me a minute to fix it up.

gavanderhoorn: I'd already started diffing :)

ted-miller:
@julienking 
We think this version should work with the MPL80.  Eric did some testing on the DX200.  This version will detect if the robot is this special case and will modify the motion commands to cancel out the automatic movement.

motoman_driver/MotoPlus/MotionServer.c
```diff
@@ -770,35 +775,16 @@ BOOL Ros_MotionServer_ServoPower(Controller* controller, int servoOnOff)
 	MP_SERVO_POWER_SEND_DATA sServoData;
 	MP_STD_RSP_DATA stdRespData;
 	int ret;
+	STATUS status;
 	
 	if (servoOnOff == OFF)
 		Ros_MotionServer_StopMotion(controller);
 
-	if (servoOnOff == ON && Ros_Controller_IsEcoMode(controller) == TRUE)
+	if (servoOnOff == ON)
```
gavanderhoorn: <del>(perhaps just for my understanding): is there a reason `Ros_Controller_IsServoOn(controller)` is not used here?</del>

wait, I didn't have enough context displayed. Ignore that.

ted-miller: Yes, it's checking the requested command rather than the actual state.

ted-miller:
@shaun-edwards 
Friendly ping

pbeeson (commented):
Tested on MH180, dx200 controller.  Works great.

gavanderhoorn (approved):
Will test soon on our SIA20, but @pbeeson reports success, so merging this now.

gavanderhoorn:
Thanks for all the work again @ted-miller.

PR #151, opened 2017-03-01
v1.4.1 - MotoROS Driver (indigo-devel)

ted-miller:
Removed extra task creation when connecting to the motion server which
already has reached the maximum number of connection.
Removed unneeded parameters from setup validation library.


gavanderhoorn:
Thanks for the additional PR, but we could probably sync `indigo-devel` with `jade-devel` just as easily (through the web-ui).

Are you ok with closing this one?

ted-miller:
Yeah, they're identical.

PR #153, opened 2017-03-07
Added a disable/endable service that stops the industrial_robot_client from accepting any incoming goals 

pbeeson:
For safety, it is sometimes desirable to have a "software level e-stop". This essentially does that for ROS-I robot by adding a service that "pauses" or "unpauses" the robot by allowing goals to come through. If the robot is paused, any current goal is canceled via the cancelCB() function, stopping the robot.

Discussed here: ros-industrial/motoman#152

PR #155, opened 2017-03-15
Fixed unsafe behavior at driver level

pbeeson:
The current Motoman driver tries to autostart the robot anytime it sees the ROS-INIT job isn't already running.  This even happens after an Estop.  While in theory the client should be listening to /robot_status and it shouldn't send commands, a dumb client could continue sending commands and the driver would auto restart the robot once an estop was cleared.  This adds a /motoman_enable service that must be called on startup and on estop in order to restart the ROS_INIT job on the Inform teach pendant. There is also a /motoman_disable that stops the job when the Motoplus is in the IDLE state.


shaun-edwards:
@pbeeson, thanks for this.  I'm planning to work on this tomorrow and Friday.  It will take me a little bit go through the issue discussion.  I had one question, but let me make sure you haven't addressed it in the issue before asking it here.

pbeeson:
No problem Shaun. I can attest that both the Motoman and the
industrial-robot-client enable stuff works on real robot testing. The
latter is less important though useful for stopping a robot to examine
things during debug (and allows a robot independent way to stop the robot
with ROS-I robots). The Motoman driver fix was needed for our customer to
sign off on any further development after they saw the previous behavior,
but can be tweaked if needed.

shaun-edwards:
I'm hesitant to approve both this PR and #153 / ros-industrial/industrial_core#174.  Having multiple services that do similar things (and have similar names) is certain to cause confusion.  I believe that this PR approach is better since it disables the driver at the lowest ROS level possible.  

Have you looked at getting this behavior into the generic `industrial_robot_client` too?  Just wondering if you have and if you think it's possible (not asking you to do this)?

pbeeson:
@shaun-edwards I'm not sure what you are asking about 
> Have you looked at getting this behavior into the generic industrial_robot_client too?

I personally see `https://github.com/ros-industrial/industrial_core/pull/174` as this same behavior, except in the industrial_robot_client. 

On the more general topic of redundancy: I agree that I would have liked to have only written this into one location.  The original industrial_core PR was because I didn't want to have to write this enable/disable into the simulator if such behavior was supported on, but different for, every ROS-I robot.  But of course, I didn't plan on the motoman driver not having this functionality already built-in somehow.

The main difference right now between the two is that the industrial_robot_client PR works for all ROS-I robots (including the industrial_sim robot) AND stops any current commanded motion.  The motoman version will disable on request, but I didn't add the extra call to stop any motion, so the disable only affects new commands, not running commands---though that's trivial to add stop_motion to the disable call.  Mainly, this #155 was expected mostly to be to manually re-enable after estop or alarming instead of auto-enabling after these events.  If you don't want to accept the industrial_robot_client PR, that's fine, but we should add stop_motion request to this PR and add a similar stop_motion and enable/disable service calls to the industrial_sim so it then works like the motoman driver -- and you'd probably want to have similar functionality in all your robot drivers (this is what I was trying to avoid by adding it to the client level to begin with).  

In any case, this PR needs to be accepted, as my Motoman integrators (and now my customer once they heard about this) won't move forward without it.  So this probably affects all future Motoman applications.  We can continue to discuss changing #153 and it's robot_client mirror to avoid confusion, but that shouldn't preclude this low-level safety change.  I can also add in that stop_motion to the disable service call if desired.

shaun-edwards:
I'm all for accepting this PR over the other two.  For this PR, I'd like to suggest the following
 - Add a "stop" command on disable, as you suggest
 - Rename the services to "enable" or "robot_enable" etc...  Ideally it we would support these same names in the `industrial_robot_client` (not in the action).  I think I can sign up to adding this functionality
 - Add some very helpful log messages that point people to the new services when a new trajectory arrives and the robot is disabled.  Since this is a change of functionality, it will cause confusion.
 - Is this change required in `indigo`?

pbeeson:
>     Add a "stop" command on disable, as you suggest

Added.

>     Rename the services to "enable" or "robot_enable" etc... Ideally it we would support these same names in the industrial_robot_client (not in the action). I think I can sign up to adding this functionality

Changed to /robot_enable and /robot_disable from /motoman_enable and /motoman_disable

>     Add some very helpful log messages that point people to the new services when a new trajectory arrives and the robot is disabled. Since this is a change of functionality, it will cause confusion.

This was already there, but updated slightly.

>     Is this change required in indigo?
> 

Yes, it is needed in Indigo.  Additionally, indigo-devel works on Kinetic systems just fine.

shaun-edwards (approved):

shaun-edwards:
Thanks for the contribution....merging.

PR #156, opened 2017-03-16
Sync indigo-devel with jade-devel

gavanderhoorn:
As per subject.

Already reviewed.


gavanderhoorn:
I'll wait for Travis, then immediately merge: this is a sync PR.

shaun-edwards (approved):

Issue #159, opened 2017-03-21
'fixing unsafe behavior at driver level' broke the build


gavanderhoorn:
See [Project Ibin_uS64__motoman_driver__ubuntu_saucy_amd64__binary/8](http://build.ros.org/view/Ibin_uS64/job/Ibin_uS64__motoman_driver__ubuntu_saucy_amd64__binary/8/consoleFull) (via [Preparing for Indigo Sync 2017-03-20](https://discourse.ros.org/t/preparing-for-indigo-sync-2017-03-20/1530)):

```
00:21:39 In file included from /tmp/binarydeb/ros-indigo-motoman-driver-0.3.6/src/joint_streaming_node.cpp:32:0:
00:21:39 /tmp/binarydeb/ros-indigo-motoman-driver-0.3.6/include/motoman_driver/joint_trajectory_streamer.h:42:30: fatal error: std_srvs/Trigger.h: No such file or directory
00:21:39  #include "std_srvs/Trigger.h"
00:21:39                               ^
00:21:39 compilation terminated.
```

Seems to be caused by a missing dependency declaration in the `motoman_driver` pkg.

gavanderhoorn:
This was merged in #155.

Strange thing is that Travis seems to have succesfully built everything before it was merged, so not sure what happened.

PR #160, opened 2017-03-21
Fix for issue #159: add missing dependency on std_srvs

gavanderhoorn:
As per subject.

No functional changes.

Let's see what Travis thinks.

### DIFF
--- a/motoman_driver/CMakeLists.txt
+++ b/motoman_driver/CMakeLists.txt
@@ -2,7 +2,7 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(motoman_driver)
 # Load catkin and all dependencies required for this package
-find_package(catkin REQUIRED COMPONENTS std_msgs roscpp actionlib
+find_package(catkin REQUIRED COMPONENTS std_msgs std_srvs roscpp actionlib
     trajectory_msgs actionlib_msgs control_msgs sensor_msgs urdf
     industrial_msgs industrial_robot_client industrial_utils simple_message motoman_msgs roslint)
 find_package(Boost REQUIRED COMPONENTS system thread)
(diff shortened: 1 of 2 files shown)


gavanderhoorn:
00b898c and 1dc51b3 are extra and just make `catkin_lint` happier.

shaun-edwards:
Looks like Travis couldn't find the branch...perhaps just a hiccup.  Restarted.

shaun-edwards:
@gavanderhoorn, this is some unbelievable bad luck.  We make a fix, Travis passes, but the build farm fails....all before a sync.  The PR, which fixes the issue, now fails on Travis in a way that I have never seen before.  Do you have any insight into why Travis is failing on just the push?  Perhaps its something `industrial_ci` did?

gavanderhoorn:
> Do you have any insight into why Travis is failing on just the push? Perhaps its something `industrial_ci` did?

yes: I should not have -- by accident -- pushed the same branch to both my personal fork *and* the main repository. Travis somehow linked the duplicated branch (on `motoman`) to this PR, I deleted the branch when I discovered my mistake, but Travis was still trying to build it.

The `travis-ci/pr` build passed, so that is the important one.

I've just pushed the duplicated branch to `motoman` again so we can let Travis complete its build. Afterwards we'll only merge the one in the PR.

That should fix / work-around this.

shaun-edwards:
That explains...no big deal.  Thanks for reacting so quickly.  I would have missed the item on Discourse.

shaun-edwards (approved):
LGTM

shaun-edwards:
I'll merge and then do the release tonight.

gavanderhoorn:
This will need to be syned with `indigo-devel` as well ..

PR #164, opened 2017-04-05
Removed jade from allowed failures list

shaun-edwards:


shaun-edwards:
This PR and #165 seem to be failing due to a roslaunch-test failure in jade.  See [here](https://travis-ci.org/ros-industrial/motoman/jobs/218703345#L579-L584) for more details.  I will get to this tomorrow.

.travis.yml
```diff
@@ -17,10 +17,9 @@ env:
     - USE_DEB=true  ROS_DISTRO="jade"   ROS_REPOSITORY_PATH=http://packages.ros.org/ros-shadow-fixed/ubuntu
 matrix:
   allow_failures:
-    - env: USE_DEB=true  ROS_DISTRO="jade"   ROS_REPOSITORY_PATH=http://packages.ros.org/ros/ubuntu
-    - env: USE_DEB=true  ROS_DISTRO="jade"   ROS_REPOSITORY_PATH=http://packages.ros.org/ros-shadow-fixed/ubuntu
+    - env: USE_DEB=true  ROS_DISTRO="indigo"   ROS_REPOSITORY_PATH=http://packages.ros.org/ros/ubuntu
+    - env: USE_DEB=true  ROS_DISTRO="indigo"   ROS_REPOSITORY_PATH=http://packages.ros.org/ros-shadow-fixed/ubuntu
```
gavanderhoorn, line 21: Same comment as in #165.

gavanderhoorn:
The `roslaunch` test is failing because `sda10f.xacro` tries to instantiate the `motoman_sda10f` macro ([here](https://github.com/ros-industrial/motoman/blob/b645a083a469ec1e0c0819fa3f93e36d3f9feccd/motoman_sda10f_support/urdf/sda10f.xacro#L5)), but that doesn't exist. The macro is actually called `sda10f` ([here](https://github.com/ros-industrial/motoman/blob/5c58be32132205e2e62b39858fde8330bcbdcf4a/motoman_sda10f_support/urdf/sda10f_macro.xacro#L3)).

I'm not sure why this is suddenly a problem. That file (`sda10f.xacro`) hasn't changed since it was originally committed.

.travis.yml
shaun-edwards: same reply ;)

shaun-edwards:
Note, prefixing does not actually work due to issues with the xacros (fixed in #169).  This change keeps everything the way it was for jade-devel (just fixes the test).  I plan to apply the fix to `kinetic-devel` in order to avoid any possibility of introducing a breaking change into jade (which has been around for a while).

shaun-edwards:
BTW, I did verify that this is not an issue in Indigo.  Apparently xacro naming was not a checked as closely?

shaun-edwards:
@Levi-Armstrong, can you approve this fix as well.  It's close to PR #169, but is a partial fix for jade.  It addresses the roslaunch failures, but not the prefix issues.  I chose this partial fix so as to limit the chances of breaking something in jade (even thought it's technically still broken if you want to use prefixing).

Levi-Armstrong (approved):

Levi-Armstrong:
:+1:  OK to merge.

PR #165, opened 2017-04-05
Updated travis build script for kinetic

shaun-edwards:


.travis.yml
```diff
@@ -15,12 +15,15 @@ env:
     - USE_DEB=true  ROS_DISTRO="indigo" ROS_REPOSITORY_PATH=http://packages.ros.org/ros-shadow-fixed/ubuntu
     - USE_DEB=true  ROS_DISTRO="jade"   ROS_REPOSITORY_PATH=http://packages.ros.org/ros/ubuntu
     - USE_DEB=true  ROS_DISTRO="jade"   ROS_REPOSITORY_PATH=http://packages.ros.org/ros-shadow-fixed/ubuntu
+    - USE_DEB=true  ROS_DISTRO="kinetic"   ROS_REPOSITORY_PATH=http://packages.ros.org/ros/ubuntu
+    - USE_DEB=true  ROS_DISTRO="kinetic"   ROS_REPOSITORY_PATH=http://packages.ros.org/ros-shadow-fixed/ubuntu
 matrix:
   allow_failures:
+    - env: USE_DEB=true  ROS_DISTRO="indigo"   ROS_REPOSITORY_PATH=http://packages.ros.org/ros/ubuntu
+    - env: USE_DEB=true  ROS_DISTRO="indigo"   ROS_REPOSITORY_PATH=http://packages.ros.org/ros-shadow-fixed/ubuntu
```
gavanderhoorn, line 23: Can you clarify why we would want to allow failures on Indigo?

shaun-edwards: This change is to support kinetic development (I updated the base branch BTW).  I don't want to support backports of the code, so I disabled previous versions.  I left the old versions in the build list since it is nice to know if the code still builds.  However, since our testing is not broad enough, we won't be able to guarantee that a backport works.

gavanderhoorn: Personally I'd like it if we could try to support both Indigo and Kinetic. They're both LTS releases, and the differences -- especially in these pkgs -- are not that big. Dropping support for Indigo does not make sense to me.

Issue #166, opened 2017-04-05
sda10f: incorrect macro usage in top-level xacro


gavanderhoorn:
As reported in #165 and #164: `sda10f.xacro` is trying to instantiate a macro called `motoman_sda10f`, but that doesn't exist, so `roslaunch` tests for that top-level xacro fail.

Travis log ([here](https://travis-ci.org/ros-industrial/motoman/jobs/218703345#L579-L584)):
```
  /usr/bin/cmake -E make_directory /root/catkin_ws/build/motoman_sda10f_support/test_results/motoman_sda10f_support
  /opt/ros/jade/share/roslaunch/cmake/../scripts/roslaunch-check -o /root/catkin_ws/build/motoman_sda10f_support/test_results/motoman_sda10f_support/roslaunch-check_test_launch_test.xml.xml /root/catkin_ws/src/motoman/motoman_sda10f_support/test/launch_test.xml 
checking /root/catkin_ws/src/motoman/motoman_sda10f_support/test/launch_test.xml
xacro: Traditional processing is deprecated. Switch to --inorder processing!
To check for compatibility of your document, use option --check-order.
For more infos, see http://wiki.ros.org/xacro#Processing_Order
unknown macro name: xacro:motoman_sda10f
```

gavanderhoorn:
Comment in #164:

> The `roslaunch` test is failing because `sda10f.xacro` tries to instantiate the `motoman_sda10f` macro ([here](https://github.com/ros-industrial/motoman/blob/b645a083a469ec1e0c0819fa3f93e36d3f9feccd/motoman_sda10f_support/urdf/sda10f.xacro#L5)), but that doesn't exist. The macro is actually called `sda10f` ([here](https://github.com/ros-industrial/motoman/blob/5c58be32132205e2e62b39858fde8330bcbdcf4a/motoman_sda10f_support/urdf/sda10f_macro.xacro#L3)).
> 
> I'm not sure why this is suddenly a problem. That file (`sda10f.xacro`) hasn't changed since it was originally committed.

PR #167, opened 2017-04-05
Added MPL80 robot model to motoman package

EricMarcil:
It includes:
motoman_mpl80_moveit_config
motoman_mpl80_support


JeremyZoss:
I haven't reviewed this PR in detail, but had a few concerns:
  1. The collision meshes look fairly detailed.  Consider simplifying to improve performance.  We often will create Convex Hull models using MeshLab (see [here](http://wiki.ros.org/Industrial/Tutorials/Create%20a%20URDF%20for%20an%20Industrial%20Robot#Collision_quality_STLs)).
  1. This is a 5-DOF robot, with a MoveIt package set to use KDL.  Have you tested this in MoveIt?  In the past, KDL has had difficulty with < 6DOF solutions, and custom (e.g. ikfast) kinematics modules were required for proper planning support.
  1. The launch files reference "dx200", but this is not currently a supported/valid controller config.  It's only referenced in comments, and you plan to add support soon, so I don't think this issue needs any actual fixes.  Just pointing it out in case others have a different opinion.

gavanderhoorn:
re: kinematics: @JeremyZoss is correct. KDL is not known to work very well with 5 dof robots. I've had similar problems with our M-430iA/2F.

re: dx200: as far as I know that controller is fully supported by motoros? There just isn't a convenience launch file in `motoman_driver`. Is that what you are referring to @JeremyZoss?

re: support pkg: this may not have been completely clear, but support packages are actually supposed to support robot *series*. I'm not completely familiar with this model, but would it be correct to say it belongs in the `mpl` series of robots? If that is the case, then the MPL80 is essentially a single variant. It would be better to rename the support package to `motoman_mpl_support`, and add the `80` as one of the supported variants. Other variants could then be added later, without requiring any additional packages.

Finally: as these are new packages, I think it would be best if we add them to [motoman_experimental](https://github.com/ros-industrial/motoman_experimental) first. That way they can soak a bit. After some time, they could then be migrated to `motoman` proper.

gavanderhoorn:
But 👍 for the PR, and 🍻 for the contribution!

JeremyZoss:
re: dx200 : yes, my concern was purely with the missing convenience launch files.  If you try to specify "dx200" at the command line, as listed in the launch-file example arguments, it will currently fail.  But I expect that to be rectified soon (I think Eric's planning to add the dx200 launch files in a separate PR), so no big concern.

re: package naming and repo : Eric's current naming matches the other current motoman packages (for what it's worth).  @shaun-edwards will probably need to weigh in with a final decision on naming and repo.

EricMarcil:
Hi Jeremy,


1.       You are correct the collision files are the same as the visual.  I didn’t have tool to simplify the STL file and didn’t take the time to manually make one from basic geometry.  I’ll check out the MeshLab.  Thanks

2.       You are correct it is a 5 DOF.  I did test it just grabbing the robot end and keeping the flange parallel to the ground I didn’t have problems.  I used MoveIt to test recent modification in the MotoRos driver or later to support these type of robots.  I should probably also have mentioned that it requires the  MotoRos driver v1.4.1 or later.

3.       Current, if you have a “dx200” controller, you can connect to it referring to “dx100” instead.  I have the modification for the “dx200” done, it’s basically copies of the “dx100” files renamed with “dx200”.  We also have the “YRC1000” controller coming out soon which is also the same.  In reality, it is only the FS100 that is different.  So,  I’m not sure if we want to keep adding the various controller type, or maybe we should just define a “default” and “fs100”.  I’ll probably do a separate pull request for that shortly.

EricMarcil:
FYI,

The MPL series are palletizing robots that always maintain the flange of the robot parallel with the floor.  The larger robots of that series only have 4 axis and the flange orientation is maintained mechanically through back links.  The MPL80 actually has 5 axis, the flange orientation is maintained using a B-axis motor that is slave to the L and U axis.  So the compensation is automatically done through software.  So it is different from other MPL robots.

But generally speaking, other models could be grouped in categories.  I just follow the pattern of existing models.  I’m assuming that generic model would require more parameters to account for robot size and the different link models…  It might be something to consider?  What is the general feeling?  Are their guideline or examples on making those?

JeremyZoss:
@gavanderhoorn 's comment was not suggesting that we create generic/parameterized files (launch, URDF, etc.).  That will get messy quickly, for anything other than trivial changes.  But, rather, it is possible to group several related robots into a single "support" package while still maintaining uniquely-named launch files, URDF files, STL mesh files, etc.  See for example [fanuc_m430ia_support](https://github.com/ros-industrial/fanuc/tree/indigo-devel/fanuc_m430ia_support).

This is purely an organizational preference, to keep from exploding the number of ROS packages.  But we're trying to standardize on a common approach, so it's easy for users to know how to find/organize their ROS-I models.

Note that some packages (esp. MoveIt) cannot be combined in this manner, and still generally require a separate package for each distinct robot variant.

shaun-edwards:
@EricMarcil, thanks for the contribution.  Please don't wait for my feedback.  @JeremyZoss and @gavanderhoorn's review is as good as mine.  I recently broke the build (see #166), and in addition to buying everyone donuts ;)...I have to fix it before we can merge your changes (if you notice the "All checks have failed", that was me).  I'll work on this over the weekend.  Thanks again for your efforts.

gavanderhoorn:
@JeremyZoss wrote:
> re: package naming and repo : Eric's current naming matches the other current motoman packages (for what it's worth). @shaun-edwards will probably need to weigh in with a final decision on naming and repo.

completely true, and we (I) should finally take / find the time to document this somewhere, but as you wrote earlier, we're trying to converge to the situation where *all* support pkgs follow this layout, so I'm always hesitant to accept new pkgs that don't.

> This is purely an organizational preference, to keep from exploding the number of ROS packages. But we're trying to standardize on a common approach, so it's easy for users to know how to find/organize their ROS-I models.

That is exactly the reason for all of this.

@EricMarcil: perhaps [Working with ROS-Industrial Robot Support Packages](http://wiki.ros.org/Industrial/Tutorials/WorkingWithRosIndustrialRobotSupportPackages) could provide you with some more descriptive info instead of just an example. As for a motoman example, [motoman_sia_support](https://github.com/gavanderhoorn/motoman_experimental/tree/refactor_sia_support/motoman_sia_support) tries to do something similar for the SIA series (but it uses the same ideas as `fanuc_m430ia_support` that @JeremyZoss mentioned).

gavanderhoorn:
@EricMarcil wrote:
> The MPL series are palletizing robots that always maintain the flange of the robot parallel with the floor. The larger robots of that series only have 4 axis and the flange orientation is maintained mechanically through back links. The MPL80 actually has 5 axis, the flange orientation is maintained using a B-axis motor that is slave to the L and U axis. So the compensation is automatically done through software. So it is different from other MPL robots. But generally speaking, other models could be grouped in categories.

I think the main discriminator here should be whether the manufacturer / vendor puts robots in the same series. If Motoman does not consider the MPL80 as being in the same series as the rest of the MPL robots, then they should not be put in the same robot support packages.

From my lay-mans perspective though, I would expect the MPL80 -- together with the 100, 160, 300, 500 and 800 -- to be in a `motoman_mpl_support` package. Not only does that avoid introducing 6 packages for what are basically variants, but it also allows us to reuse as much as possible things like common settings, parts of xacro macros and -- most importantly -- meshes (the MPLs seem to have identical / similar base meshes, no?).

@EricMarcil: I'm willing to do the changes if you'd prefer.

gavanderhoorn:
@gavanderhoorn wrote:
> Finally: as these are new packages, I think it would be best if we add them to [motoman_experimental](https://github.com/ros-industrial/motoman_experimental) first. That way they can soak a bit. After some time, they could then be migrated to `motoman` proper.

@JeremyZoss, @shaun-edwards: opinion? Our policy has been to put new packages in the experimental repositories. My preference would be to do that here as well I think.

JeremyZoss:
re: `motoman_experimental` repo : I concur.  This has been our traditional (though unofficial) approach.

I'm typically less worried about things like support and moveit packages, since those are relatively easy to verify and pose little "risk" if released in the main repo.  Things like driver or algorithm packages (e.g. planners) feel like they sit in the `*_experimental` repos for a while to go through more extensive testing before merging into the main repo.  I don't feel like there's as much need for testing on support and moveit packages, as they are more standardized.

So, I could go either way.  I'm fine with leaving these packages directly in the main repo, but I agree that releasing them first in `motoman_experimental` is more consistent with our typical process.

EricMarcil:
I just did a new pull request #168 that adds the launch files for the DX200 and YRC1000 controllers.  So that should address Jeremy's first comment.

Concerning the MoveIt package using the KDL motion planner is there a better one to use for a 5 DOF arm?  
My objective in creating the MoveIt package was simply to test the recent changes to the MotoRos driver to support MPL80 robot and the model.  I got that done.  I didn't test any obstacle avoidance or anything fancy.  Just moving from point A to point B.  I figured that since I had that working, I may as well check that in along the motoman_mpl80_support.  But if we are not sure about the proper planner, is it better to pull it out and only commit the motoman_mpl80_support?

JeremyZoss:
If you've tested basic motion planning using this moveit package, I'm fine to commit as-is.  We can "upgrade" the kinematics plugin later, if problems arise.

In some cases, the default KDL kinematics solver does not produce good results.  MoveIt either fails to plan, or you get crazy results.  We don't want to check in a "non-functional" MoveIt package, but I'm fine to commit a "sub-optimal" package.

For higher-performance or more specialized kinematics needs, we will typically create a custom kinematics plugin for a specific manipulator configuration using the ikfast tool.  This process is out of scope for a pull request discussion, and probably not needed in this case (see above comments).  If you'd like to read more, see the discussion [here](http://answers.ros.org/question/152591/kdl-kinematics-solver-limitations/) and tutorials [here](http://docs.ros.org/jade/api/moveit_ikfast/html/doc/ikfast_tutorial.html) or [here](http://wiki.ros.org/Industrial/Tutorials/Create_a_Fast_IK_Solution).  We can also discuss further offline, if you'd like.

gavanderhoorn:
The MoveIt package is ok wrt KDL being used. I just tested it. It does include a `B10deg` named state which should be removed.

@EricMarcil: is the location of the `base` coordinate system in the xacro macro correct for this type of robot? It's linked with an all-zeros transform to `base_link`.

gavanderhoorn:
@ros-industrial/motoman-push: I suggest we squash merge this PR if/when it's accepted.

EricMarcil:
The MPL80 does have a functional B-axis.  It needs to be controlled by ROS so that it stays within a 15 degrees angle perpendicular to the ground, otherwise the controller will generate an alarm.

As for the location of the base coordinate system, I’m not sure what is the standard in ROS.  I’m still new at ROS and all the various tool, but from what I’m figuring I believe that the current model is expressing the tool position relative to the S-axis (1st axis) rotation axis at ground level (under the robot base).  I had used the MPL5 as a reference when I made the MPL80.

The Motoman standard is that the robot coordinate system is at the S-axis rotation axis but at the height of the L-axis (2nd axis) rotation center.  Which gives a 540 mm difference between the two references.  So is it better to use the manufacturer coordinate system so that numbers on in ROS match numbers on the manufacturer teach pendant?

So I’m not sure which reference is the best one to use.

EricMarcil:
This is my first contribution and I’m still new with GitHub.  Was the way I did the merge correct?
What is a “squash merge”?

gavanderhoorn:
@EricMarcil wrote:
> The MPL80 does have a functional B-axis.  It needs to be controlled by ROS so that it stays within a 15 degrees angle perpendicular to the ground, otherwise the controller will generate an alarm.

Hm. That is a restriction that I'm not certain of how we would express / encode that in the URDF. I'm wondering whether a mimic joint would work for that, but perhaps a specialised IK plugin would be better.

What I was referring to with "[the moveit config] does include a `B10deg` named state which should be removed" was the MoveIt *named state* that is configured in the MoveIt configuration. There are two: a `home` named state, and a `B10deg` state. We try to standardise these a bit, and typically only include an `all-zeros` named state. If I understand you correctly, I don't believe the `B10deg` has any special meaning, correct? If so, it would probably be better to remove it.

> As for the location of the base coordinate system, I’m not sure what is the standard in ROS.  I’m still new at ROS and all the various tool, but from what I’m figuring I believe that the current model is expressing the tool position relative to the S-axis (1st axis) rotation axis at ground level (under the robot base).  I had used the MPL5 as a reference when I made the MPL80.
> 
> The Motoman standard is that the robot coordinate system is at the S-axis rotation axis but at the height of the L-axis (2nd axis) rotation center.  Which gives a 540 mm difference between the two references.  So is it better to use the manufacturer coordinate system so that numbers on in ROS match numbers on the manufacturer teach pendant?
> 
> So I’m not sure which reference is the best one to use.

The `base` link/frame (note, not `base_link`) should coincide with the mfgs *world* (or similar) frame, precisely to make it possible to check `base->tool0` transforms against what the Teach Pendant shows as Cartesian coordinates of the TCP with an all-zeros toolframe.

So in this case I think the `parent` of the [base_link-base](https://github.com/EricMarcil/motoman/blob/766ecf3303e7b322ec6ba69f5657ae27efa0c5f1/motoman_mpl_support/urdf/mpl80_macro.xacro#L173-L178) joint should probably be a link that "is at the S-axis rotation axis but at the height of the L-axis (2nd axis) rotation center".

I think we can do that if we make `joint_s` the `parent` of the `base_link-base` joint, and give it an `x` offset of `0.145` (relative to `joint_s`).

gavanderhoorn:
@EricMarcil wrote:
> Was the way I did the merge correct?

yes, perfect.

> What is a “squash merge”?

that is a merge of a PR on github (but really two git branches) that does not preserve complete history in that it merges all commits into a single one (typically the first).

I don't believe the commits I contributed are really important for the history of the repository -- the fact that the MPL80 was added is more important. So I'm suggesting we squash merge my stuff and yours into a single one which shows just your initial commit ("Added MPL80 robot model to motoman package").

gavanderhoorn:
@gavanderhoorn wrote:
> I think we can do that if we make `joint_s` the `parent` of the `base_link-base` joint, and give it an `x` offset of `0.145` (relative to `joint_s`).

And that joint should then probably no longer be called `base_link-base`.

EricMarcil:
Yes, I agree for the "squash merge" once we've got everything ironed out.


@gavanderhoorn wrote:

> What I was referring to with "[the moveit config] does include a B10deg named state 

Yes, I forgot about that.  I created for testing purpose.  It should be removed.

For the B-axis restriction, I don't see how it can be captured in the URDF.  It would required that ability to dynamically adjust the B-axis limit as the L and U axis changes.  It think that it is something that will need to be addressed in a specialized IK for palletizing that keeps the robot flange parallel to the ground.  We added the support at the demand of a user that want to do palletizing with ROS.  Hopefully they can contribute once they are done.

For the base link, if I understand the ROS standard properly, I would suggest the following changes:
- Change the name of the model from "base_link" to "link_base".  This will follow the standard for all other models starting with "link_"
- Define a link called "base", which is the Motoman (manufacturer) standard coordinate system and start the chain of links from there.
- Define the "base_link" relative to the "base" at -540 mm.

I can push the changes shortly.

gavanderhoorn:
@EricMarcil wrote:
> For the base link, if I understand the ROS standard properly, I would suggest the following changes:
> - Change the name of the model from "base_link" to "link_base".  This will follow the standard for all other models starting with "link_"

`base_link` is actually a de facto standard in ROS for the first link in a URDF model, so I suggest we keep it. Lots of tools will expect that link to be the first one in the chain.

> - Define a link called "base", which is the Motoman (manufacturer) standard coordinate system and start the chain of links from there.
> - Define the "base_link" relative to the "base" at -540 mm.

We sort-of borrowed the name `base` for the mfg coordinate system in ROS-I. See [wiki/tutorials/Create a URDF for an Industrial Robot - Additional/Standard Frames](http://wiki.ros.org/Industrial/Tutorials/Create%20a%20URDF%20for%20an%20Industrial%20Robot#Additional.2BAC8-Standard_Frames), so if we could stick to that convention that would be great.

I think the only thing we need to do is use the current `base` link, but just anchor it at the right place. Right now its `parent` is `base_link`, that would need to change to `joint_s`, and as we cannot re-use `joint_l` (as that frame also rotates over the y-axis), we just put the `x` `origin` of `base` at +`0.145`. This would basically put it at the same position as `joint_l`, but while decoupling it from the rotation of that joint.

--- 

The reason we introduced `base` and `tool0` is to be able to do mfg-agnostic Cartesian transforms across all supported robot models. `base` works quite well, but `tool0` should probably be split in a `flange` and `tool0`. The former is then the mechanical (and logical, ie: in the urdf) attachment point for EEFs and the latter would *always* represent an all-zeros toolframe. Such a split would allow the two frames to better capture the semantics of the frames, as right now `tool0` ends up being used by ppl as a flange, which either leads to them updating the `fixed` transform (which should not be done) or defining frames like `tool1`, `tool2`, etc (which is also not as it was intended to be used).

(this comment is not directly related to your PR, but I thought I'd mention it, as it might clarify some things).

EricMarcil:
I think that you've misundertood my earlier description of the Motoman base position.
The XY is that of the S-axis (1st joint) and the Z is that of the L-axis (2nd joint).  The difference is 540 mm in Z and the X-axis should not be changed.
Looking at wiki/tutorials/Create a URDF for an Industrial Robot - Additional/Standard Frames reference that you gave me, I think we can leave everything as it originally was and just add 540 mm to the Z in the base_link-base joint.  That will bring the "base" frame to the same position as the one define on the Motoman controller.

gavanderhoorn:
You're right, I did misunderstand you. I think d10d673c is fine, indeed.

shaun-edwards:
@EricMarcil, I think this is good to go now.  Thanks so much for your effort and patience.  The question now becomes whether we pull this into the main repo or the experimental repo.  The experimental repo is where we put things that need time for testing.  Once a few people sign off on the experimental version, we then move it to the official (this) repo.  How confident are you that everything is correct?

EricMarcil:
**@Shaun Edward**, I agree, I think we are good to go.
**@gavanderhoorn, @JeremyZoss**, Thanks for the review and suggestions.
I'm fairly confident with the model.  I've actually tested it with a real controller.  My biggest concern putting it in the experimental repo is that I don't expect this model to be very popular (unless palletizing interest with ROS grows) so it may take a very long time before another user make use of it.  I'm not sure how many people think of looking in the experimental repo...   
But I'll let you make the final decision.

gavanderhoorn:
I suggest we address the visibility concerns by making sure that the ROS buildfarm doc job also indexes the experimental repository. That way we get the normal ROS wiki visibility for those packages (I do that with [fanuc_experimental](http://wiki.ros.org/fanuc_experimental) as well).

gavanderhoorn:
Travis is currently failing because of #165 and #166.

@shaun-edwards: is anything else needed from @EricMarcil before we merge this?

shaun-edwards:
My plan is to manually merge this into `motoman_experimental`.  Plan to get to it today.

gavanderhoorn:
Ok.

I've submitted ros/rosdistro#14839.

shaun-edwards:
Packages have been merged into `motoman_experimental` (ros-industrial/motoman_experimental#13).  Thanks @EricMarcil for the contribution.  I'd suggest you make an announcement on the developers list about the newly supported robot.  Even if no one uses it immediately, it will show up in google searches for people in the future.

gavanderhoorn:
Thanks for contributing @EricMarcil 👍

PR #168, opened 2017-04-07
Added launch files for the DX200 and YRC1000 controllers

EricMarcil:
Added launch files for the DX200 and YRC1000 controllers to the motoman_driver folder.
Even though the DX200 and YRC1000 controllers communication are the same as the DX100 controller and that connection to those controller can be launch using the reference to "dx100" for controller, for clarity and ease of use, we add explicit launch file for the newer controllers.


gavanderhoorn:
Merging as it is needed for ros-industrial/motoman_experimental#13, and the launch files look ok.

PR #169, opened 2017-04-10
Issue166 xacro naming

shaun-edwards:
Do not merge, depends on #165.  These changes address bugs in the `sda10f` xacros.  Prefixing is now correctly supported and tested :)

This could be seen as a breaking change since some frame names will not be prefixed, whereas before they weren't.  However, if you tried to add a prefix to the robot name, the xacro would fail.

### DIFF
--- a/motoman_sda10f_support/test/sda10f_test.xacro
+++ b/motoman_sda10f_support/test/sda10f_test.xacro
@@ -1,0 +1,18 @@
+<?xml version="1.0" ?>
+
+<robot name="motoman_sda10f_test" xmlns:xacro="http://ros.org/wiki/xacro">
+	<xacro:include filename="$(find motoman_sda10f_support)/urdf/sda10f_macro.xacro" />
+	<xacro:motoman_sda10f prefix=""/>
+	<!-- xacro below should not be used, use form above-->
+	<!--xacro:sda10f/-->
+	<xacro:motoman_sda10f prefix="my_prefix_"/>
+	
+	
+	<joint name="base_link_joint" type="fixed" >
+    <origin xyz="1.0 0 0"/>
+    <parent link="torso_base_link" />
+    <child link="my_prefix_torso_base_link" />
+  </joint>
+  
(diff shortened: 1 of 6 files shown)


shaun-edwards:
@Levi-Armstrong, could you review this change.  The xacro parser in jade/kinetic is more picky about things.  I updated the travis file in PR #165, which caused CI to fail.  This PR includes that change as well as the fix for the xacro files.  In addition I added a test for xacro prefixing (since it was broken as well).  Others may want to consider this as well.  

Once approved, I'll cherry-pick this update into the jade-devel branch.

motoman_sda10f_support/test/sda10f_test.xacro
```diff
@@ -0,0 +1,18 @@
...
+<robot name="motoman_sda10f_test" xmlns:xacro="http://ros.org/wiki/xacro">
+	<xacro:include filename="$(find motoman_sda10f_support)/urdf/sda10f_macro.xacro" />
+	<xacro:motoman_sda10f prefix=""/>
+	<!-- xacro below should not be used, use form above-->
+	<!--xacro:sda10f/-->
+	<xacro:motoman_sda10f prefix="my_prefix_"/>
+	
+	
+	<joint name="base_link_joint" type="fixed" >
+    <origin xyz="1.0 0 0"/>
```
Levi-Armstrong: Everything below here, the indentation is showing up different than the rest of the file.

gavanderhoorn: Yeah. @shaun-edwards: 2-spaces indenting is the default, but seeing as almost all `xacro`s in this package use tabs, it would be better to stay consistent.

shaun-edwards:
@Levi-Armstrong, I corrected the spacing issue.  Can you complete the review?

Levi-Armstrong (approved):

Levi-Armstrong:
:+1:  Ok to merge!

Issue #172, opened 2017-05-22
IKfast solution for MH5 Robot


shaun-edwards:
Please create an IKFast solution for the [MH5](https://github.com/ros-industrial/motoman/tree/kinetic-devel/motoman_mh5_support).  To achieve this, perform the following:
 - [ ] Create an IKFast solution (see [tutorial](http://docs.ros.org/kinetic/api/moveit_tutorials/html/doc/ikfast_tutorial.html)).  Note, there is a [docker](https://hub.docker.com/r/personalrobotics/ros-openrave/) of openrave that is known to work.
 - [ ] Create a test program that verifies the resulting solution returns the nearest IK solution in joint space.

shaun-edwards:
@nsnitish, can you address this?

@gavanderhoorn, are you aware of anything that verifies the solution (i.e. the second item)?  Based on my experience, this is not the case.  Perhaps such a test should be added to the [kinematics_test](https://github.com/ros-planning/moveit_kinematics_tests) repo.

gavanderhoorn:
> @gavanderhoorn, are you aware of anything that verifies the solution (i.e. the second item)?

yes, OR actually has built-in support for testing these plugins. but that is not used with the command line that the tutorial gives you (which is actually not the correct one - see [Generating ikfast plugin for 5 DOF robot](http://answers.ros.org/question/196753) on ROS Answers).

See [How to verify generated IKfast plugin?](http://answers.ros.org/question/227219) on ROS Answers.

gavanderhoorn:
Btw, before such an IKFast plugin is generated, I would first address #122.

shaun-edwards:
FYI...#122 has been addressed.

nsnitish:
@gavanderhoorn , In [motman_mh5 catalog](http://www.motoman.co.uk/uploads/tx_catalogrobot/MH5_E.pdf) it is mentioned that it is a 6 axis robot. Also, from openrave's robot info I can see 6 links in addition to the base link. Can you please confirm if it is a 6 or 5 DOF robot?
Also, ikfastdemo.cpp referred in [How to verify generated IKfast plugin?](http://answers.ros.org/question/227219/how-to-verify-generated-ikfast-plugin/) is not available. Do you know of any other source to verify the ikfast solution?

gavanderhoorn:
[...]

PR #176, opened 2017-06-16
Created motoman_mh5_moveit_config

nsnitish:
@shaun-edwards  ! Moveit Configuration file as a part of Issue #172.


gavanderhoorn:
@nsnitish: I'm not entirely sure what happened here. The PR includes a single commit with "0.3.7" as commit message. Was that intentional?

nsnitish:
@gavanderhoorn ! That was a mistake. I will update the commit.

gavanderhoorn:
No problem, just asking.

Also: unless you discussed this with @shaun-edwards and you decided otherwise, new packages typically go into the `motoman_experimental` repository. The fact that `motoman_mh5_support` is in this repository does not mean the moveit config needs to be there as well.

shaun-edwards:
@gavanderhoorn, The URDF for the mh5 has been vetted, I'm not sure why we should wait for the MoveIt package.  We have an MH5 robot that we are actively using.  I feel this is a relatively safe contribution.

gavanderhoorn:
@shaun-edwards wrote:
> @gavanderhoorn, The URDF for the mh5 has been vetted, I'm not sure why we should wait for the MoveIt package. 

Because we have a dev process, and this is a new package.

gavanderhoorn:
Btw: now that there is an MH support package in `motoman_experimental`, should the MH5 xacro be migrated and this MoveIt config be based on that?

shaun-edwards:
The process is to make sure it is vetted.  I think it is.  However, moving it to `motoman_experimental` and migrating to `motoman_mh_support` brings up a bigger issue that I hadn't considered.  We will basically block all `mh` robots until they are all vetted.  We will have lost the ability to migrate them one by one since the `motoman_mh_support` dependency would have to be migrated to `motoman`.  Thoughts?

gavanderhoorn:
@shaun-edwards wrote:
> The process is to make sure it is vetted.

That is not all there is to the process. Placing this in experimental also allows us to make sure the package can get exposed to use-cases and environments that are different from those it has seen when it is/was used by the original author/contributor, without all the stability and bw-compatibility requirements that go with a package that has been merged into and distributed from a non-experimental repository. The fact that you use it with your robot and haven't encountered any issues with it is not really important.

> We will basically block all `mh` robots until they are all vetted.  We will have lost the ability to migrate them one by one since the `motoman_mh_support` dependency would have to be migrated to `motoman`.

I'm not sure I understand what you are saying here.

shaun-edwards:
@nsnitish, please close this and re-open against `motoman_experimental`

@gavanderhoorn, the `motoman_mh_support` package, after @nsnitish opens the PR, will contain files for the `mh5` and the `mh180`.  If we wanted to migrate just the `mh5`, we couldn't.  We would have to migrate both the `mh5` and the `mh180` because we couldn't "split" the `motoman_mh_support` package.

gavanderhoorn:
@shaun-edwards wrote:
> @gavanderhoorn, the `motoman_mh_support` package, after @nsnitish opens the PR, will contain files for the `mh5` and the `mh180`.  If we wanted to migrate just the `mh5`, we couldn't.  We would have to migrate both the `mh5` and the `mh180` because we couldn't "split" the `motoman_mh_support` package.

yes, that is an unfortunate consequence of the many-variants-in-a-single-pkg approach.

Thinking about it a bit I can only come up with creating two packages: `motoman_mh_support` and `motoman_mh_support_experimental` (or something similar) and then manually migrating (ie: moving) sets of files between those.

That is not very elegant though.

gavanderhoorn:
@nsnitish: did you intend to copy the ` kinematics_base_test` package into your PR? df2b85a seems to add the package as well as your MH5 test.

PR #183, opened 2017-10-10
MotoROS Driver v1.6.0

ted-miller:
- Add fix for the different possible lengths of ROS_MSG_MOTO_JOINT_TRAJ_PT_FULL_EX.
- Add logic to handle multiple messages (or partial messages) coming in on the socket at the same time.
- Removed redundant check for message size.
- Removed startup delay for non-DX100 controllers.
- Added additional configurations of INFORM jobs.


ted-miller:
v1.6.0
- Added separate server for I/O (based on branch from gavanderhoorn).
- Updated pendant IONAME.DAT file.
- Updated socket select statement to handle concurrent activity on
multiple servers.
- Cleaned up #include statements.

@gavanderhoorn  I just updated this to include your changes from your io-server branch.  My latest commit shows up in the pr "commits" tab, but doesn't appear here in the conversation.  Is that normal?

shaun-edwards:
>@gavanderhoorn I just updated this to include your changes from your io-server branch. My latest commit shows up in the pr "commits" tab, but doesn't appear here in the conversation. Is that normal?

That's not normal.  Click on the "Commits" tab above to make sure everything you want is included in this PR.

ted-miller:
It's all there in commit 18782a6fa438e6817ac0bd771de7d2ac363c295d
(v1.6.0_review)

gavanderhoorn:
Yeah, I'm not sure what happened here. I see all commits as well, but it doesn't show in the *Conversation*.

gavanderhoorn:
General comment / question (which I'd hoped we could've discussed before you merged my changes into this PR): would we want to keep the IO implementation in `MotionServer.c` around for 1 or 2 versions, just to maintain bw-compatibility with non-ROS clients out there?

We can update the ROS nodes in `motoman_driver`, but afaic `MotoROS` is also being used without the ROS components, so it might perhaps be nice to those users to first deprecate the IO infrastructure in `MotoROS` and then remove it in some future version.

Thoughts @ted-miller ?

motoman_driver/MotoPlus/MotionServer.c
```diff
@@ -463,24 +436,26 @@ int Ros_MotionServer_SimpleMsgProcess(Controller* controller, SimpleMsg* receive
 		ret = Ros_MotionServer_JointTrajPtFullExProcess(controller, receiveMsg, replyMsg);
 		break;
 
+
+//Maintain backward compatibility for users who are sending I/O over motion-server
```
ted-miller: @gavanderhoorn  Here is where I continue to process I/O requests on the motion server.  It just calls the io function from ioserver.c

ted-miller:
I just jumped on this right away because I was already tinkering with it and I need to move on to other things next week.

I didn't remove the command from the motion server.  It will accept these commands on either server to maintain backward compatibility.  I plan to create a wiki page at some point that directs people to use the I/O server, but I don't intend to remove the old functionality.

motoman_driver/MotoPlus/MotionServer.c
gavanderhoorn: O, that is very nice. I'd missed that. :+1: 

Are the MotoPlus IO functions thread safe? If `MotionServer` and `IoServer` are run in different threads, would `*_ReadIOBit(..)` (and the others) need to introduce locking/mutexes?

gavanderhoorn:
No problem. It was just something I thought we should discuss before merging a potentially breaking change like this.

And I missed that you'd already taken care of it.

>  I plan to create a wiki page at some point that directs people to use the I/O server, but I don't intend to remove the old functionality.

Ok. As long as it doesn't lead to duplication of (maintenance) effort *and* we clearly document the deprecation of the `MotionServer` IO interface.

motoman_driver/MotoPlus/MotionServer.c
ted-miller: The I/O API is thread safe.

gavanderhoorn (changes requested):
we've run into some issues while testing this (on our SIA20F): the TP displays an error after installing the MotoPlus application:

> 8000[0]: Could not save PARAMETERS file

Until those are resolved, this should not be merged.

gavanderhoorn (approved):
Tested this with our SIA20F (+rail): works for me, so +1.

gavanderhoorn:
Error reported in https://github.com/ros-industrial/motoman/pull/183#pullrequestreview-75447433 was due to having multiple instances of the MotoROS application on the controller (one was invisible due to an 'incompatible filename').

After remedying that, everything worked as expected.

gavanderhoorn:
We should pull this into `indigo-devel` as well.

gavanderhoorn:
@shaun-edwards: there are quite some changes missing from `indigo-devel` ([comparison](https://github.com/ros-industrial/motoman/compare/indigo-devel...kinetic-devel)), but all should be Indigo compatible.

Should we do a sync between `kinetic-devel` and `indigo-devel`?

PR #184, opened 2017-10-12
WIP: MotoROS: refactor IO server into stand-alone task.

gavanderhoorn:
As per subject.

Just leaving this here as I did this some time ago but never got around to making a PR.

@ted-miller: is this something that you would consider merging into MotoROS?

I know you don't want to encourage users to use these interfaces as a replacement for a proper fieldbus - and neither do I - but at least in this way IO msgs won't cause trouble with motion command streams.

### DIFF
--- a/motoman_driver/MotoPlus/Controller.c
+++ b/motoman_driver/MotoPlus/Controller.c
@@ -34,6 +34,7 @@
 #include "CtrlGroup.h"
 #include "SimpleMessage.h"
 #include "Controller.h"
+#include "IoServer.h"
 #include "MotionServer.h"
 #include "StateServer.h"
 #include "RosSetupValidation.h"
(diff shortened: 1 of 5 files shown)


gavanderhoorn:
Note that I have not run this on an actual controller, I've only compile-tested it.

gavanderhoorn:
I noticed there is some duplication in how Simple Message msg traffic is (de)serialised. If we merge this it might be worthwhile to take a look at moving some of that into a file/library of its own.

ted-miller:
I'm seeing increased situations where basic I/O control is needed.  So I agree that this is justified.

I cleaned up the deserializing code a bit... but there's still some duplication.  I personally think it's ok, as it's easier to understand when the code for each of the three servers is localized in its own file.

See #183 
commit 18782a6fa438e6817ac0bd771de7d2ac363c295d

gavanderhoorn:
Ok. Thanks for that.

I'll keep this open until #183 is merged.

Discussion to continue on #183.

gavanderhoorn:
Closing this as #183 includes it.

Issue #186, opened 2017-11-22
Sync indigo with kinetic


gavanderhoorn:
@shaun-edwards: there are quite some changes missing from `indigo-devel` ([comparison](https://github.com/ros-industrial/motoman/compare/indigo-devel...kinetic-devel)), but all should be Indigo compatible.

Should we do a sync between `kinetic-devel` and `indigo-devel`?

gavanderhoorn:
Note: the changes to `.travis.yml` ([comparison](https://github.com/ros-industrial/motoman/compare/indigo-devel...kinetic-devel#diff-354f30a63fb0907d4ad57269548329e3)) should probably not be merged into `indigo-devel`.

gavanderhoorn:
@shaun-edwards: 🛎 ?

gavanderhoorn:
Sync was completed with the merge of #194.

PR #194, opened 2018-01-24
Sync kinetic into indigo

gavanderhoorn:
This is a first attempt at addressing #186.

Initially I considered cherry-picking relevant commits into `indigo-devel`, but that changes commit hashes, and we'd have to keep track (in the future) of whether specific changes were back-ported or not from `kinetic-devel`.

This is a straight `merge` from `kinetic-devel` into `indigo-devel`, with one additional commit that reverts the changes to the travis config (ab90e74).

There are actually only a few commits in `kinetic-devel` that touch non-MotoROS artefacts, so syncing the two branches should be ok.

I've compile-tested `indigo-devel` with this PR and ran the available tests under Indigo. All passed.


gavanderhoorn:
We could also opt for a hybrid approach: cherry-pick commits from `kinetic-devel` but merge them using PRs that reference the PRs they originally came from.

That way we keep some form of connected-history.

gavanderhoorn:
@shaun-edwards.

shaun-edwards (approved):

shaun-edwards:
@gavanderhoorn, thanks for this.  Sorry for the delay in review.

PR #197, opened 2018-02-08
add doc strings to all launch file args.

hsd-dev:
Using the 'doc' attribute on 'arg' elements.

### DIFF
--- a/motoman_driver/launch/motion_streaming_interface.launch
+++ b/motoman_driver/launch/motion_streaming_interface.launch
@@ -3,10 +3,10 @@
 -->
 <launch>
 	<!-- IP of robot (or PC running simulation) -->
-	<arg name="robot_ip" />
+	<arg name="robot_ip" doc="IP of controller (only required if not using industrial simulator)" />
 
 	<!-- Load the byte-swapping version of robot_state if required -->
-	<arg name="use_bswap" />
+	<arg name="use_bswap" doc="If true, robot driver will byte-swap all incoming and outgoing data (only required if not using industrial simulator)" />
 
 	<!-- put them on the parameter server -->
 	<param name="robot_ip_address" type="str" value="$(arg robot_ip)" />
--- a/motoman_driver/launch/motion_streaming_interface.launch
+++ b/motoman_driver/launch/motion_streaming_interface.launch
@@ -3,10 +3,10 @@
 -->
 <launch>
 	<!-- IP of robot (or PC running simulation) -->
-	<arg name="robot_ip" doc="IP of controller (only required if not using industrial simulator)" />
+	<arg name="robot_ip" doc="IP of controller" />
 
 	<!-- Load the byte-swapping version of robot_state if required -->
-	<arg name="use_bswap" doc="If true, robot driver will byte-swap all incoming and outgoing data (only required if not using industrial simulator)" />
+	<arg name="use_bswap" doc="If true, robot driver will byte-swap all incoming and outgoing data" />
 
 	<!-- put them on the parameter server -->
 	<param name="robot_ip_address" type="str" value="$(arg robot_ip)" />
--- a/motoman_mh5_support/launch/robot_interface_streaming_mh5.launch
+++ b/motoman_mh5_support/launch/robot_interface_streaming_mh5.launch
@@ -9,10 +9,10 @@
     robot_interface_streaming_mh5.launch robot_ip:=<value> controller:=<fs100|dx100>
 -->
 <launch>
-	<arg name="robot_ip" />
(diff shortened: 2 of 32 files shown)


motoman_driver/launch/motion_streaming_interface.launch
```diff
@@ -3,10 +3,10 @@
 -->
 <launch>
 	<!-- IP of robot (or PC running simulation) -->
-	<arg name="robot_ip" />
+	<arg name="robot_ip" doc="IP of controller (only required if not using industrial simulator)" />
```
gavanderhoorn: This launch file cannot be used with the `industrial_robot_simulator`, so this seems like a strange comment as a doc string.

Can you clarify?

motoman_mh5_support/launch/robot_interface_streaming_mh5.launch
```diff
@@ -9,10 +9,10 @@
     robot_interface_streaming_mh5.launch robot_ip:=<value> controller:=<fs100|dx100>
 -->
 <launch>
-	<arg name="robot_ip" />
+	<arg name="robot_ip" doc="IP of controller (only required if not using industrial simulator)" />
 	
   <!-- controller: Controller name (fs100 or dx100) -->
-  <arg name="controller"/>
+  <arg name="controller" doc="Name of the controller"/>
```
gavanderhoorn: This is a bit ambiguous: it's really the controller *series* or *type* that is specified by the `controller` `arg`.

gavanderhoorn: Would probably also be good to add the possible values there (ie: `dx100`, `fs100`, `dx200` and `yrc1000`).

motoman_sda10f_moveit_config/launch/default_warehouse_db.launch
```diff
@@ -1,6 +1,6 @@
 <launch>
 
-  <arg name="reset" default="false"/>
+  <arg name="reset" default="false" doc="Reset DB"/>
```
gavanderhoorn: This file (and all others in MoveIt configurations) will all be overwritten if the Setup Assistant is ever run again.

Do you feel it's worth it to manually add doc strings here, instead of updating the MSA templates upstream?

motoman_driver/launch/motion_streaming_interface.launch
gavanderhoorn: Additionally: the `arg` is really a required arg (there is no `default` attribute), so it's always required.

gavanderhoorn (changes requested):
Thanks for the PR. I've added some comments.

motoman_mh5_support/launch/robot_interface_streaming_mh5.launch
hsd-dev: Modified.

motoman_sda10f_moveit_config/launch/default_warehouse_db.launch
hsd-dev: Wasn't quite sure about this, reverted back.

On a side note, I couldn't find documentation for args/params used in move_group.launch like `max_safe_path_cost`

motoman_driver/launch/motion_streaming_interface.launch
hsd-dev: Somehow overlooked the missing default. Modified.

motoman_sda10f_moveit_config/launch/default_warehouse_db.launch
gavanderhoorn: That would be something to take up with the moveit team :)

motoman_sda10f_moveit_config/launch/move_group.launch
```diff
@@ -13,7 +13,7 @@
   <arg     if="$(arg info)" name="command_args" value="--debug" />
 
   <!-- move_group settings -->
-  <arg name="allow_trajectory_execution" default="true"/> 
+  <arg name="allow_trajectory_execution" default="true"/>
```
gavanderhoorn: Just noticed this whitespace change. Could you revert the changes to this file? They are out-of-scope for this PR.

gavanderhoorn:
Thanks for the updated PR. One minor comment left. If you could address that we can merge.

gavanderhoorn (approved):

gavanderhoorn:
Just waiting on Travis.

gavanderhoorn:
Thanks for the PR @ipa-nhg-hd. 👍

hsd-dev:
you are welcome :)

PR #199, opened 2018-03-07
MotoROS v1.7.1

ted-miller:
- Remove alarm for controllers with more than 4 control groups.

On DX100 and FS100, the presence of this alarm will cause initialization of the individual control groups to fail (Ros_CtrlGroup_Create).  The alarm has been removed to allow initialization to complete.

Additionally, there was concern about the fact that this alarm would happen after every reboot.  In an operational environment, this would require manual intervention on the pendant after every restart. This is not ideal.


motoman_driver/MotoPlus/Controller.c
```diff
@@ -214,7 +214,6 @@ BOOL Ros_Controller_Init(Controller* controller)
 
 	if (controller->numGroup > MOT_MAX_GR)
 	{
-		mpSetAlarm(8001, "WARNING: Too many groups for ROS", 0); //force user to acknowledge ignored groups
```
gavanderhoorn, line 217: You've probably thought about this -- and I haven't checked yet myself -- but is this alarm now raised at least once?

ted-miller: No. I removed the alarm entirely. I don't have a "good" way to trigger the alarm only once.

My thought is that this issue should become pretty obvious very quickly, since the base protocol only supports four groups.  Plus, if someone does try to send a command to an invalid group number, MotoROS will return an error code ROS_RESULT_INVALID_GROUPNO.

gavanderhoorn: > I don't have a "good" way to trigger the alarm only once.

Is there no per-application storage facility or something similar? Range of system parameters that could be used? File system perhaps with a marker file? It's not terribly important, but it'd be nice if we could see whether we can avoid users having to actually try and *use* the driver before they run into this issue/error code.

ted-miller: > Is there no per-application storage facility or something similar? Range of system parameters that could be used? File system perhaps with a marker file?

Nope.

The only thing I could do is generate a dummy robot job with some fixed comment in it. But, the API's to read and write that job don't work when there's an active alarm. Plus, there are restrictions about what mode the robot must be in... so there are cases where it wouldn't work.

gavanderhoorn: <del>I'm not an expert, so this may be a stupid question, but using `mpCreate(..)` to create a marker file and then `mpStat(..)` on subsequent bootups to see whether we need to post the alarm is not an option?</del>

hm, this of course leads to problems with users removing and then reinstalling the `motoros` application. As there is no real 'uninstall' procedure, the file would not be removed.

gavanderhoorn: Just wondering: could the 'MOTOMAN DRIVER' setting under 'OPTION FUNCTION' not do some rudimentary checking?

ted-miller: I don't have any reliable storage area for such a file.  The "DRAM" drive is volatile and erases on reboot.  The "SRAM" drive requires a $1500 addon board.  The USB drive only works on DX200 and newer.

ted-miller: And I cannot write the file to the internal storage card.  This causes conflicts with the internal controller software which can eventually brick the system. (I've tried that in the past.)

ted-miller: The "MOTOMAN DRIVER" setting in Maintenance mode is just a shortcut that enables someone with the management-password to activate protected parameters in the background.  

I don't have the ability to customize what actions are performed by MOTOMAN DRIVER, nor do I have the ability to designate parameters for my purposes.  This would need to go through our global-development-committee and would take some significant time/effort to make modifications to the feature. I don't think this little thing is worth it.

gavanderhoorn: I understand. Was just trying to come up with a way to see whether we could somehow check for the `<= 4 groups` condition only once by exploiting something that was already there.

gavanderhoorn (approved):
Tested on our FS100.

No regressions detected, although I cannot test everything.

gavanderhoorn:
Sorry for leaving this open so long @ted-miller.

I'll go ahead and squash and merge this.

PR #200, opened 2018-03-12
motoros: add newline at end of MotoROS header

gavanderhoorn:
As per subject.

I keep getting compiler warnings about this otherwise.


ted-miller:
Are you using mpBuilder or bypassing it so you can compile on linux?  When mpBuilder converts the files to ascii, it checks to see if the file ends with a newline and adds it as needed. (specifically to avoid this warning)

ted-miller (approved):

gavanderhoorn:
@ted-miller wrote:
> Are you using mpBuilder or bypassing it so you can compile on linux? 

maybe ;)

> When mpBuilder converts the files to ascii, it checks to see if the file ends with a newline and adds it as needed. (specifically to avoid this warning)

yes, I figured that out after I compared the sources to those in `..\mpAscii 0`.

I figured we might as well just add it. Save `mpBuilder` some work :)

gavanderhoorn:
Thanks @ted-miller for the review :+1:

PR #201, opened 2018-03-12
motoros: remove UTF-8 BOM from source files.

gavanderhoorn:
As per subject.

See the commit msg for some more info.

Visual Studio adds a BOM when creating files, so that's probably how these got added in the first place. But it should not add it to files you open with it that don't have it.

See also [UTF-8 without BOM](https://stackoverflow.com/questions/5406172/utf-8-without-bom) on SO.


gavanderhoorn:
Note that the diff appears to show 'no changes': the BOM is not rendered by Github, so that is expected.

ted-miller (changes requested):
I'm personally against removing the BOM.  General users will be using the MotoPlus SDK for Visual Studio, which supports the BOM and properly removes it before compiling.

For international users (such as Japan), the UTF encoding allows them to comment the code in their native language.

gavanderhoorn:
The BOM is not required to be able to use UTF. In fact, the files still are UTF-8 encoded. VS (older versions at least) automatically added the BOM, but it's debatable whether that is/was a good thing.

ted-miller:
Of course it's not required.  But it makes it easier to handle the files in other editors and/or text-compare tools.

I frequently deal with source files with Japanese comments.  These often get corrupted due to different editors/email-clients/tools misinterpreting the encoding and then re-saving the files in the wrong encoding. I really feel that the BOM is a good thing because it makes the file more deterministic.

(I know of multiple groups in Japan using MotoROS.)

gavanderhoorn:
Ok. Let's solve this some other way.

Note btw that not all files have a BOM. There seems to be some inconsistency.

Issue #203, opened 2018-04-13
sia5f joint_u and joint_b rotation sign error


tdl-ua:
As I ran the launch-file _test_sia5d.launch_ in _motoman/motoman_sia5d_support/launch/_, I noticed that the simulated behavior of the robot is not the same as the physical robot. The joints U and B rotate in the opposite direction when comparing simulation and physical robot. This issue also affects the planning and execution of trajectories with MoveIt!.

I have now made changes to the file _sia5d_macro.xacro_ and updated the file _sia5d_macro.urdf_ using xarco.py to fix the issue. I will create a pull request shorty that contains the changes made. 

It should be noted that our lab uses the sia5f robot version, where _f_ indicates the controller type FS100. Therefore, the behavior of the sia5d version could be correct when using the unchanged version of the urdf model.

gavanderhoorn:
This was fixed in #204.

PR #204, opened 2018-04-13
Change sign of joint_u and joint_b rotation vector (#203)

tdl-ua:
The changes made here fix the bug described in (#203) where the simulated
behavior of the robot (SIA5F) is not the same as the physical robot.
The joints U and B rotate in the opposite direction when comparing
simulation and physical robot.


shaun-edwards (approved):

shaun-edwards:
@tdl-tbslab, thanks for the contribution.

PR #205, opened 2018-04-17
MotoROS v1.7.2

ted-miller:
- Add velocity feedback for DX100 and FS100 controllers.


gavanderhoorn (approved):
This seems to work well in a quick test. Haven't noticed any regressions.

Velocity data coming out of this is a bit noisy, and there appears to be some lag, but it does not look as discretised as the previous implemention.

gavanderhoorn:
I'll wait on Travis before merging this.

gavanderhoorn:
Thanks for adding this @ted-miller.

@mbharatheesha fyi.

gavanderhoorn:
I'm slightly confused about the commits included in this PR, but a squash merge will probably 'fix' that.

gavanderhoorn:
Thanks @ted-miller :+1:

Danfoa:
Hi @ted-miller @gavenderhoorn, first thanks for offering this support. I just want to mention that with this update, the troubleshooting web page of the motoman package needs to be updated, since it mentions that velocity feedback is not present in the FS100 controller.

PR #207, opened 2018-05-01
Multiple fixes for build scripts and package manifests

gavanderhoorn:
Connected to ros-industrial/motoman_experimental#29.

Apart from (in some cases major) cosmetic changes, there are a few changes to pkgs that could affect current users:

 - removal of `Inform` and `MotoPlus` directories from `motoman_driver`'s include path: these should not be on the include path at all, as there is nothing in them that is relevant to the ROS side of `motoman_driver`. This has a very low chance of affecting current users, as those paths were not exported by `catkin_package(..)`.
 - `roslint` has been changed to a `test_depend`: it was a `build_depend` (as that was required for package format 1), and never exported as a dependency, but still a change
 - package manifest and build script for `motoman_msgs` has been cleaned up: `genmsg` was incorrectly listed as a `build_depend` and a `run_depend`. It was also exported as a dependency. It's not needed (`message_generation` already brings it in) and also should not be exported. Users 'benefiting' from this mistake will now have to depend on `genmsg` directly.

Note: even though most/all support pkgs are marked as `deprecated`, I've updated all the manifests to format 2 and done some cleanup to their build scripts as well. I'm not sure when we'll get to actually removing these, so in the meantime I thought it'd be nicer to harmonise everything in the repository.

### DIFF
--- a/motoman/package.xml
+++ b/motoman/package.xml
@@ -1,20 +1,25 @@
-<package>
+<?xml version="1.0"?>
+<package format="2">
   <name>motoman</name>
   <version>0.3.5</version>
   <description>The motoman stack constains libraries, configuration files, and ROS nodes for controlling a Motoman robot from ROS-Industrial</description>
   <maintainer email="sedwards@swri.org">Shaun Edwards</maintainer>
   <license>BSD</license>
-  <url type="website">http://ros.org/wiki/motoman</url>
+  <url type="website">http://wiki.ros.org/motoman</url>
   <author email="sedwards@swri.org">Shaun Edwards</author>
 
   <buildtool_depend>catkin</buildtool_depend>
-  
-  <run_depend>motoman_driver</run_depend>
-  <run_depend>motoman_mh5_support</run_depend>
-  <run_depend>motoman_sia10d_support</run_depend>
-  <run_depend>motoman_sia20d_moveit_config</run_depend>
-  <run_depend>motoman_sia20d_support</run_depend>
-  <run_depend>motoman_sia5d_support</run_depend>
+
+  <exec_depend>motoman_driver</exec_depend>
+  <exec_depend>motoman_mh5_support</exec_depend>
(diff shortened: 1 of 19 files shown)


gavanderhoorn:
I've compile-tested this on a Kinetic installation with `catkin_tools`, so I'm fairly certain the various inter-pkg-dependencies are still ok.

Note also: this does not resolve all of `catkin_lint`'s complaints. In particular, the following warnings are still present when running `catkin_lint -W2 src/motoman`:

```
motoman: notice: package description starts with boilerplate 'The motoman'
motoman_driver: warning: exported package include path but no exported library
motoman_driver: notice: package description starts with boilerplate 'This package'
motoman_driver: CMakeLists.txt(172): notice: list of source files should be sorted
motoman_driver: CMakeLists.txt(212): notice: list DIRECTORY should be sorted
motoman_msgs: notice: package description starts with boilerplate 'Set of'
motoman_msgs: CMakeLists.txt(12): notice: list FILES should be sorted
catkin_lint: checked 11 packages and found 7 problems
```

The 'should be sorted' notices are strange: according to my editor, everything *is* sorted.

The 'boilerplate description' notices I'm going to ignore.

The 'exported pkg include path but no library' is true: this is I believe because of swap vs no-swap libraries. We could ask ourselves whether a driver package should export its headers though. The only use-case I can see is if downstream users would want to use the libraries directly, instead of the nodes/ROS API.

gavanderhoorn:
This PR also includes 8472f05, which adds @ted-miller and @EricMarcil as `author`s to the `motoman_driver` manifest.

gavanderhoorn:
@shaun-edwards: could I ask you to do a regular merge if/when you decide to merge the PR? That way the commit history is retained, which I believe in this case is valuable.

gavanderhoorn:
And a friendly :hamburger:

shaun-edwards (approved):

Issue #208, opened 2018-05-01
Velocity reported on FS100 seems to be 10x too large


gavanderhoorn:
I just ran some Python scripts with MotoROS v1.7.2 (merged in #205) on an FS100 and noticed something which I believe is not as it should be.

![plot0](https://user-images.githubusercontent.com/4550046/39475312-18a6da54-4d58-11e8-884e-769d39b25d80.png)

This plot shows the position and velocity as reported by the driver for a single joint. Note the velocity peak reaches 1.2 m/s (this is a linear axis).

![plot1](https://user-images.githubusercontent.com/4550046/39475404-6943ffe6-4d58-11e8-9710-d51359d28bc6.png)

The plotting tool can calculate the first derivative for me, so I asked it to and we end up with this second plot. Notice how this is essentially the same curve, but it reaches a max of 0.12 m/s.

I've not (yet) checked the API docs for `mpSvsGetVelTrqFb(..)`, but could it be that there is a 10x scaling factor somewhere we forgot to take into account?

gavanderhoorn:
> could it be that there is a 10x scaling factor somewhere we forgot to take into account?

From the documentation for `mpSvsGetVelTrqFb(..)` I have the impression that is the case:

> Specify the pointer to the feedback speed receiving array. The feedback speed is written in the array in units of 0.1 pulse/sec.

So that is indeed a 10x difference.

Sorry for not noticing this in the review of #205 @ted-miller.

gavanderhoorn:
With the merge of #213 this is now fixed.

PR #210, opened 2018-05-03
MotoROS v1.7.3

ted-miller:
- Fix unit scaling for velocity feedback on DX100 and FS100.


ted-miller:
@gavanderhoorn  This should fix #208

ted-miller:
@Tom-Pfahler Please make sure to pull this into your fork and recompile before posting your updates.

gavanderhoorn:
@ted-miller: not sure what is going on here with the commits. I would've expected to see only 07fd8c3?

ted-miller:
I have no idea why, but it says that I'm 9 commits ahead.  I also thought it was strange.  Maybe I just need to re-base?
9 commits ahead

ted-miller:
Also, I'm thinking maybe we should kill this PR.  @Tom-Pfahler is going to submit a new one today for v1.8.0 that includes support for the YRC1000 Micro controller.

So, I'm thinking that I should just merge his changes locally and submit a new PR.  I think this would be easier for the revision history.  Your thoughts?

gavanderhoorn:
Hm.

Well you could just hard reset your local `kinetic-devel` to `origin/kinetic-devel` (provided that `origin` points to `ros-industrial/kinetic-devel`).

You would lose your local changes (ie: 07fd8c3), but that is just a few lines (and you could even save those if you'd really want). Then first create a feature branch from`kinetic-devel` (fi: `motoros173`), re-apply the changes, push the feature branch and open a new PR.

Then after things get merged, do a `git fetch origin` (again: assuming it points to the main repository), and then a `git checkout kinetic-devel` (just to be sure) and finally a `git merge origin/kinetic-devel`. That should cleanly update your local `kinetic-devel` with the latest changes.

From that point on just avoid working in `kinetic-devel`, and your history should stay managable.

If things get really messy, you could always `git checkout kinetic-devel && git reset --hard origin/kinetic-devel` (note: you will lose local changes committed to `kinetic-devel`).

gavanderhoorn:
If that sounds too complicated, we can always close this and have the fix for #208 as part of 1.8.0.

gavanderhoorn:
Just tested this and 've verified that it works.

Would you want to submit a new PR @ted-miller?

ted-miller:
Yeah. I'm going to submit a new one on Monday. There is another edit I need to make to the setup-validation library.

ted-miller:
Closing. Will submit new PR with additional updates on top of these.

PR #212, opened 2018-05-03
Adding new robot support pkgs

EricMarcil:
Adding models:
- GP7 (YRC1000)
- GP8 (YRC1000)
- GP12 (YRC1000)
- MH12 (DX200)
- MH50 (DX200)


EricMarcil:
Following comments received:
1.	For links, origin be zero and modification to the geometry (aka. scale). This make it easier to review and maintain going forward.
2.	The meshes are two large. There are two reasons for keeping these as small as possible. GitHub limits the size to 2GB for a given repository and the second is performance when using this within ROS.
- Visual: I recommend decimating the to be less than 1MB
- Collision: I recommend using decimating or even using convex hulls to be less than 100 KB
3.	Format Changes 
- The file names will need to be lower case and match the current style of existing packages: example "motoman_axis_u"
- The link names should be "base_link, link_1_s, link_2_l, .....)
- The joint name should be "joint_1_s, joint_2_e, .....)

The following corrections we made:

Models GP7, GP8, GP12, MH12, MH50
- Modified STL model to remove scaling and origin correction from the URDF file.
- Change filename to lowercase
- Changes link and joint names to indicate link order and motoman name: example: joint_1_s

gavanderhoorn:
@EricMarcil: not ignoring you. :+1: for the contribution.

If @shaun-edwards doesn't get to this sooner, I'll review this.

EricMarcil:
@shaun-edwards this has been out for a while.  Can you take a look at this pull request and please process it.  I worked closely with SwRI and @Levi-Armstrong review it and send me his comments directly.  I incorporated his suggestions in the May 11th commit.  If there are any other changes required, let me know. Thanks.

shaun-edwards (approved):
There a few minor issues (nothing worth holding this up any longer), but overall good.

shaun-edwards:
@EricMarcil, sorry (really sorry) for taking so long to merge this.  Thanks for taking the time to put these together.  I'm going to merge these once Travis comes back successful.  Thanks again!

PR #215, opened 2018-05-16
 point-streaming: further development based on PR #88 

tdl-ua:
I tried to address some of the comments made by @shaun-edwards in PR #88 (created by @jim-rees):
- _Comment 1_: Made `JointTrajectoryInterface::jointCommandCB()` pure virtual so that `JointTrajectoryInterface::jointCommandCB()` implementation can be deleted
- _Comment 2_: Message was changed to `ROS_DEBUG`
- _Comment 3_: Not sure what the point of that sleep call is here so not sure what the constant would be called. No change made.
- _Comment 4_: It is the same reconnect logic as in `case TransferStates::STREAMING`. I am assuming that it was added to be consistent with `case TransferStates::STREAMING`. Therefore, no change made.
- _Comment 5_: see Comment 4.

As to why the state `TransferStates::POINT_STREAMING` comes up multiple times; I have no idea. Was wondering the same thing. Obviously, there must be a good reason for having a class `MotomanJointTrajectoryStreamer` that extends `JointTrajectoryStreamer` but I don't really understand it at this point.

I have tested the code on an sia5 manipulator and found that commanded constant-velocity motions are carried out very smoothly. I also teleoperated the manipulator using a 6 DOF haptic interface with this point-streaming implementation and also experienced very smooth end-effector motion.


gavanderhoorn:
Thanks @tdl-tbslab for getting this up-to-date again. I'll see if I can test this out sometime soon.

@ted-miller: are there any things in here that catch your eye?

gavanderhoorn:
@alemme, @andreaskoepf any comments? This sounds similar to what you described on ROS Discourse, sans the trapezoidal velocity generator (and possibly some other things you didn't mention).

andreaskoepf:
@gavanderhoorn The approach of this PR and other point-streaming impls seems to be a dynamic online trajectory generation. The main difference to our TVP controller which runs inside the realtime OS of the controller is that we operate setpoint based and the setpoint can be changed in every cycle .. e.g. it does not continue the move to a previous target but immediately reacts upon the next target, which allows us especially to minimize latency, see my [joystick video](https://www.youtube.com/watch?v=KgzYKmDy8jk).

shaun-edwards:
This has been suggested by multiple people.  I'm inclined to accept it as is.  Any thoughts @gavanderhoorn, @ted-miller, @alemme, @andreaskoepf?

I'll wait a couple of days, otherwise, I'll merge.

motoman_driver/src/joint_trajectory_streamer.cpp
```diff
@@ -460,6 +460,74 @@ void MotomanJointTrajectoryStreamer::streamingThread()
...
+        {
+          ROS_INFO("Point[%d] sent to controller", this->current_point_);
+          this->current_point_++;
+          time_since_last = 0.0;
+          time_of_last = ros::Time::now().toSec();
+          this->streaming_queue_.pop();
+        }
+        else if (reply_status.reply_.getResult() == MotionReplyResults::BUSY)
+        {
+          ROS_INFO("silently resending.");
```
gavanderhoorn: `BUSY` can occur quite frequently. Changing this to a throttled log output might be good. Or perhaps even a debug.

motoman_driver/src/joint_trajectory_streamer.cpp
```diff
@@ -460,6 +460,74 @@ void MotomanJointTrajectoryStreamer::streamingThread()
...
+        MotionReplyMessage reply_status;
+        if (!reply_status.init(reply))
+        {
+          ROS_ERROR("Aborting point stream operation.");
+          this->state_ = TransferStates::IDLE;
+          break;
+        }
+        if (reply_status.reply_.getResult() == MotionReplyResults::SUCCESS)
+        {
+          ROS_INFO("Point[%d] sent to controller", this->current_point_);
```
gavanderhoorn: Same here.

motoman_driver/src/industrial_robot_client/joint_trajectory_streamer.cpp
```diff
@@ -159,6 +159,73 @@ void JointTrajectoryStreamer::jointTrajectoryCB(const trajectory_msgs::JointTraj
...
+      ROS_INFO("Empty point received, cancelling current trajectory");
+      return;
+    }
+
+    //Else, Push point into queue
+    SimpleMessage message;
+    trajectory_msgs::JointTrajectoryPoint rbt_pt, xform_pt;
+
+    // select / reorder joints for sending to robot
+    if (!select(msg->joint_names, msg->points[0], this->all_joint_names_, &rbt_pt))
```
gavanderhoorn: Does `select(..)` also check that all required joints are present in the incoming `msg`?

If not, such a check would probably need to be added.

motoman_driver/src/joint_trajectory_streamer.cpp
```diff
@@ -460,6 +460,74 @@ void MotomanJointTrajectoryStreamer::streamingThread()
         }
       }
       break;
+
+    case TransferStates::POINT_STREAMING:
```
gavanderhoorn, line 464: @shaun-edwards: does this need to be added twice? Once here and once in the `industrial_robot_client` extension part of this package?

motoman_driver/src/joint_trajectory_streamer.cpp
```diff
@@ -460,6 +460,74 @@ void MotomanJointTrajectoryStreamer::streamingThread()
...
+      if (!this->connection_->isConnected())
+      {
+        ROS_DEBUG("Robot disconnected.  Attempting reconnect...");
+        connectRetryCount = 5;
+        break;
+      }
+      // otherwise, send point to robot.
+      tmpMsg = this->streaming_queue_.front();
+      msg.init(tmpMsg.getMessageType(), CommTypes::SERVICE_REQUEST,
+               ReplyTypes::INVALID, tmpMsg.getData());  // set commType=REQUEST
```
gavanderhoorn: Should the comment be removed?

motoman_driver/include/motoman_driver/joint_trajectory_streamer.h
```diff
@@ -142,6 +144,10 @@ class MotomanJointTrajectoryStreamer : public JointTrajectoryStreamer
 
   static bool VectorToJointData(const std::vector<double> &vec,
                                 industrial::joint_data::JointData &joints);
+                                
+  double time_of_last;
+  double time_since_last;
```
gavanderhoorn: Suggestion: rename these to include "of what" they are the "time of last". Right now these variable names say nothing about what this is used for, and without the context of them being used by the new point streaming code, it's rather difficult to understand what they are doing here.

There are also no comments associated with these lines.

motoman_driver/include/motoman_driver/joint_trajectory_streamer.h
```diff
@@ -142,6 +144,10 @@ class MotomanJointTrajectoryStreamer : public JointTrajectoryStreamer
 
   static bool VectorToJointData(const std::vector<double> &vec,
                                 industrial::joint_data::JointData &joints);
+                                
+  double time_of_last;
+  double time_since_last;
+  static const double point_streaming_timeout = 3.0;
```
gavanderhoorn: Please add `// seconds`.

motoman_driver/include/motoman_driver/industrial_robot_client/joint_trajectory_streamer.h
```diff
@@ -129,11 +132,13 @@ class JointTrajectoryStreamer : public JointTrajectoryInterface
 
   boost::thread* streaming_thread_;
   boost::mutex mutex_;
-  int current_point_;
+  int current_point_, streaming_sequence;
```
gavanderhoorn: is this missing a trailing underscore?

Additional suggestion: what is this a sequence number of / for?

gavanderhoorn:
I've added a few more comments (apologies, I should'v done a review instead).

Overall question: does this support multi-group systems?

tdl-ua:
Thanks for all the revisions @gavanderhoorn. I will revise the code accordingly and add a commit. 

As for multi-group support; I don't know. Am not too familiar with how the driver handles multiple manipulators. But @jim-rees comments the following in PR #88: "I've had a report that this does not work with dual arm robots, because they require motoman_msgs/DynamicJointTrajectory messages which this doesn't implement."

gavanderhoorn:
Thanks for iterating on this @tdl-tbslab. Much appreciated.

re: multi-group: yes, the [DynamicJointPoint](http://docs.ros.org/kinetic/api/motoman_msgs/html/msg/DynamicJointPoint.html) would be needed. It should not be too hard to add, but we might not want to complicate things in this PR.

tdl-ua:
I am assuming that multi-group refers to multi-arm setups such as the SDA5. Unfortunately, we do not have such a dual arm setup in our lab, so I would probably not be able to implement that feature.

motoman_driver/src/industrial_robot_client/joint_trajectory_streamer.cpp
tdl-ua: The function ``select(.)`` internally checks whether any element of ``msg->joint_names`` is empty or whether a matching "ROS element" can be found for the name of each joint in the message. If either of those two conditions are false, the function returns ``false``. So I think that would be the check for all required joints being present.

gavanderhoorn:
> I am assuming that multi-group refers to multi-arm setups such as the SDA5.

not necessarily. Work cells with a single arm on a linear track, or an arm with a 2-axis positioner can also be setup as multi-group systems.

The SDA is just a clever way to use this to create a 'humanoid-like' torso + arms setup.

But as I wrote above: let's not complicate this PR.

shaun-edwards:
@gavanderhoorn, is this good to go?

tdl-ua:
ping @gavanderhoorn

dambrosio:
All,
I am willing to update this PR with all the latest kinetic-devel commits and test on our GP7. I did have a few additions that I think we should add (see comments in PR commits).

Thoughts on trying to get this tested, updated with latest `kinetic_devel` and the least confusing way to update this PR?

motoman_driver/include/motoman_driver/joint_trajectory_streamer.h
```diff
@@ -143,6 +145,11 @@ class MotomanJointTrajectoryStreamer : public JointTrajectoryStreamer
   static bool VectorToJointData(const std::vector<double> &vec,
                                 industrial::joint_data::JointData &joints);
 
+  // variables for point streaming
+  double time_ptstreaming_last_point_;  // time at which the last point was received
+  double dt_ptstreaming_points_;        // elapsed time between two received points
+  static const double ptstreaming_timeout_ = 3.0; // seconds
```
dambrosio, line 151: Should we make this a parameter that defaults to 3.0 seconds?

motoman_driver/src/industrial_robot_client/joint_trajectory_streamer.cpp
```diff
@@ -159,6 +159,73 @@ void JointTrajectoryStreamer::jointTrajectoryCB(const trajectory_msgs::JointTraj
...
+  ROS_DEBUG("Current state is: %d", state);
+
+  //If current state is idle, set to POINT_STREAMING
+  if (TransferStates::IDLE == state)
+  {
+    this->mutex_.lock();
+    this->state_ = TransferStates::POINT_STREAMING;
+    this->current_point_ = 0;
+    this->ptstreaming_seq_count_ = 0;
+    this->streaming_start_ = ros::Time::now();
```
dambrosio: Reset / clear the queue to prevent sending old trajectory points:
```c++
this->ptstreaming_queue_ = std::queue<SimpleMessage>();
```

motoman_driver/src/industrial_robot_client/joint_trajectory_streamer.cpp
```diff
@@ -159,6 +159,73 @@ void JointTrajectoryStreamer::jointTrajectoryCB(const trajectory_msgs::JointTraj
...
+    ROS_INFO("First joint point received. Starting on-the-fly streaming.");
+  }
+
+  //if current state is POINT_STREAMING, process incoming point.
+  if (TransferStates::POINT_STREAMING == state)
+  {
+    if (msg->points.empty())
+    {
+      ROS_INFO("Empty point received, cancelling current trajectory");
+      return;
```
dambrosio: Force a transition to the IDLE state if we get a message with no-points, this functionality will allow users publishing to the `joint_command` topic to "instantly" stop POINT_STREAMING without waiting for timeout.
```c++ 
this->mutex_.lock();
trajectoryStop();
this->mutex_.unlock();
```

tdl-ua:
Hi @dambrosio, thanks for the reviews. I'll have a look at them later tonight and might have time to take care of this on Wednesday including the `kinetic-devel` update and incorporating the suggested modifications. I'll then add a commit. I think that should be the least messy way of adding your modifications to this PR.

dambrosio:
> Hi @dambrosio, thanks for the reviews. I'll have a look at them later tonight and might have time to take care of this on Wednesday including the `kinetic-devel` update and incorporating the suggested modifications. I'll then add a commit. I think that should be the least messy way of adding your modifications to this PR.

Sounds good. I will be testing this on a GP7 this afternoon and will let you know how things go.

dambrosio:
> > Hi @dambrosio, thanks for the reviews. I'll have a look at them later tonight and might have time to take care of this on Wednesday including the `kinetic-devel` update and incorporating the suggested modifications. I'll then add a commit. I think that should be the least messy way of adding your modifications to this PR.
> 
> Sounds good. I will be testing this on a GP7 this afternoon and will let you know how things go.

Quick update: I tried these changes yesterday and ran into issues with MotoRos reply messages in response to sending a single trajectory point. Here is the ROS console output
```
[ WARN] [1549936100.819839659]: Motoman robot is now enabled and will accept motion commands.
[ INFO] [1549936105.338233468]: First joint point received. Starting on-the-fly streaming.
[ WARN] [1549936105.410549892]: Reply String: Invalid message, Sub string: Trajectory start position doesn't match current robot position
[ERROR] [1549936105.410620147]: Aborting point stream operation.  Failed to send point (#0): Invalid message (3) : Trajectory start position doesn't match current robot position (3011)
```
The `trajectory_msgs/JointTrajectory` message contains a **single** `JointTrajectoryPoint` that has different joint positions from those being reported on the `joint_states` topic.

I was thinking of updating the POINT_STREAMING mode to process **two** trajectory points in the `jointCommandCB()`, where the first is the current robot joint positions and the second is the desired joint positions. I realized that these suggested updates would almost replicate the functionality of what is provided in the STREAMING mode?

_Update_: Seems like the original PR was tested using [this](motoman_jogger/src/motoman_jogger.cpp) node, I'll try to run with it and see if I get similar errors to what I described above. 

_Update 2/14/19_: Yesterday I managed to get point streaming to work on our GP7 manipulator. The jogger node used two very important (yet masked) steps in getting the streaming points successfully propagated through MotoROS. 

1. The first trajectory point (let's call this the *start point*) must contain the current joint positions with all zero velocity values
2. Each following points must contain non-zero velocity values for each position and the `time_from_start` value must be greater than the previous point.

Item 2 is extremely important. I believe the *start point* should have a `time_from_start` equal to zero and each point should increase from there. For example: start pt - 0.0 [s],  pt 1 - 0.02 [s], pt2 - 0.04 [s], etc. The time expressed in the example is just for reference and not actual values.

After minimal testing it was found that point streaming can successfully operate around 50 Hz max. If the user is not careful and sends `joint_command` messages faster than this the internal point streaming queue can grow without bounds. It is probably a good idea to check for a maximum queue size and either drop messages that would cause the queue to grow past this maximum or to error and transition back to IDLE.

@tdl-ua I have added checks to help the user follow these requirements when using point streaming mode, I will push to my forked branch and give you a link. If you would like to use any changes I have made please feel free.

motoman_driver/src/industrial_robot_client/joint_trajectory_interface.cpp
```diff
@@ -126,6 +126,8 @@ bool JointTrajectoryInterface::init(SmplMsgConnection* connection, const std::ve
   this->sub_cur_pos_ = this->node_.subscribe(
                          "joint_states", 1, &JointTrajectoryInterface::jointStateCB, this);
 
+  this->sub_joint_command_ = this->node_.subscribe("joint_command", 0, &JointTrajectoryInterface::jointCommandCB, this);
```
dambrosio, line 129: We should just double check that an infinite queue size makes sense for point streaming. Should we drop commands if the streaming node can cycle fast enough, or should we continue to fill up the command queue?

motoman_driver/src/industrial_robot_client/joint_trajectory_streamer.cpp
```diff
@@ -159,6 +159,73 @@ void JointTrajectoryStreamer::jointTrajectoryCB(const trajectory_msgs::JointTraj
...
+{
+  // read current state value (should be atomic)
+  int state = this->state_;
+
+  ROS_DEBUG("Current state is: %d", state);
+
+  //If current state is idle, set to POINT_STREAMING
+  if (TransferStates::IDLE == state)
+  {
+    this->mutex_.lock();
```
dambrosio, line 226: I think we could get in an undesired state transition if we get a new msg with empty points and we are we are currently in IDLE. Maybe we should check if the `msg->points.empty()` and if true report a warning, and return.

motoman_driver/src/industrial_robot_client/joint_trajectory_streamer.cpp
dambrosio: > The function `select(.)` internally checks whether any element of `msg->joint_names` is empty or whether a matching "ROS element" can be found for the name of each joint in the message. If either of those two conditions are false, the function returns `false`. So I think that would be the check for all required joints being present.

I agree with you on this. I think checking the return on select is adequate.

tdl-ua:
> All,
> I am willing to update this PR with all the latest kinetic-devel commits and test on our GP7. I did have a few additions that I think we should add (see comments in PR commits).
> 
> Thoughts on trying to get this tested, updated with latest `kinetic_devel` and the least confusing way to update this PR?

@dambrosio I've added you as a collaborator to the tbs-ualberta fork now so that you can push commits to the PR directly as I currently don't have time to work on this. Feel free to merge point-streaming with `kinetic-devel` and add your changes. Once I have the chance, I'll also test the updated PR on our SIA5.

dambrosio:
> > All,
> > I am willing to update this PR with all the latest kinetic-devel commits and test on our GP7. I did have a few additions that I think we should add (see comments in PR commits).
> > Thoughts on trying to get this tested, updated with latest `kinetic_devel` and the least confusing way to update this PR?
> 
> @dambrosio I've added you as a collaborator to the tbs-ualberta fork now so that you can push commits to the PR directly as I currently don't have time to work on this. Feel free to merge point-streaming with `kinetic-devel` and add your changes. Once I have the chance, I'll also test the updated PR on our SIA5.

I'll be able to take a look at all of this next week. I will merge and add any of my changes for review. If I have time I will try and provide a mechanism you can use to do tool tip velocity control using the new POINT_STREAM state.

tdl-ua:
> > > All,
> > > I am willing to update this PR with all the latest kinetic-devel commits and test on our GP7. I did have a few additions that I think we should add (see comments in PR commits).
> > > Thoughts on trying to get this tested, updated with latest `kinetic_devel` and the least confusing way to update this PR?
> > 
> > 
> > @dambrosio I've added you as a collaborator to the tbs-ualberta fork now so that you can push commits to the PR directly as I currently don't have time to work on this. Feel free to merge point-streaming with `kinetic-devel` and add your changes. Once I have the chance, I'll also test the updated PR on our SIA5.
> 
> I'll be able to take a look at all of this next week. I will merge and add any of my changes for review. If I have time I will try and provide a mechanism you can use to do tool tip velocity control using the new POINT_STREAM state.

I've developed something in the past that does essentially velocity control in cartesian space (basically end-effector teleoperation): [https://github.com/tbs-ualberta/manipulator_pose_following](https://github.com/tbs-ualberta/manipulator_pose_following) where I think I've addressed some of the issues you mention above. 

The package is a work in progress as obtaining the IK still needs to be improved and the code needs to be made more manipulator-agnostic but it is working.

motoman_driver/src/industrial_robot_client/joint_trajectory_streamer.cpp
dambrosio: This should be handled above on line 185.

dambrosio:
@tdl-ua quick FYI, this branch should be good for testing.

tdl-ua:
> 
> 
> @tdl-ua quick FYI, this branch should be good for testing.

@dambrosio I'll see if I can test it tonight.

motoman_driver/src/industrial_robot_client/joint_trajectory_streamer.cpp
tdl-ua: This would be a good addition I think.

motoman_driver/include/motoman_driver/joint_trajectory_streamer.h
tdl-ua: We can but I guess it is already covered by "Force a transition to the IDLE state if we get a message with no-points" right?

tdl-ua:
@dambrosio I've finally had a chance to test your commits on our SIA5. I tried to jog the EEF in cartesian space and did not run into any issues. 

I saw that you added checks on streaming queue bounds and `time_from_start` increase. I would agree that those checks are important to help the user follow the point streaming requirements. We can also calculate the frequency of incoming points from the difference between the `time_from_start` of two subsequent points and stop streaming with an error message if the frequency exceeds 50 Hz. The feedback to the user would then be much more informative.

jim-rees:
Please see my comments about latency and jitter at #252 .

tall1:
Hi,
Following [#555](https://github.com/ros-industrial/motoman/discussions/555) - I want to use this PR as @gavanderhoorn suggested it might work.

I have an HC10DTP + YRC1000micro controller.
We're currently on ROS noetic (Ubuntu 20.04).

Will this work? What adjustments do I need to make in order to enable point streaming with my controller?

DanManN:
At Rutgers we have a sda10f robot (one of the dual arm manipulators) and I maintain a [custom fork](https://github.com/DanManN/motoman) of this repo, mostly containing custom configurations for our experimental setup. Based on this pull request, however, I was able to recently implement the point streaming functionality. I don't think my code modifications are production ready by any means but I wanted to advertise them here for anyone who was having trouble getting point streaming working for their system. I would also love to help merge this functionality upstream but I'm not sure how to test code compatibility with other controllers and robot systems.

Issue #91, opened 2016-05-17
driver: assertion in joint_relay_handler when controller has < 4 motion groups configured


gavanderhoorn:
As reported in #89: the code hits the following assertion quickly after connecting to a controller:

``` c++
[FATAL] [1463490740.608713916]: ASSERTION FAILED
    file = /path/to/src/motoman/motoman_driver/src/industrial_robot_client/joint_relay_handler.cpp
    line = 322
    cond = all_joint_state.positions.size() == all_joint_names.size()
```

Observed on a system with an FS100 with two motion groups:
1. SIA20
2. linear track

`motoman_driver` based on the code in #89.

gavanderhoorn:
I'll edit later to add some more info, but see https://github.com/jettan/motoman/commit/5e6a3747e6673bc0f072b3e9a8c6f3f8f67b5cd6 (the changes to `motoman_utils.(h|cpp)`) for a commit that works around the issue for now.

jettan:
Alright, so here's what is happening under the hood:
Suppose that we have 2 motion groups. The current driver processes the `JointFeedbackMessage` from the simple messages in reverse order (meaning group 3 -> 2 -> 1 ->0 if we have 4 groups).
But if we only have 2 groups, [simple_message/joint_feedback_ex.cpp:120](https://github.com/jettan/motoman/blob/1f6ccae517a4e4e88ab7f6511417e312c50689b7/motoman_driver/src/simple_message/joint_feedback_ex.cpp#L120) will only try to get the first two `JointFeedbackMessage`, which will result in processing the two invalid groups (since simple messages are of a fixed length). 

Therefore, `convert_message(..)` in [JointFeedbackExRelayHandler](https://github.com/jettan/motoman/blob/7ab584221d10f4e9e07ed7624a3404de879276b2/motoman_driver/src/industrial_robot_client/joint_feedback_ex_relay_handler.cpp#L162) will get a lot of trash as data to process eventually leading to invalid `all_joint_states` being created after the `convert_messages(..)`, resulting in the assertion error reported by @gavanderhoorn

gavanderhoorn:
> Therefore, the `create_messages` [..] will get a lot of trash as data to process eventually leading to invalid all_joint_states being created after the convert_messages

Not really 'trash', an all-zeros `JOINT_FEEDBACK` message (which is essentially a valid message).

[...]

PR #225, opened 2018-06-22
driver: fix deserialisation of JointFeedbackEx

gavanderhoorn:
See comment on c9c62dd for more info.

Summary: always deserialise every submsg, but only keep those that contain valid information.

I've verified this on our SIA20F, which has two motion groups: robot itself + a linear rail.

Instead of 'upgrading' to `unloadFront(..)` (as suggested by @andreaskoepf), I decided to make this a minimal change and kept `unload(..)`.

### DIFF
--- a/motoman_driver/src/simple_message/joint_feedback_ex.cpp
+++ b/motoman_driver/src/simple_message/joint_feedback_ex.cpp
@@ -117,23 +117,40 @@ bool JointFeedbackEx::unload(industrial::byte_array::ByteArray *buffer)
     return false;
   }
 
-  for (int i = 0; i < this->groups_number_; i++)
+  // deserialise all JointFeedback submsgs contained in the buffer, going from
+  // back to front (ie: start with the last and end with the first). But only
+  // retain deserialised msgs that actually contain valid data.
+  //
+  // Note: we cannot assume that there is a 1-to-1 mapping between the order of
+  // JointFeedback msgs in the buffer and motion groups on the controller.
+  // Because of that we have to deserialise all submsgs and check validity
+  // of each individually (ie: we cannot skip submsgs 3 & 4 if there are only
+  // two motion groups, as the data for grp1 could be in submsg 3 fi).
+  for (std::size_t i = 0; i < MAX_NUM_GROUPS; ++i)
   {
     JointFeedbackMessage tmp_msg;
     JointFeedback j_feedback;
-
-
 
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
This is a fix for #91.

Partially based on the work by @jettan and @alemme (via @andreaskoepf).

gavanderhoorn:
@shaun-edwards: would you have time for a review? If not, I'll ask someone else.

gavanderhoorn:
@alemme, @andreaskoepf: would you perhaps see a chance to test this on your SDA? You would only need to change [this file](https://github.com/Xamla/motoman/blob/dbe5929cfd0d1acb8e911c94d692d47ee7359b65/motoman_driver/src/simple_message/joint_feedback_ex.cpp#L124-L125), right?

alemme:
Hi everyone. I successfully tested your changes on our sda10d system.

gavanderhoorn:
Thanks @alemme. Appreciated.

gavanderhoorn:
@sinaiaranda-CIDESI: the assertion you're running into is most likely the same as in #91. This PR should fix that one.

There are some more issues with your setup and `motoman_driver` in `kinetic-devel` though, so things won't directly work after this is merged, but it's a step.

I don't completely understand how, when using #179 things don't work for you though. But please open a separate issue and we can continue there.

shaun-edwards (approved):

gavanderhoorn:
Thanks @shaun-edwards for the review. And thanks @alemme as well.

SinaiAranda:
@gavanderhoorn The problem has been resolved with [this](https://github.com/Jmeyer1292/motoman/blob/9e7b50b860c47934011909d5854ea1d38dde6858/motoman_driver/src/simple_message/joint_feedback_ex.cpp#L110-L151) modification of [PR](https://github.com/ros-industrial/motoman/issues/91).

gavanderhoorn:
I'm slightly confused: #91 is the issue that reported the problem, not a PR. And the commit you link should be the "hacky" way this was fixed, with the PR we're discussing here being a more appropriate way.

If I may suggest: use the fix in this PR or just use current `kinetic-devel`, which contains the fix.

Fixes for #226 should fix the remaining issues and make things operational again (note: those haven't been merged yet).

SinaiAranda:
@gavanderhoorn Excellent, have do not problem.

Issue #220, opened 2018-06-19
wiki: update links to motoros binaries on installation tutorial page


gavanderhoorn:
From https://github.com/ros-industrial/motoman/issues/196#issuecomment-398357434:

> @ted-miller: we'll need to remember updating the links on the wiki on the [installation tutorial](http://wiki.ros.org/motoman_driver/Tutorials/indigo/InstallServer).

ted-miller:
Done

gavanderhoorn:
Awesome, thanks.

Issue #219, opened 2018-06-03
Implications of high Command - Feedback latency


andreaskoepf:
In order to evaluate the realtime control capabilities of a SDA10D (DX100) with the Motoman ROS driver we conducted a couple ot realtime control experiments at Xamla. We modified the MotoPlus part of the driver to report not only joint-states but also additional values like the command position. We found that our DX100 with the `mpMeiIncrementMove()` command has a relative high latency for executing commands of at least 150ms which cannot be explained by the physical limitation of the motors. We think a processing part like a moving average smoothing filter in the Yaskawa motor controllers could explain the observed behavior, e.g. Eric Macil of Yaskawa America writes in a [forum post regarding CMD-FB latency](https://groups.google.com/d/msg/swri-ros-pkg-dev/adT8nuTRQ4o/7KllzlwzpEUJ) "That command goes through an averaging filter for motion smoothness and then PID control loop command the motion to reduce the error between the feedback position and command position.".

We asked Yaskawa whether it would be possible to reduce the system latency but we learned that it is not possible to reduce the latency significantly and that even a minor possible reductions would result in a complete loss of warranty. Although the CMD-delay is a significant problem for ROS based realtime control loops, like force-torque control, we managed to get this working to some extend with work-arounds like look-ahead. Some videos of FT behaviour that we managed to get working are: [Admittance Control](https://youtu.be/wV2XHSPId5c) and [Impedance Control](https://www.youtube.com/watch?v=5D72o1vfU34)

I would like to share some plots that we recorded about a year ago:

## 1. Single Joint Sinus-Trajectory

![sin_joint7_ft](https://user-images.githubusercontent.com/9976399/40885553-dc06720e-6728-11e8-93be-eefe34793725.png)

In this plot the yellow `q_cmd` line represents the position that has been commanded, meaning the sample was generated after the command had been processed by `mpMeiIncrementMove()`. The green `q_fb` line is the joint feedback reported by the Motoman driver (which is normally published on the `/joint_states` topic). You see a clear phase shift of >100ms between yellow and green lines.

The light blue `forces_abs` line is the before zeroed norm reading of a ATI gamma force torque sensor. The temporal correlation between measured absolute force (~abs acceleration) and joint-feedback indicate that the joint-feedback is probably the actual position of the robot reported only with minimal latency, e.g. directly read out of the joint-encoders. 

## 2. Force-Torque based Move Until Contact

[...]

PR #227, opened 2018-06-26
MotoROS v1.8.1

gavanderhoorn:
See commit comments.


gavanderhoorn:
@EricMarcil: fyi.

gavanderhoorn:
Note that we'll need to update the links on the wiki again after this gets merged (because of the new binaries). See #220.

motoman_driver/MotoPlus/CtrlGroup.c
```diff
@@ -362,6 +364,21 @@ BOOL Ros_CtrlGroup_GetFBServoSpeed(CtrlGroup* ctrlGroup, long pulseSpeed[MAX_PUL
...
+	{
+		FB_AXIS_CORRECTION *corr = &ctrlGroup->correctionData.correction[i];
+		if (corr->bValid)
+		{
+			int src_axis = corr->ulSourceAxis;
+			int dest_axis = corr->ulCorrectionAxis;
+			pulseSpeed[dest_axis] -= (int)(pulseSpeed[src_axis] * corr->fCorrectionRatio);
+		}
+	}
+
```
gavanderhoorn, line 381: Could this `for` be merged with the one directly above it? Would avoid iterating over the `pulseSpeed` array again.

motoman_driver/MotoPlus/Controller.c
```diff
@@ -640,6 +640,50 @@ int Ros_Controller_GetNotReadySubcode(Controller* controller)
...
+	else if (bDataInQ == ERROR)
+		return ERROR;
+	else
+	{
+		//for each control group
+		for (groupNo = 0; groupNo < controller->numGroup; groupNo++)
+		{
+			//Check group number valid
+			if (!Ros_Controller_IsValidGroupNo(controller, groupNo))
+				continue;
```
gavanderhoorn, line 665: Just checking: if all `groupNo` are invalid, it immediately means that this function returns `FALSE`. I'm not sure a robot with all groups being invalid can be in motion, but this is on purpose, correct?

motoman_driver/MotoPlus/CtrlGroup.c
```diff
@@ -455,6 +457,8 @@ BOOL Ros_CtrlGroup_GetTorque(CtrlGroup* ctrlGroup, double torqueValues[MAX_PULSE
 	memset(dst_trq.data, 0, sizeof(MP_TRQCTL_DATA));
 	dst_trq.unit = TRQ_NEWTON_METER; //request data in Nm
 
+	memset(&dst_vel, 0x00, sizeof(MP_GRP_AXES_T));
```
gavanderhoorn, line 460: style: the `memset(..)` above this one uses a plain, decimal, `0`.

motoman_driver/MotoPlus/IoServer.c
```diff
@@ -97,6 +97,7 @@ void Ros_IoServer_StartNewConnection(Controller* controller, int sd)
 			controller->sdIoConnections[connectionIndex] = INVALID_SOCKET;
 			controller->tidIoConnections[connectionIndex] = INVALID_TASK;
 			Ros_Controller_SetIOState(IO_FEEDBACK_FAILURE, TRUE);
+			mpSetAlarm(8004, "MOTOROS FAILED TO CREATE TASK", 7);
```
gavanderhoorn, line 100: Is the `7` a nr with a special significance? Fi, `7` indicates the IoServer task could not be created?

If so: should we create constants for them, to avoid the magic nrs?

gavanderhoorn: And the same comment for the other `mpSetAlarm(..)` invocations for task creation failure.

motoman_driver/MotoPlus/MotionServer.c
```diff
@@ -1552,7 +1556,7 @@ int Ros_MotionServer_GetQueueCnt(Controller* controller, int groupNo)
 	}
 		
 	printf("ERROR: Unable to access queue count.  Queue is locked up!\r\n");
-	return -1;
+	return ERROR;
```
gavanderhoorn, line 1559: :+1:

motoman_driver/MotoPlus/MotionServer.c
```diff
@@ -1563,11 +1567,15 @@ int Ros_MotionServer_GetQueueCnt(Controller* controller, int groupNo)
...
 	int groupNo;
+	int qCnt;
 	
 	for(groupNo=0; groupNo<controller->numGroup; groupNo++)
 	{
-		if(Ros_MotionServer_GetQueueCnt(controller, groupNo) > 0)
+		qCnt = Ros_MotionServer_GetQueueCnt(controller, groupNo);
+		if (qCnt > 0)
 			return TRUE;
+		else if (qCnt == ERROR)
```
gavanderhoorn, line 1577: This could be just an `if`, right (because of the `return`s)? Seeing an `else if` in isolation without an `else` always looks weird to me.

gavanderhoorn (commented):
Haven't tested the new binaries yet, but here are some comments.

gavanderhoorn:
Checked this on our SIA20F+linear rail.

This PR definitely improves the `in_motion` situation. `in_motion` stays `true` for the majority of the motion, and then goes to `false` three (3) `JointFeedback` messages before the robot has come to a complete stop. Joint velocity is also `> 0` for the three `JointFeedback` messages after `in_motion` has changed to `false`.

I haven't checked, but that could well be within the `START_MAX_PULSE_DEVIATION` that @EricMarcil mentions in https://github.com/ros-industrial/motoman/issues/219#issuecomment-400314030 (robot was moving rather slowly along the rail, and 3 msgs at 50 Hz is 75ms).

gavanderhoorn:
@EricMarcil: I've added some comments.

motoman_driver/MotoPlus/CtrlGroup.c
EricMarcil: If you look closely, the axis correction makes reference to another source axis pulseSpeed.   Granted that the source axis is normally a previous axis but there is a possibility that it wouldn't.  So it is safer to calculate the pulseSpeed for all axes before doing the correction.

motoman_driver/MotoPlus/Controller.c
EricMarcil: That is correct and intentional.  If all groups are invalid, there really nothing to move.

motoman_driver/MotoPlus/IoServer.c
EricMarcil: Yes, I've given each mpSetAlarm a different subcode (pretty much in the order they are created), so that if the alarm 8004 occurs, it could help identify which task wasn't started.  Generally speaking, I doubt that we would dig down to that level because whenever that 8004 occurs, it's probably because of parameters settings or you have multiple versions of the MotoRos running.  So I didn't go as far as creating constants for them.

motoman_driver/MotoPlus/MotionServer.c
EricMarcil: Yes, it could.  That change was made by @ted-miller.  Personally, I tend to build the if, else if, else structure based on the conditions that is to be consider (in this case the qCnt value).  It's independently of what occurs after the condition is identified.  So I tend to agree with the way it is.

gavanderhoorn:
@alemme / @andreaskoepf: do you see any chance of testing fe26038 perhaps on your SDA?

It would seem to address at least some of the comments by @andreaskoepf in #219.

andreaskoepf:
@EricMarcil Great to see some improvements here!

@gavanderhoorn We will defiintely test this during the next two weeks. (Currently we are in final preparations for Rosvita v0 release (which hopefully will be the best or one of the best ROS GUIs for controlling the SDA). I plan to sync our Motoman driver with this repo in the July timeframe.)

gavanderhoorn:
@andreaskoepf wrote:
> We will defiintely test this during the next two weeks. (Currently we are in final preparations for Rosvita v0 release (which hopefully will be the best or one of the best ROS GUIs for controlling the SDA). I plan to sync our Motoman driver with this repo in the July timeframe.)

Ok, good.

In the meantime I'll go ahead and merge this, as it solves some of the other issues we have with goals being flagged as completed too early.

If anything pops up during your testing please open a new issue.

gavanderhoorn:
I'm not in the habit of doing this btw, but the test in this case would be in addition to the verification that has already been done by me and @ted-miller and @EricMarcil (at Motoman), and is also more a test to see whether the situation that you described in #219 has improved.

gavanderhoorn:
@EricMarcil: as this PR was opened by me, I need one other maintainer to review this PR and approve it. Could you do that?

EricMarcil (approved):

gavanderhoorn:
Thanks for the review @EricMarcil :+1:

PR #234, opened 2018-07-31
SIA robots: fix ros-i base frame parent.

gavanderhoorn:
Not sure why I attached `base` to `link_l` in #121, as it's clearly incorrect: world frame is located at the same location as `link_l`, but should not be attached to it, as that will cause it to rotate with `link_l`.

Instead, we attach it to `base_link` now at the same `z` as `link_l`.

This now corresponds to how the `base` frame is defined in the macros introduced in #212 fi.


gavanderhoorn:
Together with a local fix for #105 this leads to almost correct `base->tool0` transforms for the SIA20 we have here in the lab (almost correct, as the robot probably has been calibrated and the xacro macro does not take that into account).

gavanderhoorn:
@shaun-edwards: :bellhop_bell:

gavanderhoorn:
Would be great if we could merge this one as well @shaun-edwards.

gavanderhoorn:
Ping @shaun-edwards.

shaun-edwards (approved):

PR #239, opened 2018-08-15
Advertise robot_{enable/disable} services without '/' prefix

### LINKED ISSUE #218: Why is the root namespace specified for "/robot_enable" & "/robot_disable"?
In `MotomanJointTrajectoryStreamer::init()` the handy `robot_enable` and `robot_disable` services are advertised, see [here](https://github.com/ros-industrial/motoman/blob/808a8c16d9120814c11438b8809cfb957bf3a70f/motoman_driver/src/joint_trajectory_streamer.cpp#L89-L91). Unfortunately the service name during the registration contains a '/' at the beginning and therefore overrides any namespace defined in a `<group>` element in a launch file, e.g. in Rosvita the launch-file generated for motoman driver looks like this:

```
<launch>
  <group ns="sda10d">
    <param name="robot_ip_address" type="str" value="10.0.0.1" />
    <node name="joint_state" pkg="motoman_driver" type="robot_state" />
    <node name="motion_streaming_interface" pkg="motoman_driver" type="motion_streaming_interface" />
    <node name="joint_trajectory_action" pkg="motoman_driver" type="motoman_driver_joint_trajectory_action" />
  </group>
</launch>
```

e.g. we use namespaces in order to allow multiple robots to be configured at once. The motoman actions and the `stop_motion` service are correctly placed under the specified namespace but robot_enable and robot_disable stay at the root namespace. Was the decision for the root-ns made deliberately (otherwise I could create a PR)?

andreaskoepf:
fixes #218

### DIFF
--- a/motoman_driver/src/joint_trajectory_streamer.cpp
+++ b/motoman_driver/src/joint_trajectory_streamer.cpp
@@ -86,9 +86,9 @@ bool MotomanJointTrajectoryStreamer::init(SmplMsgConnection* connection, const s
     motion_ctrl_map_[robot_id] = motion_ctrl;
   }
 
-  disabler_ = node_.advertiseService("/robot_disable", &MotomanJointTrajectoryStreamer::disableRobotCB, this);
+  disabler_ = node_.advertiseService("robot_disable", &MotomanJointTrajectoryStreamer::disableRobotCB, this);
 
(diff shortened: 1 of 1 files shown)


gavanderhoorn (approved):
Thanks for fixing this @andreaskoepf.

gavanderhoorn:
Thanks again.

PR #246, opened 2018-09-18
Add missing dependency to libxmlrpc-c++

fsuarez6:
When installing the `motoman` metapackage dependencies manually and with the `--no-install-recommends` flag, `libxmlrpc-c++` doesn't get installed. This PR force the library to be installed.

Without `libxmlrpc-c++`, the build will fail somewhere here:
https://github.com/ros-industrial/motoman/blob/kinetic-devel/motoman_driver/src/industrial_robot_client/motoman_utils.cpp#L46


fsuarez6:
My bad. This seems to be an issue of the `industrial_utils` package.

PR #250, opened 2018-10-26
Added new MotoMINI robot support package

EricMarcil:
New support package for the YRC1000micro MotoMINI robot model.


EricMarcil:
Hi,
I added the MotoMINI model.  A couple of notes:
1. The package make reference to only "mini" as the model name and not "motomini" since the package is already refering to motoman, hence: "motoman_mini"
2. The MotoMINI base extends below the bolting surface, therefore the base coordinate system is inside the base model, at the bottom of the bolting surface in the center of the bolt pattern.

gavanderhoorn:
thanks for the PR.

I've added some commits to address some issues with it.

>     2. The MotoMINI base extends below the bolting surface, therefore the base coordinate system is inside the base model, at the bottom of the bolting surface in the center of the bolt pattern.

I'm not entirely sure I follow you here, but it would be nice if we could make `base_link` be at the bottom of the base mesh.

See [Coordinate Frames for Serial Industrial Manipulators: Coordinate Frames](http://gavanderhoorn.github.io/rep/rep-xxxx.html#coordinate-frames):

> Note that this `base_link` does not need to be coincident with the origin of the controller's Cartesian frame, but rather it should be located in the logical base position of the robot (ie: its attachment to the world).
> 
> Discrepancies between a controller's internal Cartesian frame and REP 103 shall be resolved by defining the correct transform between base_link (or any other convenient frame) and the base frame.

Could we place `base_link` at the bottom of the `base_link.stl` mesh and then increase the `z` component of the `base_link->link_1` and `base_link->base` transform to compensate?

Alternatively, we could also introduce an extra link between `base_link` and `link_1` (with a `fixed` joint) for the 68 mm transform between the bottom of the 'case' and the location where you have `base_link` now.

gavanderhoorn:
I also just noticed there are quite a few gaps between the meshes of the consecutive links:

Between `base_link` and `link_1`:

rviz_screenshot_2018_11_01-12_14_21

Between `link_2` and `link_3`:

rviz_screenshot_2018_11_01-12_14_48

Between `link_3` and `link_4`:

rviz_screenshot_2018_11_01-12_15_26

Between `link_4` and `link_5`:

rviz_screenshot_2018_11_01-12_15_52

gavanderhoorn:
@EricMarcil wrote:
> The package make reference to only "mini" as the model name and not "motomini" since the package is already refering to motoman, hence: "motoman_mini"

if we want to be consistent and adhere to ROS-I naming rules it should actually be "motomini", as that is the name of the series/model.

Personally I find consistency to be a good thing.

Additionally, it's remote, but if Yaskawa ever releases a product called "mini" and we'd like to support that, we'd have a problem.

EricMarcil:
@gavanderhoorn : Thanks for the review and commits.  I tried to follow the existing patterns but I guess I missed a few details.  I'll try to improve on the next one...

For the MotoMini, the "base_link" coordinate doesn't correspond to the manufacturer coordinate system, which is represented by the "base" coordinate system.
I believe that the current position, in the center of the attachment point to the floor or mounting surface, is the correct one to use.  To mount the MotoMini, you basically need to make a rectangle in the mounting surface to insert part of the base below the mounting surface.  I feel that placing the coordinate system at the bottom of the model is not practical because the robot will never be mounted on a surface where that is the resting point.

For the gaps, to lighten the model, we removed the internal components (gearbox) and kept only the outter shell.  The gaps are rather small and I don't think they will affect the visual or practical representation of the model.

As for the name, that is a good point that in the future, it would be problematic if we ever release a "mini" product.  So I agree with renaming everything to "motomini".  
@gavanderhoorn: Do you want me to rename it, or can you do it?

gavanderhoorn:
> As for the name, that is a good point that in the future, it would be problematic if we ever release a "mini" product. So I agree with renaming everything to "motomini".
> @gavanderhoorn: Do you want me to rename it, or can you do it?

I'll rename.

gavanderhoorn:
> @gavanderhoorn : Thanks for the review and commits. I tried to follow the existing patterns but I guess I missed a few details. I'll try to improve on the next one...

No problem.

Some of the things I had to fix are not that obvious actually.

And most of the support pkgs in `ros-industrial/motoman` (and `motoman_experimental`) violate at least some of the current best practices and could do with some cleanup.

> For the MotoMini, the "base_link" coordinate doesn't correspond to the manufacturer coordinate system, which is represented by the "base" coordinate system.
> I believe that the current position, in the center of the attachment point to the floor or mounting surface, is the correct one to use. To mount the MotoMini, you basically need to make a rectangle in the mounting surface to insert part of the base below the mounting surface.

Ok. If the "hole in the floor" is indeed the recommended way to mount this robot, then we can leave `base_link` like this.

> For the gaps, to lighten the model, we removed the internal components (gearbox) and kept only the outter shell. The gaps are rather small and I don't think they will affect the visual or practical representation of the model.

The gaps just look strange, especially if you have a background that's lighter than the robot itself. But it's only an aesthetic issue.

gavanderhoorn:
If you can check you are ok with the updated PR, we could merge it @EricMarcil.

EricMarcil:
@gavanderhoorn: Everything is good.  I've updated with the latest commits and regenerated a MoveIt package without issues.

gavanderhoorn (approved):
This can be merged now.

There are a few things that could be fixed / improved, but for consistency we'll do that at the sametime as the other packages in this repository.

gavanderhoorn:
Thanks @EricMarcil for the contribution.

PR #256, opened 2018-11-16
MotoROS v1.8.2

ted-miller:
- Increase starting pulse deviation threshold.
- Add 4 hour limit to a single trajectory timestamp.
- Add message to retrieve DH parameters of arm.


ted-miller:
My robot is tied up with other testing right now. 

@EricMarcil or @gavanderhoorn 
If one of you is available next week to test this, I would appreciate the help. Otherwise, I will perform testing on Nov 26.

EricMarcil:
@gavanderhoorn : I've started doing some testing.  I've found a mistake that prevent the message from sending a proper response.  I'll review with Ted and let you know once we have that part working.

motoman_driver/MotoPlus/MpRosAllControllers.vcxproj
```diff
@@ -113,7 +113,7 @@
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='YRC1000|Win32'">
     <NMakeBuildCommandLine>mpbuilder.exe -c $(Configuration) -p "$(ProjectDir)\" -n "MotoRosYRC1_" -o build</NMakeBuildCommandLine>
-    <NMakeOutput>MotoRosYRC1_.out</NMakeOutput>
+    <NMakeOutput>$(MP_VS_Install)DevTools\OnlineDownload.exe</NMakeOutput>
```
gavanderhoorn, line 116: Is this correct? Seems like a strange name for the output binary?

ted-miller: Yes, this is correct. That field doesn't actually control the file name of the output binary (this isn't a true makefile project). This trick allows Visual Studio to launch the remote-upload utility when you build and attempt to "run" the application. So, after you build, you can easily push the updated binary into the robot controller without booting into maintenance mode. (This only works on YRC.)

gavanderhoorn (changes requested):
I cannot review this properly without my hw working, but I've got confirmed reports that it works and don't want to keep this from being merged any longer.

@ted-miller: could you please remove the `*.vcxprojAssemblyReference.cache` from the PR? I tried to do that for you, but it appears you've unchecked "allow edits from maintainers".

ted-miller:
I have made the change. BTW, "allow edits from maintainers" is actually checked.

gavanderhoorn:
> I have made the change.

thanks.

> BTW, "allow edits from maintainers" is actually checked.

hm, strange. Just tried to push an update to `.gitignore` but I'm seeing this:

```
 ! [remote rejected] tedmiller_kinetic-devel -> tedmiller_kinetic-devel (permission denied)
```

perhaps it's because you have created the PR from a branch you've protected?

gavanderhoorn (approved):
:+1:-ing these changes.

Further fixes and improvements will be made in follow-up PRs.

gavanderhoorn:
We'll wait for Travis and then merge.

gavanderhoorn:
Thanks for the PR @ted-miller and thanks for testing @EricMarcil.

gavanderhoorn:
Updated wiki links [here](http://wiki.ros.org/action/info/motoman_driver/Tutorials/indigo/InstallServer?action=diff&rev2=36&rev1=35).

Issue #260, opened 2019-02-09
motoman_driver compile warnings when using -Wall 


dambrosio:
Various compile warnings when compiling with `-DCMAKE_CXX_FLAGS=-Wall -DCMAKE_C_FLAGS=-Wall`.

Fix warnings related to:

- [x] Fix warnings related to: comparison between signed and unsigned integer expressions [-Wsign-compare]
  - [x] Update ROS console logs to use %lu instead of %d now that for loop index variable will be of type `size_t` instead of `int`

- [x] Fix warnings related to: no return statement in function returning non-void [-Wreturn-type]
  - [x] Return value should be based on each call to create_messages() inside for-loop, probably want to return false if any of the `create_messages` function call retturns false

I have fixes for these and can create a PR for these in the next day or so.

gavanderhoorn:
Closing as #388 was merged.

Issue #262, opened 2019-02-10
Add roslint_add_test() to motoman_driver and fix roslint errors


dambrosio:
`roslint_cpp()` is included in the `motoman_driver` package but has 2290 lint errors. The lint errors can be viewed by manually running `catkin_make roslint_motoman_driver`. These lint errors are masked because `roslint_add_test()` is not included in the CMakeLists file. 

I propose the following updates/fixes:

- [x] Fix all lint errors in the `motoman_driver` package 
- [x] Add `roslint_add_test()` to `if(CATKIN_ENABLE_TESTING)`
- [x] Verify roslint unit tests pass by running `catkin_make run_tests`

gavanderhoorn:
Closing as #387 was merged.

PR #264, opened 2019-02-11
Fixes ros-industrial/motoman#260: fix warnings with -Wall

dambrosio:
Changed for loop index variables to use size_t when comparing with vector size and corresponding ROS console output format string to %lu, JointFeedbackExRelayHandler create_messages now has return value


gavanderhoorn:
As in my comment to #262: could I ask you to please use separate commits for formatting changes / code changes and other changes? Especially the "JointFeedbackExRelayHandler create_messages now has return value" bit should absolutely be in a separate commit.

dambrosio:
> As in my comment to #262: could I ask you to please use separate commits for formatting changes / code changes and other changes? Especially the "JointFeedbackExRelayHandler create_messages now has return value" bit should absolutely be in a separate commit.

Separated into multiple commits. Let me know if this is what you had in mind.

gavanderhoorn:
Closing.

Rebased version in #388.

gavanderhoorn:
Thanks @dambrosio for your initial PR.

And apologies for merging it so late of course, that was not the intention.

Issue #232, opened 2018-07-26
sda10f: dimensions in URDF does not match specifications


hsd-dev:
While working on https://github.com/ros-industrial/ros_industrial_issues/issues/28 for motoman, I noticed that the dimensions in URDF for SDA10f differ from that in [specification: page 32](http://cdn2.hubspot.net/hubfs/366775/downloads/documentation/156402-1CD.pdf?t=1468601845515).

For eg: according to the document distance between vertical axes of arm_left_link_2_l and arm_right_link_2_l is 530mm but 

```
$ rosrun tf tf_echo /arm_left_link_2_l /arm_right_link_2_l
```
shows
```
At time 1532606439.014
- Translation: [0.000, -0.506, 0.000]
- Rotation: in Quaternion [0.000, -0.000, 1.000, -0.001]
            in RPY (radian) [-0.000, 0.000, -3.140]
            in RPY (degree) [-0.000, 0.000, -179.909]
```
i.e. 506mm. Similarly, arm_right_link_4_u and arm_right_link_2_l have an offset of 360mm 

```
$ rosrun tf tf_echo /arm_right_link_4_u /arm_right_link_2_l
```
shows
```
At time 1532606196.314
- Translation: [0.000, -0.350, 0.054]
- Rotation: in Quaternion [0.000, 0.000, 0.000, 1.000]
            in RPY (radian) [0.000, -0.000, 0.000]
            in RPY (degree) [0.000, -0.000, 0.000]
```
i.e. 350mm

This seems to be the case with other dimensions as well. What is the reference used for the URDF?

gavanderhoorn:
This package was introduced by @thiagodefreitas in 45bd3273. According to the package manifest added in [that commit](https://github.com/ros-industrial/motoman/commit/45bd3273ac7c4eafdc0c0415406e4f46b34b1d48#diff-f9e698b0c8991e87e5409a66df4a0dafR22), the reference is/was: `http://www.motoman.com/datasheets/sda10f.pdf`. No mention of which version or document nr though, and that link doesn't load a pdf any longer it appears.

Current version is [here](https://www.motoman.com/hubfs/Robots/SDA10F.pdf).

Note btw that for many urdfs in ROS (not just ROS-I), the inter-joint-origin distances do not always correspond to what the datasheet(s) list.

It's considered adequate if the total arm length (ie: from root to the last link) is correct.

[...]

PR #271, opened 2019-03-01
Fix sda10f urdf

asierfernandez:
We propose a more accurate urdf for sda10f robot. Updated dimensions have been extracted from [3D CAD model](https://www.yaskawa.eu.com/index.php?eID=dumpFile&t=f&f=876&token=f8bae09ca74b0f765920db794a1df3c24acae314) as described in [#232](https://github.com/ros-industrial/motoman/issues/232#issuecomment-468613990)


motoman_sda10f_support/urdf/sda10f_macro.xacro
```diff
@@ -6,13 +6,13 @@
...
 
-        
-        <xacro:motoman_arm name="${prefix}arm_left" prefix="${prefix}arm_left_" parent="${prefix}torso_link_b1" reflect="1">
-            <origin xyz="0.09996 0.0275 0.32214" rpy="1.57 0 0" />
-        </xacro:motoman_arm>      
-        
-        <xacro:motoman_arm name="${prefix}arm_right" prefix="${prefix}arm_right_" parent="${prefix}torso_link_b1" reflect="-1">
-            <origin xyz="0.09996 -0.0275 0.32214" rpy="1.57 0 0" />
+
+        <xacro:motoman_arm name="${prefix}arm_left" prefix="${prefix}arm_left_" parent="${prefix}torso_link_b1" reflect="1" joint_limited="false">
```
gavanderhoorn: Is the `joint_limited` a copy-pasta? It's not an argument supported by the `motoman_arm` macro afaict.

gavanderhoorn:
@asierfernandez: CI seems to fail for `motoman_sda10f_support` with the proposed change.

Can you check locally?

motoman_sda10f_support/urdf/sda10f_macro.xacro
asierfernandez: sorry I messed up with some other branches

asierfernandez:
I'll check CI next week

asierfernandez:
> @asierfernandez: CI seems to fail for `motoman_sda10f_support` with the proposed change.
> 
> Can you check locally?



> @asierfernandez: CI seems to fail for `motoman_sda10f_support` with the proposed change.
> 
> Can you check locally?

I'll check it next week

gavanderhoorn:
Looks like the `joint_limited` param was the problem.

motoman_sda10f_support/urdf/sda10f_macro.xacro
gavanderhoorn: So you have a `joint_limited` version of the urdf?

gavanderhoorn:
Getting back to this: the urdf is fixed, but the origins of the meshes are not updated in this PR, which leads to rendering issues (and potential problems with MoveIt's collision detection/avoidance).

Not asking @asierfernandez to fix that after all this time, just making a note about this here.

Example:

rviz_screenshot_2019_08_28-17_24_29

asierfernandez:
We didn't pay attention to meshes, because we were only interested on solving properly the IK.
Meshes should also be updated, but unfortunately we won't work on that in the next months.
@gavanderhoorn  thanks for the comment!

gavanderhoorn:
Just to make sure we don't accidentally merge this in, I'm going to submit a review with *request changes*.

Thanks for the additional info @asierfernandez.

gavanderhoorn (changes requested):
Mesh origins would need to be updated as well.

gavanderhoorn:
We now have #321, which I believe will address the issues raised here as well.

Thanks @asierfernandez for the initial PR :+1:

Issue #276, opened 2019-05-27
tool0 frame not at end of link_6_t in motoman_mh50_support


npd26:
Hello,

this is issue regarding my question on ROS answers regarding motoman_mh50_support package.

> https://answers.ros.org/question/324031/motoman_mh50_support-urdf-frame-location/?comment=324060#post-id-324060

why is there some gap? it should be exactly at the end, or?

Thanks.

---

Edit (by @gavanderhoorn): including the screenshot @npd26 included in his ROS Answers post here (to avoid it going missing):

![Screenshot-from-2019-05-26-15-47-21](https://user-images.githubusercontent.com/4550046/95187991-21fec680-07cc-11eb-9178-262f8a2850d9.jpg)

gavanderhoorn:
@EricMarcil: would you have any idea? Is this a (small) issue with the origin of the mesh for the T link?

EricMarcil:
It is an issue with the origin of the mesh but not just the T-axis.  It's the mesh of the whole upper arm that seem to be shifted backward by about 13 mm.  It's a bit hard to eyeball the proper correction.  I've request new model files from my design department and will try to generate proper mesh from them.

npd26:
can this cause problem to vision based pick of object?

gavanderhoorn:
I believe so: if you've attached your EEF to the urdf at that link then FK and IK would probably not work correctly.

npd26:
I have attached my realsense camera URDF directly to tool0. I detect object center (x,y) in camera_frame and then directly transforming it to base_link. since 2 week I am trying and tool is not reaching to correct position. can I consider this as one reason apart from calibration  when I write in my thesis?

gavanderhoorn:
Any vision process is going to have (quite) a bit of error estimating poses, but if your kinematic model is incorrect then anything using that model will also include that error.

> I detect object center (x,y) in camera_frame and then directly transforming it to base_link. 

I would recommend you use `base` instead as that would allow you to compare results to the coordinate frame of the robot / controller itself.

npd26:
ok I will check it. I hope we will get new support package quickly.

EricMarcil:
I don't agree.  The problem should be strictly visual.  If you are attaching your tool to the Tool0, then it should resolve properly.  As a reference, when the robot is at Home (all angles at 0) the Tool0 position should be: 
X: 1345  Y: 0  Z:1080 (or Z: 1620 relative to the ground level).

[...]

PR #280, opened 2019-06-14
MH50 meshes corrections

EricMarcil:
Correction of the MH50 meshes in relation to issue ros-industrial#276 "tool0 frame not at end of link_6_t in motoman_mh50_support"


gavanderhoorn:
@EricMarcil: quick question: if you only fixed `link_t`, why does the PR show all meshes have changed?

EricMarcil:
Actually the error was with the L-axis and then propagated all the way to the T-axis.
So yes, the Base and S-axis could have stayed the same but the other models had to be changed.  I didn't really think about it at the time and just redid everything.

gavanderhoorn (changes requested):
Apologies for taking so long.

I've *requested changes* as it seems the origin of `link_6_t` is not in the centre of the disk:

rviz_screenshot_2019_08_28-17_14_42

This is after rotating `joint_6_t`: notice how the last link is not covering `link_5_b` (the grey link, I've temporarily changed the colour).

EricMarcil:
I've made correction and pushes under Issue #276.  Please review when you have a chance.  Thanks.

gavanderhoorn (dismissed):

gavanderhoorn:
@EricMarcil: unfortunately it would appear the T link is still not correctly centred.

The easiest way to check would be to load the 'test' launch file:

```
roslaunch motoman_mh50_support test_mh50.launch
```

zoom in on the flange and then rotate `joint_6_t`. You'll notice the disk of the flange has a slight 'wobble'.

Is the source mesh not centred either? Is this something you have to do manually?

gavanderhoorn:
Replaced by #316.

PR #283, opened 2019-08-08
MotoROS driver v1.9.0

ted-miller:
Add support for HC robots.
Add error codes to detect PFL motion override.
Add command to modify active tool.
Embed setup-validation routine to prevent removal from source code.


gavanderhoorn:
I can do a full review later, but first some logistical issues:

 - there appear to be 8 commits in this PR where I'd expect 1
 - the `1.9.0` commit combines too many changes in a single commit. Please introduce at least the `motoman_hc_support` package in a separate commit. Ideally the rest of the changes would also be separate commits. That would make both reviewing and potential reverting/updating much easier.

ted-miller:
Moved to PR #284

PR #284, opened 2019-08-12
MotoROS Driver v1.9.0 with HC-10 support

ted-miller:
Add support for HC robots.
Add error codes to detect PFL motion override.
Add command to modify active tool.
Embed setup-validation routine to prevent removal from source code.


motoman_driver/MotoPlus/Controller.c
```diff
@@ -92,68 +92,6 @@ void reportVersionInfoToController()
...
-
-	//For all other error codes, please contact Yaskawa Motoman
-	//to have the MotoROS Runtime functionality enabled on your
-	//robot controller.
-	default:
-		mpSetAlarm(MOTOROS_SETUPERROR_ALARMCODE, "MotoROS: Controller cfg invalid", parameterValidationCode);
-		return FALSE;
-	}
-}
-
```
gavanderhoorn: This commit seems to remove all of the setup validation code, including the libraries, but I don't see it moved to somewhere else. How is the validation handled now?

gavanderhoorn, line 156: This commit seems to remove all of the setup validation code, including the libraries, but I don't see it moved to somewhere else. How is the validation handled now?

---

Edit: has this been integrated into the parameter extraction code?

motoman_driver/MotoPlus/Controller.c
```diff
@@ -198,13 +136,6 @@ BOOL Ros_Controller_Init(Controller* controller)
 	
 	// wait for controller to be ready for reading parameter
 	Ros_Controller_WaitInitReady(controller);
-	
-	bInitOk = Ros_Controller_CheckSetup();
-	if (!bInitOk)
-		return FALSE; //Don't allow initialization to continue
-
-	// Wait for alarms to clear, in case Ros_Controller_CheckSetup raised an alarm
-	Ros_Controller_WaitInitReady(controller);
```
gavanderhoorn, line 207: Is this not needed at all any more?

What about other causes of alarms, separate from the `CheckSetup(..)`?

motoman_driver/MotoPlus/MotionServer.c
```diff
@@ -69,13 +70,6 @@ void Ros_MotionServer_ConvertToJointMotionData(SmBodyJointTrajPtFull* jointTrajD
 STATUS Ros_MotionServer_DisableEcoMode(Controller* controller);
 void Ros_MotionServer_PrintError(USHORT err_no, char* msgPrefix);
 
-// IO functions:
-int Ros_MotionServer_ReadIOBit(SimpleMsg* receiveMsg, SimpleMsg* replyMsg);
-int Ros_MotionServer_WriteIOBit(SimpleMsg* receiveMsg, SimpleMsg* replyMsg);
-int Ros_MotionServer_ReadIOGroup(SimpleMsg* receiveMsg, SimpleMsg* replyMsg);
-int Ros_MotionServer_WriteIOGroup(SimpleMsg* receiveMsg, SimpleMsg* replyMsg);
-
-
```
gavanderhoorn, line 78: Were these removed because the IO functionality was moved to the IoServer?

motoman_driver/MotoPlus/MotionServer.c
```diff
@@ -286,6 +280,9 @@ int Ros_MotionServer_GetExpectedByteSizeForMessageType(SimpleMsg* receiveMsg, in
 	case ROS_MSG_MOTO_GET_DH_PARAMETERS:
 		expectedSize = minSize; //no additional data on the request
 		break;
+	case ROS_MSG_MOTO_SELECT_TOOL:
+		expectedSize = minSize + sizeof(SmBodySelectTool);
+		break;
```
gavanderhoorn, line 285: This seems to introduce a new message and some infrastructure, but it's unclear to me what the use-case is.

Is this needed for the PFL functionality (information about masses and inertias based on tool configuration)?

The rest of MotoROS is joint control only, so introducing a tool configuration setter seems unrelated.

motoman_driver/MotoPlus/MotionServer.c
```diff
@@ -1873,3 +1876,25 @@ int Ros_MotionServer_GetDhParameters(Controller* controller, SimpleMsg* replyMsg
...
+
+
+int Ros_MotionServer_SetSelectedTool(Controller* controller, SimpleMsg* receiveMsg, SimpleMsg* replyMsg)
+{
+	int groupNo = receiveMsg->body.selectTool.groupNo;
+	int tool = receiveMsg->body.selectTool.tool;
+
+	if (groupNo >= 0 && groupNo < controller->numRobot)
+	{	
+		if (tool >= 0 && tool < 64)
```
gavanderhoorn: Would it be possible to avoid this magic nr? Introduce a constant?

gavanderhoorn: Is `64` always the maximum nr of tool configurations? Across all supported controllers?

motoman_driver/MotoPlus/MpRosAllControllers.vcxproj
```diff
@@ -170,11 +170,6 @@
     <None Include="ParameterExtraction.mpLib" />
     <None Include="ParameterExtraction.yrcLib" />
     <None Include="ParameterExtraction.yrcmlib" />
-    <None Include="RosSetupValidation.dnLib" />
-    <None Include="RosSetupValidation.fsLib" />
-    <None Include="RosSetupValidation.mpLib" />
-    <None Include="RosSetupValidation.yrcLib" />
-    <None Include="RosSetupValidation.yrcmlib" />
```
gavanderhoorn, line 177: See my earlier question about these libraries disappearing.

motoman_driver/MotoPlus/SimpleMessage.h
```diff
@@ -72,6 +72,7 @@ typedef enum
 
 	ROS_MSG_MOTO_JOINT_TRAJ_PT_FULL_EX = 2016,
 	ROS_MSG_MOTO_JOINT_FEEDBACK_EX = 2017,
+	ROS_MSG_MOTO_SELECT_TOOL = 2018,
```
gavanderhoorn, line 75: This should be accompanied by an update to [REP-I4](https://github.com/ros-industrial/rep/blob/master/rep-I0004.rst).

motoman_driver/MotoPlus/Controller.c
```diff
@@ -515,6 +446,8 @@ void Ros_Controller_StatusInit(Controller* controller)
 	controller->ioStatusAddr[IO_ROBOTSTATUS_ESTOP_CTRL].ulAddr = 80027;   		// Controller E-Stop
 	controller->ioStatusAddr[IO_ROBOTSTATUS_WAITING_ROS].ulAddr = IO_FEEDBACK_WAITING_MP_INCMOVE; // Job input signaling ready for external motion
 	controller->ioStatusAddr[IO_ROBOTSTATUS_INECOMODE].ulAddr = 50727;			// Energy Saving Mode
+	controller->ioStatusAddr[IO_ROBOTSTATUS_PFL_STOP].ulAddr = 81702;			// PFL function stopped the motion
+	controller->ioStatusAddr[IO_ROBOTSTATUS_PFL_ESCAPE].ulAddr = 81703;			// PFL function escape from clamping motion
```
gavanderhoorn, line 449: There is no updated `IONAME.DAT` in this PR, is that something we'd want to add, or do these IOs already have names/comments associated with them (note: I don't have an HC-10 or FSU, so this may be a stupid question).

motoman_driver/MotoPlus/Controller.c
```diff
@@ -791,6 +735,18 @@ BOOL Ros_Controller_StatusUpdate(Controller* controller)
 						}
 						break;
 					}
+					case IO_ROBOTSTATUS_PFL_STOP: // PFL Stop
+					case IO_ROBOTSTATUS_PFL_ESCAPE: //  PFL Escaping
+					{
+						if (ioStatus[i] == 1)  // signal turned ON
```
gavanderhoorn: Not necessarily in this PR, but might be nice if we could use the constant here instead of `i`.

motoman_driver/MotoPlus/SimpleMessage.c
```diff
@@ -151,6 +151,12 @@ int Ros_SimpleMsg_MotionReply(SimpleMsg* receiveMsg, int result, int subcode, Si
 		replyMsg->body.motionReply.sequence = receiveMsg->body.jointTrajDataEx.sequence;
 		replyMsg->body.motionReply.command = ROS_MSG_MOTO_JOINT_TRAJ_PT_FULL_EX;
 	}
+	if (receiveMsg->header.msgType == ROS_MSG_MOTO_SELECT_TOOL)
+	{
+		replyMsg->body.motionReply.groupNo = ctrlGrp;
+		replyMsg->body.motionReply.sequence = -1;
```
gavanderhoorn: Is `-1` a proper value for `sequence` here in a reply? The other messages get the incoming `sequence` copied into the reply.

gavanderhoorn (commented):
First round of comments.

One action item:

 - could we update `.gitignore` to avoid adding all those `*.cache` files to the repository (ie: `MpRosAllControllers.vcxprojAssemblyReference.cache`)? They are computer-specific afaik.

motoman_driver/MotoPlus/Controller.c
ted-miller: > Edit: has this been integrated into the parameter extraction code?

Yes. As mentioned in the commit comments, we want to prevent someone from removing the validation routine. So it is embedded in some of the required parameter extractor functions.

motoman_driver/MotoPlus/Controller.c
ted-miller: This has been embedded in the parameter extraction code.

motoman_driver/MotoPlus/MotionServer.c
ted-miller: Yes

motoman_driver/MotoPlus/MotionServer.c
ted-miller: > Is this needed for the PFL functionality (information about masses and inertias based on tool configuration)?

Yes

Edit: Please see this document http://wiki.ros.org/motoman_driver?action=AttachFile&do=view&target=HC+Notes.pdf

motoman_driver/MotoPlus/MotionServer.c
ted-miller: Yes.

motoman_driver/MotoPlus/SimpleMessage.h
ted-miller: Ok

motoman_driver/MotoPlus/Controller.c
ted-miller: The 8xxxx series are "internal control status" signals. These are controlled by internal controller software and cannot be named by the user.

motoman_driver/MotoPlus/Controller.c
ted-miller: ok

motoman_driver/MotoPlus/SimpleMessage.c
ted-miller: The other messages are replying to a specific trajectory command; each of which has a sequence identifier in the trajectory. The SELECT_TOOL command doesn't have a sequence identifier, so -1 is used.

gavanderhoorn: I think it may still make sense to copy the incoming sequence nr, as it will allow clients to identify which response belongs to which request.

There is very little support for asynchronous interaction with the driver, but this is small thing to not make it even more difficult than it already is.

motoman_driver/MotoPlus/SimpleMessage.h
gavanderhoorn: Links: https://github.com/ros-industrial/rep/pull/26.

motoman_driver/MotoPlus/SimpleMessage.c
ted-miller: You're right. I was thinking it was a trajectory sequencer, rather than a packet sequencer. But that's a different command...

EDIT: Actually the incoming command doesn't have a sequence number. It is of type SmBodySelectTool. So, I'm adding a sequence id to this structure.

motoman_driver/MotoPlus/_buildLog/MotoROSDX100.log
```diff
@@ -1,8 +1,8 @@
-﻿
+﻿EXEC : warning : TED - Don't forget to remove the hardware-key short circuit...
```
gavanderhoorn: Is this a TODO that's still relevant?

ted-miller: Oops. 
That doesn't actually affect the build. So, I have just ignored the buildlog files.

motoman_driver/MotoPlus/SimpleMessage.h
gavanderhoorn: REP update was merged.

gavanderhoorn:
@ted-miller: could you take a look at this last request from the review?

>  - could we update `.gitignore` to avoid adding all those `*.cache` files to the repository (ie: `MpRosAllControllers.vcxprojAssemblyReference.cache`)? They are computer-specific afaik.

gavanderhoorn:
Travis hickup (job timed out). Just restarted.

gavanderhoorn:
Not sure whether this is related, but I've opened #287.

ted-miller:
A colleague of mine has found a possible issue with the param-extraction library for FS100 when using four control-groups. I still need to test/validate this.

gavanderhoorn:
> A colleague of mine has found a possible issue with the param-extraction library for FS100 when using four control-groups. I still need to test/validate this.

Is the expectation that the problem is caused by a change in `v1.9.0`?

If not, then I'd rather not hold the merge for it.

We can always release `v.1.9.1`.

ted-miller:
> Is the expectation that the problem is caused by a change in v1.9.0?

No. The issue is not caused by the new changes. I'm fine with just doing 191 later.

ted-miller:
We just discovered another problem. This one is caused by the v1.9.0 changes.

A colleague is doing additional testing on Monday and Tuesday. If testing goes well, I'll push the change on Tuesday or Wednesday.

ted-miller:
Both of fixes in 2ec0ad2fa7ee242b2dffe329f8c5db6e988d35c2 have been tested by Yaskawa personnel.

gavanderhoorn:
> 
> 
> Both of fixes in [2ec0ad2](https://github.com/ros-industrial/motoman/commit/2ec0ad2fa7ee242b2dffe329f8c5db6e988d35c2) have been tested by Yaskawa personnel.

Seeing as that commit only changes a set of binary files: could you post a descriptive comment of what the problem was and how we could recognise it if it does happen again or regresses?

ted-miller:
> Fix validation of system-software version for HC-10 on YRC1000 micro controller.

YRC1000micro would erroneously report alarm [8003[8] 'MotoROS not compatible with HC10'](http://wiki.ros.org/motoman_driver/Troubleshooting#Alarm:_8003.5B8.5D). This is caused by the detection mechanism that verifies the minimum system software of YBS2.30.00-00.

> Fix parameter-extraction functions when using four control groups on an FS100.

If the FS100 was configured with four control groups (extra arms or external axes), then MotoROS would freeze during initialization of the fourth group and the software would be unresponsive to ROS-I commands.

gavanderhoorn:
I've just pushed https://github.com/ros-industrial/motoman/compare/motoros190, which is a cleaner version of this PR (cleaner as in: no merge commits, HC-10 support still in a separate commit).

@ted-miller: I've rebased all of that on top of `kinetic-devel`, so you should be able to reset your `kinetic-devel` to that branch and the PR would auto-update as well (after a force push).

motoman_hc10_support/urdf/hc10_macro.xacro
```diff
@@ -0,0 +1,165 @@
...
+        <origin xyz="0 0 -0.500" rpy="0 0 0" />
+        <axis xyz="0 0 1" />
+        <limit lower="-3.14159" upper="3.14159" effort="31.36" velocity="3.1415"/>
+    </joint>
+    <joint name="${prefix}joint_5_b" type="revolute">
+        <parent link="${prefix}link_4_r"/>
+        <child link="${prefix}link_5_b"/>
+        <origin xyz="0 0.162 0" rpy="0 0 0" />
+        <axis xyz="0 -1 0" />
+        <limit lower="-2.6179" upper="2.6179" effort="31.36" velocity="4.36333"/>
```
gavanderhoorn: The `MOTOMAN-HC10 INSTRUCTIONS` manual (`HW1483896`) specs `[-180; 180]` as limits for joint B. That would be `3.1415..` radians, not `2.6179`.

Is the xacro macro correct, or should it in fact be `+-180` degrees?

gavanderhoorn:
@EricMarcil @ted-miller: I've spent some time working on the HC10 model.

Would it be OK if I push this to this PR?

hc10_before_after

EricMarcil:
Yes, that looks very nice.
You are also correct about the B-axis range.  It should be 180 deg.
I've also noticed that the R-axis effort is lower.  It should be changed to effort="39.2"

gavanderhoorn:
I have another of other small fixes and cleanups.

As such I would like to suggest to omit `motoman_hc10_support` from this PR: the PR specifically adds MotoROS `1.9.0`. It would be good to just do that in this PR.

I'll open a new PR that adds just the `motoman_hc10_support` package.

Agreed @ted-miller @EricMarcil?

EricMarcil:
@gavanderhoorn : I don't have a problem with the it.

ted-miller:
If this will be merged into the main branch very soon, then I'm ok with them being separated. We have had a lot of interest in the past couple weeks regarding the HC-10. Since both the support-package and the driver are required for the new robot type, it is good to have them in one place for people to download.

gavanderhoorn:
I plan to merge everything *tonight* (ie: your day).

The only outstanding changes are to the `motoman_hc10_support`. As soon as I've cleaned up the MotoROS commit history, we can merge it in.

At that point I can submit the support package PR.

gavanderhoorn:
I've opened #298 which adds just `motoman_hc10_support`.

I'll update the `motoros190` branch with just the MotoROS commits.

gavanderhoorn:
I've updated the `motoros190` branch and dropped the commit that added `motoman_hc10_support`.

See [ros-industrial/motoman/compare/motoros190](https://github.com/ros-industrial/motoman/compare/motoros190).

@ted-miller: could you reset your `kinetic-devel` to `motoros190` again (just as you did last-time (email)).

That should update this PR as well.

gavanderhoorn:
Thanks. We'll wait for Travis.

I'll probably merge this tomorrow morning (my morning).

gavanderhoorn (approved):

gavanderhoorn:
Looks like Travis was happy.

So am I :smile: 

Thanks for the updates @ted-miller :+1: 

:robot:

gavanderhoorn:
Will you update the various wiki links to `1.9.0` @ted-miller ?

ted-miller:
Thanks for the merge.
Yes, I will update the wiki links.

PR #290, opened 2019-09-05
Add basic support for Yaskawa Motoman GP88 6DOF arm

### LINKED ISSUE #288: Support Request for YRC1000 GP88
From https://github.com/ros-industrial/motoman/pull/277#issuecomment-527829509:
> I am currently working on basic GP88 support and would like to add the correct efforts there as well. Are these YRC1000 parameter charts available for download anywhere, or could you post the values for the GP88?

cjue:
Meshes are derived from official Yaskawa 3D model files, collision meshes are convex hulls.

Includes mesh processing scripts by @jmarsik in motoman_gp180_support https://github.com/ros-industrial/motoman/pull/277.

Closes https://github.com/ros-industrial/motoman/issues/288 when merged.


cjue:
Fixed the mesh issue with `meshes/gp88/collision/link_2_l.stl`.

EricMarcil:
@cjue  & @gavanderhoorn : I've reviewed the PR and tested the model in RViz.  Everything seems good to me.
Comment: 
1) In the gp88_macro.xacro, I usually put the last offset from the T-axis to the Flange in the "joint_6_t-tool0" instead of the "joint_6_t" because that's how it is done in our internal model but mathematically it works out to the same thing.
2) There's a typo in the comment of robot_interface_streaming_gp88.launch. 
 Usage:
    robot_interface_streaming_gp**188**.launch robot_ip:=<value> controller:=<yrc1000> 
Should probably be:
 Usage:
    robot_interface_streaming_gp**88**.launch robot_ip:=<value> controller:=<yrc1000>

Nice work.  Thanks for make this available.

cjue:
@EricMarcil: Thanks for the feedback. I fixed the typo and applied the change you mentioned regarding `joint_6_t` and `tool0`. 

The same urdf issue is also present in #277.

gavanderhoorn:
I'm not against applying the final transform between `joint_6_t` and `tool0`, but it is a departure from the models that were not contributed by Yaskawa Motoman.

We might want to update all other models as well if we believe this is a better way of modelling the kinematics for these robots.

gavanderhoorn:
@cjue: thanks for the PR, but I believe it would be good if we could reduce the quality of the `visual` meshes a bit.

Right now the `gp88/visual` folder is 28 MB in total, which is almost 1/5th of the total size of the repository. A quick test with Blender shows that the meshes can probably be reduced by about 80 to 90% without loss of visual quality.

I always use Blender with the *Decimate* modifier, but you could also take a look at the [scripts](https://github.com/ros-industrial/motoman/tree/197d2f8da805c9849a821e8b627748811b3242a1/motoman_gp180_support/tools/process_meshes) added in #277 (the simplification script).

gavanderhoorn (changes requested):
@cjue: thanks for the PR, but I believe it would be good if we could reduce the quality of the `visual` meshes a bit.

Right now the `gp88/visual` folder is 28 MB in total, which is almost 1/5th of the total size of the repository. A quick test with Blender shows that the meshes can probably be reduced by about 80 to 90% without loss of visual quality.

I always use Blender with the *Decimate* modifier, but you could also take a look at the [scripts](https://github.com/ros-industrial/motoman/tree/197d2f8da805c9849a821e8b627748811b3242a1/motoman_gp180_support/tools/process_meshes) added in #277 (the simplification script).

cjue:
> I'm not against applying the final transform between joint_6_t and tool0, but it is a departure from the models that were not contributed by Yaskawa Motoman.

One note: In this motoman repository all Urdfs except for the new gp180_120 use this convention.

> @cjue: ... it would be good if we could reduce the quality of the visual meshes a bit.

I increased the decimation parameter, setting the goal for the scripts from #277 to 15% remaining polygons. The meshes are now 9MB in total, similar to the MH5.

gavanderhoorn:
> > I'm not against applying the final transform between joint_6_t and tool0, but it is a departure from the models that were not contributed by Yaskawa Motoman.
> 
> One note: In this motoman repository all Urdfs except for the new gp180_120 use this convention.

Unless I'm misunderstanding you, at least the [SIA20](https://github.com/ros-industrial/motoman/blob/197d2f8da805c9849a821e8b627748811b3242a1/motoman_sia20d_support/urdf/sia20d_macro.xacro#L180-L191) model does not have a translation between `link_t` and `tool0`. And there are probably others.

As I wrote though: I'm fine with both.

> I increased the decimation parameter, setting the goal for the scripts from #277 to 15% remaining polygons. The meshes are now 9MB in total, similar to the MH5.

That's already quite a bit better. But still quite large. I'll see what I can do in Blender.

re: MH5: yes, that model is too detailed as well, but was unfortunately merged like that.

gavanderhoorn:
@cjue: it also looks like you saved `link_6_t.stl` over `base_link.stl`: there is no `base_link.stl` any more in your PR now.

I've opened cjue/motoman#1 against your PR branch, which contains an alternative set of decimated meshes (and a fixed `base_link.stl`).

I'm still not entirely happy about this, as `link_3_u.stl` for instance is still 2MB+.

I'm actually starting to wonder whether it would make sense to remove the cables modelled in the mesh: they don't update based on pose of the robot and collision meshes could just be made to include them instead (so they would still be considered in motion planning). That would significantly reduce the size of the `visual` meshes.

---

Edit: `link_3_u.stl` also contained some visual artefacts, probably due to decimation with Meshlab. Those have been corrected as well.

cjue:
Thanks for the update, and catching the lost base_link - no idea how that happened.

> I'm actually starting to wonder whether it would make sense to remove the cables

Yes, removing them from just the visual mesh sounds like a good idea, in order to keep the size down.

> Unless I'm misunderstanding you, at least the SIA20 model does not have a translation between link_t and tool0. And there are probably others.

Sorry, I was only checking for `link_6_t` without thinking about other variants of the name.

EricMarcil:
Those cables are standard on the robot.  I would tend to keep them unless it really has a significant impact on mesh size and performance...

cjue:
@EricMarcil would prefer keeping the cables.

Is the current size of the meshes a blocking issue?

gavanderhoorn:
No, the cables are not blocking this.

cjue:
As I failed to mention it specifically: I merged your changes to reduce file size and fix the base_link mesh: https://github.com/cjue/motoman/pull/1

gavanderhoorn (approved):
I've added a few fixup commits, but overall this is an OK PR.

Thanks for iterating @cjue, and thanks for contributing the model.

:+1:

cjue:
Thanks as well for your feedback and contributions to the PR and the overall package @gavanderhoorn @EricMarcil!

PR #286, opened 2019-08-30
meta: add Michael as maintainer.

gavanderhoorn:
As per subject.

Welcome @marip8 ! :fireworks:

PR #299, opened 2019-10-11
Add GP25 Support Package

EricMarcil:
Add support package for the Yaskawa Motoman GP25 robot model.


motoman_gp25_support/launch/test_gp25.launch
```diff
@@ -0,0 +1,8 @@
+
+<launch>
+	<include file="$(find motoman_gp25_support)/launch/load_gp25.launch" />
+	<param name="use_gui" value="true" />
+	<node name="joint_state_publisher" pkg="joint_state_publisher" type="joint_state_publisher" />
```
marip8: I believe the `use_gui` parameter should technically be a private parameter of the `joint_state_publisher` node

motoman_gp25_support/package.xml
```diff
@@ -0,0 +1,53 @@
...
+    <p>
+      <b>Specifications</b>
+    </p>
+    <ul>
+      <li>gp25 - Default</li>
+    </ul>
+    <p>
+      Joint limits and maximum joint velocities are based on the information 
+      found in the online 
+      https://www.motoman.com/hubfs/Robots/gp25.pdf
```
marip8: The link to this document did not work for me

motoman_gp25_support/package.xml
```diff
@@ -0,0 +1,53 @@
...
+      in this package, be sure to check they are correct for the particular
+      robot model and configuration you intend to use them with.
+    </p>
+  </description>
+  <author>Eric Marcil</author>
+  <maintainer email="shaun.edwards@gmail.com">Shaun Edwards</maintainer>
+  <maintainer email="g.a.vanderhoorn@tudelft.nl">G.A. vd. Hoorn (TU Delft Robotics Institute)</maintainer>
+  <license>BSD</license>
+
+  <url type="website">http://wiki.ros.org/motoman_gp25_support</url>
```
marip8, line 36: This may be in preparation for build farm release, but technically this website does not exist.

motoman_gp25_support/test/launch_test.xml
```diff
@@ -0,0 +1,59 @@
...
+       sub-namespaces under the main namespace.
+
+  Notes:
+
+  1. XML extension is used in order to avoid beinging included
+  in roslaunch auto-complete.
+
+  2. Group tags with namespaces are used to avoid node name collisions when
+  testing multpile launch files
+-->
```
marip8: These comments seem unnecessary and could be deleted

motoman_gp25_support/urdf/gp25_macro.xacro
```diff
@@ -0,0 +1,165 @@
...
+    <link name="${prefix}tool0"/>
+    <!-- end of link list -->
+
+    <!-- joint list -->
+    <joint name="${prefix}joint_1_s" type="revolute">
+        <parent link="${prefix}base_link"/>
+        <child link="${prefix}link_1_s"/>
+        <origin xyz="0 0 0.505" rpy="0 0 0" />
+        <axis xyz="0 0 1" />
+        <limit lower="-3.1416" upper="3.1416" effort="3087" velocity="3.6651"/>
```
marip8: The limit tags might be a little more readable (and accurate) if you used math expressions to evaluate degrees (i.e. `${pi}` or `${radians(90)}`, etc.). It might be worth changing this everywhere you use radian values

motoman_gp25_support/urdf/gp25_macro.xacro
```diff
@@ -0,0 +1,165 @@
...
+        <origin xyz="0.795 0 0.200" rpy="0 0 0" />
+        <axis xyz="-1 0 0" />
+        <limit lower="-3.4906" upper="3.4906" effort="88.2" velocity="7.3303"/>
+    </joint>
+    <joint name="${prefix}joint_5_b" type="revolute">
+        <parent link="${prefix}link_4_r"/>
+        <child link="${prefix}link_5_b"/>
+        <origin xyz="0 0 0" rpy="0 0 0" />
+        <axis xyz="0 -1 0" />
+        <limit lower="-2.6179" upper="2.6179" effort="70.56" velocity="7.3303"/>
```
marip8: I have received the suggestion in the past regarding robot URDFs to only translate joints in one direction at a time. You might consider translating joint R in the Z-axis only by 0.2m, and then translating joint B in the X-axis only by 0.795m for consistency with other packages

motoman_gp25_support/urdf/gp25_macro.xacro
```diff
@@ -0,0 +1,165 @@
...
+        <child link="${prefix}link_6_t"/>
+        <origin xyz="0 0 0" rpy="0 0 0" />
+        <axis xyz="-1 0 0" />
+        <limit lower="-7.9412" upper="7.9412" effort="50.31" velocity="15.4461"/>
+    </joint>
+    <joint name="${prefix}joint_6_t-tool0" type="fixed">
+        <origin xyz="0.100 0 0" rpy="3.1415926 -1.570796 0"/>
+        <parent link="${prefix}link_6_t"/>
+        <child link="${prefix}tool0"/>
+    </joint>
```
marip8: Typically joint T is translated in the X-axis from joint B, and the `tool0` frame only has a rotation from joint T. There is nothing technically incorrect about this approach, but you might consider changing this to be consistent with the other robot models

marip8 (changes requested):
Looks good and functions properly in simulation for me. I just have a few comments mostly about convention and documentation

motoman_gp25_support/package.xml
EricMarcil: Yes, our website was recently redone.  I just replace the robot model of an older model with GP25 without paying attention to it.  I'll correct the link to:
https://www.motoman.com/en-us/products/robots/industrial/assembly-handling/gp-series/gp25

motoman_gp25_support/launch/test_gp25.launch
EricMarcil: I'm not very familiar with the launch files.  I just used a previous version (GP7) as a reference.  How would you go about making that private? 
@gavanderhoorn do you agree that it should be privarte?

motoman_gp25_support/urdf/gp25_macro.xacro
EricMarcil: Yes, I agree.  I'll make the change.

motoman_gp25_support/package.xml
EricMarcil: @gavanderhoorn:  Will this be created or should it be removed?

motoman_gp25_support/test/launch_test.xml
EricMarcil: @gavanderhoorn: Do you agree?

motoman_gp25_support/urdf/gp25_macro.xacro
EricMarcil: As above, it matches the standard Yaskawa model construction.  It then makes it easier to create models and compare them with Yaskawa's MotoSim software.

gavanderhoorn:
Overall comment: @marip8 recently joined the maintainer group for `ros-industrial/motoman` (see #286).

Thanks @marip8 for your initial review.

motoman_gp25_support/package.xml
gavanderhoorn: @EricMarcil: I would actually recommend we start noting document numbers/IDs and revisions here, instead of linking to sites.

If (or most likely *when*) the Motoman site gets updated again, all links in package manifests run the risk of going stale again and I doubt we'll pay enough attention to catch that.

I've done this in the HC10 support package:

https://github.com/ros-industrial/motoman/blob/501d952a0ebae84399f13f0a8cc28fb2da8a54ed/motoman_hc10_support/package.xml#L20-L22

We could see whether this needs to be updated, but I would recommend we start doing this for all new support packages (and actually update the old ones, whenever we have the change).

motoman_gp25_support/package.xml
gavanderhoorn: Yes, we'll create them.

motoman_gp25_support/test/launch_test.xml
gavanderhoorn: I'm OK with this being removed.

The instructions are more for the support package author than for users of the package.

motoman_gp25_support/urdf/gp25_macro.xacro
gavanderhoorn: :+1: to this suggestion.

I've done the same in 07b8fc0c (which is one of the commits in the HC10 PR).

motoman_gp25_support/urdf/gp25_macro.xacro
gavanderhoorn: @EricMarcil: does the internal model of the robot really translate over both the X and Z axis between joints `R` and `B`?

motoman_gp25_support/launch/test_gp25.launch
gavanderhoorn: Yes, I agree (but haven't bothered you with it in the past).

You could refer to [motoman_epx_support/launch/test_epx1250.launch](https://github.com/ros-industrial/motoman_experimental/blob/474ed145161b6468eacb78ab6b42e4c1ed8f3eab/motoman_epx_support/launch/test_epx1250.launch) for an example.

gavanderhoorn:
Additional initial review comment: please do not repeat robot model/series/variant prefixes in the mesh filenames (such as `gp25_link_2_l.stl`). The files are stored in a package with subdirectories that already sufficiently namespace those files (ie: prevent collisions with other names).

I would suggest to use names corresponding to the `link` `name` attributes in the `_macro.xacro`: `link_2_l.stl`.

motoman_gp25_support/urdf/gp25_macro.xacro
EricMarcil: Oops, it seems that my comment above that I was referencing was never posted but it's basically the same topic.
There is a translation in both X and Z between the U and R axes (like the physical joint).  We use DH representation so it corresponds to the "a" and "d" offsets.  In our model, the R, B and T axes origins all super-impose on the same spot in space (we refer to it a the "P" point, which is in the middle of the wrist.  We then have an offset between the "T-axis" and the robot flange, which corresponds to the "Tool 0".  All the Yaskawa models that I have build recently are like that.

EricMarcil:
I'll also correct the mesh files names and the reference to them.

@gavanderhoorn  and @marip8: thanks for the review.

motoman_gp25_support/urdf/gp25_macro.xacro
gavanderhoorn: Thanks for the clarification.

It would still be advisable to separate the translations and only have one per dimension, as this seems to help tools like IKFast generate IK solvers (at least, it did last time I checked).

I would recommend to do what @marip8 suggested.

EricMarcil: Ok, if you think it may help the IK Solvers.  I'll have to adjust the Mesh to relocate the origin of the model.  I'll try to get it done today.

motoman_gp25_support/urdf/gp25_macro.xacro
EricMarcil: Changes made to move the 0.795 mm offset between the R and B joints.

EricMarcil:
@gavanderhoorn & @marip8 : I've made all the changes.  When you have a chance please review the newest commit and let me know if you think anything else is needed.
Thanks!

marip8 (approved):

PR #309, opened 2019-11-25
Adding HC10DT robot variant to motoman_hc10_support package 

EricMarcil:
Adding the HC10DT variant which has the Direct Teach module at the end of the HC10.  Moving the tool0 to 170 mm (instead of 130 mm) from the link_6_t joint.


gavanderhoorn:
@EricMarcil: I've added some commits to your PR.

None of them are too important, but 446b7e3 adds a slightly nicer mesh for the DT:

rviz_screenshot_2020_01_07-18_03_22

Could you see whether you agree with these changes?

EricMarcil:
@gavanderhoorn 👍 I've reviewed the changes.  It's all good with me.  Thanks.

gavanderhoorn:
Thanks :+1: 

I've merged it.

PR #319, opened 2020-02-29
Fix #258

### LINKED ISSUE #258: Motoman Driver Error: Failed to init message, buffer size too small: 3
Hi,

for safety reasons I want to call `/robot_disable` service, directly after an emergency stop occurred, to make shure, that the robot doesn't start to move after the emergency stop is reseted. This works properly 95% of the times, but sometimes I can not call /robot_disable service and get the following error message:

```
ERROR [/tmp/binarydeb/ros-kinetic-simple-message-0.6.0/src/simple_message.cpp:107(simple_message::SimpleMessage::init) [topics: /rosout] Failed to init message, buffer size too small: 3
```

Unfortunately I am not able to reproduce, under what circumstances this error occurres exactly, but it seems to be only while executing trajectories and not when the robot stands still. Is there an option to increase the buffer size, or can I be sure that the robot doesn't accept motion commands, after an emergency stop reset, regardless of whether calling `/robot_disable` service or not.

yszheda:
Fix the thread safety issue of using `industrial::smpl_msg_connection::SmplMsgConnection` in #258

### DIFF
--- a/motoman_driver/include/motoman_driver/joint_trajectory_streamer.h
+++ b/motoman_driver/include/motoman_driver/joint_trajectory_streamer.h
@@ -32,6 +32,7 @@
 #ifndef MOTOMAN_DRIVER_JOINT_TRAJECTORY_STREAMER_H
 #define MOTOMAN_DRIVER_JOINT_TRAJECTORY_STREAMER_H
 
+#include <boost/thread/mutex.hpp>
 #include <map>
 #include <string>
 #include <vector>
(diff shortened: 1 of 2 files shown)


driftregion (approved):

gavanderhoorn:
Closing. Rebased version: #389.

gavanderhoorn:
Thanks for the initial PR @yszheda.

Issue #125, opened 2016-08-29
sda10f_support: missing face in S-link (visual quality)


gavanderhoorn:
There appears to be a face (or multiple) missing in the mesh associated with `link_1_s` (the black area underneath the rounded upper part):

![sshot](https://cloud.githubusercontent.com/assets/4550046/18049703/1ecfe5fa-6dea-11e6-9dc1-a189e632d9b0.png)

This is the `motoman_axis_s.stl` file, in the `meshes/sda10f/visual` directory of the package.

Even though it is in a rather hard to reach location, missing faces are potentially problematic when using this mesh for collision detection, and thus should be fixed.

EricMarcil:
@gavanderhoorn fixed it in the PR #327.  Please review...

PR #321, opened 2020-03-14
SDA10F Support Package Update

### LINKED ISSUE #232: sda10f: dimensions in URDF does not match specifications

EricMarcil:
Restructured kinematic model to match standard Yaskawa model
Relocation position of mesh origin
Update of B and E axis meshes to correct errors


gavanderhoorn:
@EricMarcil: I've pushed some additional commits to this PR to fix some small issues. Most were not introduced by you, but as we're taking a look at this support package anyway, I thought we might as well fix those.

As to the new meshes/updates: it appears links B and E still/again have something wrong with them. Specifically, the origins of the meshes for those links are not completely correct.

See the following screenshots:

rviz_sda10f_link_b

and

rviz_sda10f_link_e

notice how in the first, the disk of link T does not appear to be centred on the previous link (B), and in the second screenshot, link E does not appear to be correctly centred on the output of the previous link.

Both of these are caused by the origin of the meshes not being exactly where it should be.

I can fix this with Blender, but first wanted to know whether you agree.

gavanderhoorn (changes requested):
Mesh origins need to be checked.

EricMarcil:
@gavanderhoorn Sorry about that.  Corrected the meshes origin.  Please confirm correction when you have a chance.

motoman_sda10f_support/urdf/common_torso_macro.xacro
```diff
@@ -1,77 +1,73 @@
...
+          <color rgba="0 0 1 1"/>
+        </material>
+      </visual>
+      <collision>
+        <geometry>
+          <mesh filename="package://motoman_sda10f_support/meshes/sda10f/collision/base.stl" />
+        </geometry>
+      </collision>
+    </link>
+    <link name="${prefix}link_b1">
```
gavanderhoorn, line 20: It looks like `link_b1` and `link_b2` both load the same geometry. This seems weird and unnecessary: it's ok for `link`s not to have any associated geometry.

Would it be an idea to drop the `visual` and `collision` elements from `link_b2`?

gavanderhoorn (commented):
Meshes are OK now.

Added a single in-line comment.

motoman_sda10f_support/urdf/common_torso_macro.xacro
EricMarcil: I agree.  link_b2 is a virtual link that is simply used to serve as the 2nd arm base.  It will always be the same as link_b1.  It should have any geometry.

gavanderhoorn: Removed it in 0165144.

gavanderhoorn:
Looks like #125 is not fixed by this PR.

@EricMarcil: did you only update the origins of the meshes? That could explain why the face is still missing.

EricMarcil:
@gavanderhoorn: No, I did not.  I was not aware of #125 
Looking back, I should have redone the whole model from scratch, rather than try to fix the existing one.  It's turning out to be more work...  
I do have another model for that link.  I'll try to switch it out later today.

gavanderhoorn:
We don't need to fix #125 now, in this PR.

I just wanted to make sure I understood what was included in this PR.

gavanderhoorn (approved):
This fixes the major outstanding issues with this support package.

Thanks @EricMarcil :+1:

gavanderhoorn:
Thanks for iterating on the PR @EricMarcil.

PR #327, opened 2020-05-13
Update link_s.stl for SDA10F

### LINKED ISSUE #125: sda10f_support: missing face in S-link (visual quality)

EricMarcil:
Fixed missing surface in the link_s mesh


EricMarcil:
Oops, I forgot  to mention that this is for the SDA10F model.

gavanderhoorn:
This PR adds the missing face, but seems to have introduced some artefacts in other places.

I'm not sure whether we feel this is an issue, but I wanted to note it here.

See the screenshots for a comparison of the previous version of the mesh (first screenshot) and the version in this PR (second screenshot):

rviz_screenshot_2020_05_19-12_23_48

rviz_screenshot_2020_05_19-12_24_54

Note how some of the features in the geometry look like they have been corrupted almost.

EricMarcil:
This occurred during the file size reduction.  I could increase the file size to keep more details.  I didn't evaluate those lost of geometry as being critical since those surface will be covered by the next link.  I can easily generate a new model with more detail, if you prefer.

gavanderhoorn:
> I can easily generate a new model with more detail, if you prefer.

Yes, please do.

We can subsample when required here. But let's start with a bit more detail.

EricMarcil:
How's this?
It's double the previous version but 1/5 of my original model.

image

EricMarcil:
Actually this might be a better compromise at 1/4 of my original resolution.

image

gavanderhoorn:
@EricMarcil: I've just added a commit which takes your latest mesh and reduces it in size by about 200KB without too much visible reduction in detail I believe.

EricMarcil:
What tool are you using?  I use Meshlab: Quadratic Edge Collapse Decimation algorithm

gavanderhoorn:
I use [Blender](https://www.blender.org). It takes some getting used to, but is a really powerful 3D modelling tool.

To reduce mesh size, I use the [Decimate modifier](https://docs.blender.org/manual/en/latest/modeling/modifiers/generate/decimate.html).

gavanderhoorn:
Mesh now looks like this:

meshlab_link_s01

EricMarcil:
Looks good.  I'll have to look into Blender.  Thanks

gavanderhoorn (approved):

gavanderhoorn:
Thanks for fixing this :+1: :robot:

gavanderhoorn:
Had to fix this -- I'm sure I had "Squash and Merge" selected, but Github created a plain merge commit.

Changes have now been merged in `kinetic-devel` in 4a29807.

PR #333, opened 2020-06-01
Remove bit-rotted move_to_joint script

gavanderhoorn:
As per subject.

This script confuses users, and also uses an approach which should not be used.


marip8 (approved):

gavanderhoorn:
Thanks @marip8 :+1:

PR #337, opened 2020-06-19
Add SIA50 support

### LINKED ISSUE #334: SIA50D robot model support
Hi,
Thank you for maintaining this high-quality package!

I was looking into an application using a [SIA50D](https://www.motoman.com/en-us/products/robots/industrial/assembly-handling/sia-series/sia50d). As part of this I am looking for the robot description but it is not yet included in this repository.

Are there plans to add support for this model (even if only description/MoveIt config) to this package?

I am happy to contribute time and effort to adding the robot model but do not have access to meshes/further specifications. If someone from Yaskawa/Motoman follows this repository, or if you have access within ROS-Industrial, I am happy to help getting it added to this package if of interest.

Thanks,
Wolfgang

wxmerkt:
This is a WIP PR to add support for the SIA50D. This PR is based on the support for the SIA20D and the mesh file processing for the GP88. Thanks to @EricMarcil for providing the meshes, model, and instructions.

The automatically created collision mesh/convex hull for the 6th link is currently not great (used the same script as for GP88).

I am planning on adding approximate CoG, inertia, and mass information from CAD assuming uniform mass distribution in each link (as we are planning to use the model for dynamic trajectory optimisation and the correct values are not available for open source release). Please let me know if this is OK or if you prefer to omit the inertial tags in this case.

Looking forward to your comments and valuable feedback, and thanks for all the support,
Wolfgang

Please note: This has _not_ been tested with a real robot.

Resolves #334


gavanderhoorn:
Thanks for the PR.

High-level comment to start with: I'd like to remove the `d` suffix on the robot variant identifier (ie: `sia50d`).

Afaik this encodes for the controller (a DX100 in your case most likely), but afaik, the robot would not be different when used with an FS100 (I can't find it with an FS100 though).

@EricMarcil / @ted-miller: could you confirm?

There are already too many support packages which differentiate on the controller (ie: `motoman_sia5d_support` and `motoman_sia5f_support`), but most of the times it's the same robot.

See also #65.

Additionally: if you'd really like to prefix your commit msgs with the pkg name, please drop the `motoman_` prefix. All commits are pushed to the `motoman` repository, so everything here is a `motoman_` package. We like to avoid repeating ourselves, so no need for that `motoman_` there again. As a suggestion: `sia50_support: <your commit comment>`.

motoman_sia50d_support/test/launch_test.xml
```diff
@@ -0,0 +1,59 @@
...
+  <group ns="test_sia50d">
+	  <include file="$(find motoman_sia50d_support)/launch/test_sia50d.launch"/>
+  </group>
+
+
+  <group ns="robot_interface_streaming_sia50d">
+    <group ns="dx100" >
+	    <include file="$(find motoman_sia50d_support)/launch/robot_interface_streaming_sia50d.launch">
+		    <arg name="robot_ip"   value="$(arg req_arg)" />
+        <arg name="controller" value="$(arg dx100)"/>
```
gavanderhoorn: CI is likely failing because of this line (and the same one below for the `visualize` launch file).

You're passing the `controller` arg, but it's not actually something you can set: it already has a `value` (namely: `dx100`).

wxmerkt:
Thanks for the review and feedback. I added a fix for the CI and will monitor it. From the website, I believe the SIA50D exists only with the DX100 controller - there is no F variant listed: https://www.motoman.com/en-us/products/robots/industrial/assembly/sia
I'll wait for feedback from Motoman and will rename the package to drop the "d" postfix then accordingly.

As for adding approximate inertial properties, is that okay or would you prefer to omit them?

gavanderhoorn:
> From the website, I believe the SIA50D exists only with the DX100 controller

That's not really the point: from other robots we know that the last letter encodes the controller. With other support pkgs we also don't add that to the variant identifier (here in this repository, nor anywhere else in ROS-Industrial).

Tbh we never should've done that with the SIA5 or the SDA series either.

> As for adding approximate inertial properties, is that okay or would you prefer to omit them?

I would personally remove them, but it depends a bit on how inaccurate they are: if they lead to OK behaviour in Gazebo, we could keep them, but we would have to explicitly state they are approximations.

wxmerkt:
Thanks for the feedback. I've renamed the package.

I've committed the approximate inertial parameters in https://github.com/ros-industrial/motoman/pull/337/commits/79e3d84e3000e9919151f4364a336f0ed831e1e3. These values have been obtained by using the full mass of the manipulator according to spec sheet (650kg), and assuming a uniform mass distribution across all links, and then derived from CAD. If @EricMarcil could please comment if they are approximately OK I would leave them in the PR; otherwise I will revert/drop this commit from the PR. (We have not validated them in Gazebo but for motion planning using full rigid-body dynamics. If there is a Gazebo package for Motoman, I am happy to adapt/test it with it).

EricMarcil:
@gavanderhoorn Though it may no longer be relevant, to answer your question, the SDA50 is only available with the DX100 since the power requirement are to big for the FS100 which is a compact controller for smaller robots.

gavanderhoorn:
> @gavanderhoorn Though it may no longer be relevant, to answer your question, the SDA50 is only available with the DX100 since the power requirement are to big for the FS100 which is a compact controller for smaller robots.

Thanks @EricMarcil.

It's more the general naming in this case, not so much whether this specific robot would be available with a specific controller.

re: power requirements: yes, I'm somewhat jealous of @wxmerkt having access to an SIA50.

EricMarcil:
@wxmerkt First, thank you for you contribution.
Concerning the mass and other properties, I haven't gone into details but I can quickly tell you that the URB mass is over evaluated by more than 50%.  
Unfortunately, note that I will not comment any further on this topic of estimating these values, since this information are considered confidential.  If you need further details for your application then I would invite you to contact your local Yaskawa representative and establish a NDA with them.  Once that is establish, they can (or I can) directly assist you.

motoman_sia50_support/launch/robot_interface_streaming_sia50d.launch
```diff
@@ -0,0 +1,21 @@
...
+   - 7 joints
+
+  Usage:
+    robot_interface_streaming_sia50d.launch robot_ip:=<value> controller:=<dx100>
+-->
+<launch>
+  <arg name="robot_ip" doc="IP of controller" />
+	
+  <!-- controller: Controller name (dx100) -->
+  <arg name="controller" value="dx100" doc="Series of the controller (dx100)" />
```
gavanderhoorn: Could you make this an `arg` with a `default` instead?

It'd better align with the other `.launch` files.

We could also opt for removing the `arg` completely, but that would mean we'd have to change these files again if/when Yaskawa decides that there is going to be a SIA50 with a YRC1000 (or something like that).

motoman_sia50_support/launch/robot_state_visualize_sia50d.launch
```diff
@@ -0,0 +1,28 @@
...
+   - 7 joints
+
+  Usage:
+    robot_state_visualize_sia50d.launch robot_ip:=<value> controller:=<dx100>
+-->
+<launch>
+	<arg name="robot_ip" doc="IP of controller" />
+
+  <!-- controller: Controller name (dx100) -->
+  <arg name="controller" value="dx100" doc="Series of the controller (dx100)" />
```
gavanderhoorn: Same here: please `default` instead of `value`.

motoman_sia50_support/launch/test_sia50d.launch
```diff
@@ -0,0 +1,7 @@
+<launch>
+	<include file="$(find motoman_sia50_support)/launch/load_sia50d.launch" />
+	<param name="use_gui" value="true" />
+	<node name="joint_state_publisher" pkg="joint_state_publisher" type="joint_state_publisher" />
```
gavanderhoorn: We'll deal with the `joint_state_publisher_gui` issue in a later PR, for all support packages.

motoman_sia50_support/package.xml
```diff
@@ -0,0 +1,62 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>motoman_sia50_support</name>
+  <version>0.3.5</version>
+  <description>
+    <p>ROS-Industrial support for the Motoman SIA50D (and variants).</p>
```
gavanderhoorn:
```suggestion
    <p>ROS-Industrial support for the Motoman SIA50 (and variants).</p>
```

motoman_sia50_support/package.xml
```diff
@@ -0,0 +1,62 @@
...
+    <p>
+      <b>Specifications</b>
+    </p>
+    <ul>
+      <li>SIA50D - Default</li>
+    </ul>
+    <p>
+      Joint limits and maximum joint velocities are based on the information 
+      found in the online 
+      https://www.motoman.com/en-us/products/robots/industrial/assembly-handling/sia-series/sia50d
```
gavanderhoorn: Please mention document title and number/revision/ID instead of linking to a website.

The link will go stale.

motoman_sia50_support/package.xml
```diff
@@ -0,0 +1,62 @@
...
+      All urdfs are based on the default motion and joint velocity limits, 
+      unless noted otherwise.
+    </p>
+    <p>
+      Please note: The dynamic properties in this robot model are approximate 
+      and derived from CAD information. For exact dynamic properties, please 
+      contact Motoman (available via NDA).
+    </p>
+    <p><strong>!!! THIS PACKAGE HAS NOT BEEN TESTED WITH A PHYSICAL ROBOT !!!</strong></p>
```
gavanderhoorn: I appreciate you adding this, but let's make it a regular paragraph in the manifest: revert the all-caps and remove the `strong`.

motoman_sia50_support/package.xml
```diff
@@ -0,0 +1,62 @@
...
+    <p>
+      Before using any of the configuration files and / or meshes included
+      in this package, be sure to check they are correct for the particular
+      robot model and configuration you intend to use them with.
+    </p>
+  </description>
+  <author>Wolfgang Merkt</author>
+  <author>Shaun Edwards</author><!-- Author of original SIA20D support package on which this one is based -->
+  <maintainer email="opensource@wolfgangmerkt.com">Wolfgang Merkt</maintainer>
```
gavanderhoorn: Are you sure you are going to be maintaining this?

motoman_sia50_support/package.xml
```diff
@@ -0,0 +1,62 @@
...
+    </p>
+    <p><strong>!!! THIS PACKAGE HAS NOT BEEN TESTED WITH A PHYSICAL ROBOT !!!</strong></p>
+    <p>
+      Before using any of the configuration files and / or meshes included
+      in this package, be sure to check they are correct for the particular
+      robot model and configuration you intend to use them with.
+    </p>
+  </description>
+  <author>Wolfgang Merkt</author>
+  <author>Shaun Edwards</author><!-- Author of original SIA20D support package on which this one is based -->
```
gavanderhoorn: Please remove.

URDFs/`.xacro`s are so 'minimal' that we don't retain authorship of originals in derivatives, especially not as you would've ended up with the same file if you'd use the templates in the support package generator.

motoman_sia50_support/package.xml
```diff
@@ -0,0 +1,62 @@
...
+  <maintainer email="opensource@wolfgangmerkt.com">Wolfgang Merkt</maintainer>
+  <!-- tbd: who from ROS-Industrial would be maintainer? -->
+  <!-- <maintainer email="shaun.edwards@gmail.com">Shaun Edwards</maintainer> -->
```
gavanderhoorn: Remove.

motoman_sia50_support/package.xml
```diff
@@ -0,0 +1,62 @@
...
+  <!-- <maintainer email="g.a.vanderhoorn@tudelft.nl">G.A. vd. Hoorn (TU Delft Robotics Institute)</maintainer> -->
```
gavanderhoorn:
```suggestion
  <maintainer email="g.a.vanderhoorn@tudelft.nl">G.A. vd. Hoorn (TU Delft Robotics Institute)</maintainer>
```

motoman_sia50_support/test/launch_test.xml
```diff
@@ -0,0 +1,57 @@
...
+       sub-namespaces under the main namespace.
+
+  Notes:
+
+  1. XML extension is used in order to avoid beinging included
+  in roslaunch auto-complete.
+
+  2. Group tags with namespaces are used to avoid node name collisions when
+  testing multpile launch files
+-->
```
gavanderhoorn: Remove

motoman_sia50_support/test/launch_test.xml
```diff
@@ -0,0 +1,57 @@
...
+<launch>
+	<arg name="req_arg" value="default"/>
+  <arg name="dx100" value="dx100"/>
```
gavanderhoorn: Unless you make `controller` a real `arg` again in the `.launch` files in which it is used, this can be removed.

wxmerkt:
> re: power requirements: yes, I'm somewhat jealous of @wxmerkt having access to an SIA50.

Alas, we don't have access to one (thus this is all simulation only). We need the model to replicate an experimental setup for comparison that uses a SIA50. The 2m+ workspace and 7-DoF class is quite a narrow subset of the overall market.

> Concerning the mass and other properties, I haven't gone into details but I can quickly tell you that the URB mass is over evaluated by more than 50%.
Thanks for the feedback, @EricMarcil. I have removed the inertial properties from the pull request. 

Looking forward to your other feedback on the pull request and happy to amend as required.

motoman_sia50_support/urdf/sia50d.xacro
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" ?>
+<robot name="motoman_sia50d" xmlns:xacro="http://ros.org/wiki/xacro">
+  <xacro:include filename="$(find motoman_sia50_support)/urdf/sia50d_macro.xacro" />
+  <xacro:motoman_sia50d prefix=""/>
+</robot>
```
gavanderhoorn: Please rename this file to remove the `d` suffix, as there is nothing DX100 specific in this file.

Make sure to update the `name` attribute of `robot` and the xacro macro name as well.

motoman_sia50_support/urdf/sia50d_macro.xacro
```diff
@@ -0,0 +1,209 @@
+<?xml version="1.0" ?>
```
gavanderhoorn: Rename this file as well.

motoman_sia50_support/urdf/sia50d_macro.xacro
```diff
@@ -0,0 +1,209 @@
...
+<robot xmlns:xacro="http://ros.org/wiki/xacro">
+    <xacro:macro name="motoman_sia50d" params="prefix">
```
gavanderhoorn: Remove the `d` suffix.

motoman_sia50_support/urdf/sia50d_macro.xacro
```diff
@@ -0,0 +1,209 @@
...
+        <!-- link list -->
+        <link name="${prefix}base_link">
+            <visual>
+                <geometry>
+                    <mesh filename="package://motoman_sia50_support/meshes/sia50d/visual/base_link.stl" />
```
gavanderhoorn: Please also remove the `d` suffix from the `meshes` sub dir.

motoman_sia50_support/urdf/sia50d_macro.xacro
```diff
@@ -0,0 +1,209 @@
...
+        <link name="${prefix}tool0" />
+        <!-- end of link list -->
+        <!-- joint list -->
+        <joint name="${prefix}joint_s" type="revolute">
+            <parent link="${prefix}base_link"/>
+            <child link="${prefix}link_s"/>
+            <origin xyz="0 0 0.312000" rpy="0 0 0"/>
+            <axis xyz="0 0 1" />
+            <!-- Max Acc/Dec Torque: 4903.32, Rated Motor Torque: 18.5 -->
+            <limit lower="-3.1415" upper="3.1415" effort="18.5" velocity="2.967051" />
```
gavanderhoorn: Please use the `radians(..)` conversion factor and specify limits in degrees (so they can be easily compared to the specs).

Additionally: `18.5` Nm for the first joint? This seems rather low.

If you want to include an `effort` limit here, you should probably list the other number (ie: maximum acc/dec torque).

motoman_sia50_support/urdf/sia50d_macro.xacro
```diff
@@ -0,0 +1,209 @@
...
+                <material name="yellow"/>
+            </collision>
+        </link>
+        <link name="${prefix}tool0" />
+        <!-- end of link list -->
+        <!-- joint list -->
+        <joint name="${prefix}joint_s" type="revolute">
+            <parent link="${prefix}base_link"/>
+            <child link="${prefix}link_s"/>
+            <origin xyz="0 0 0.312000" rpy="0 0 0"/>
```
gavanderhoorn: Is there a specific reason you have the Z translation in 6 significant digits here?

motoman_sia50_support/urdf/sia50d_macro.xacro
```diff
@@ -0,0 +1,209 @@
...
+            <axis xyz="0 0 1" />
+            <!-- Max Acc/Dec Torque: 4903.32, Rated Motor Torque: 18.5 -->
+            <limit lower="-3.1415" upper="3.1415" effort="18.5" velocity="2.967051" />
+        </joint>
+        <joint name="${prefix}joint_l" type="revolute">
+            <parent link="${prefix}link_s"/>
+            <child link="${prefix}link_l"/>
+            <origin xyz="0.145 0 0.228" rpy="0 0 0"/>
```
gavanderhoorn: Ideally `joint`s have translations in a single dimension only. None of the other `joint`s in your macro have this double translation.

Is it required?

motoman_sia50_support/urdf/sia50d_macro.xacro
```diff
@@ -0,0 +1,209 @@
...
+            <limit lower="-1.047198" upper="2.181662" effort="28.4" velocity="2.268928" />
+            <!-- 5883.99 -->
+        </joint>
+        <joint name="${prefix}joint_e" type="revolute">
+            <parent link="${prefix}link_l"/>
+            <child link="${prefix}link_e"/>
+            <origin xyz="0 0 0.875" rpy="0 0 0"/>
+            <axis xyz="0 0 1" />
+            <limit lower="-2.967064" upper="2.967064" effort="22" velocity="2.26893" />
+            <!-- 2549.73 -->
```
gavanderhoorn: Please remove these comments.

motoman_sia50_support/urdf/sia50d_macro.xacro
```diff
@@ -0,0 +1,209 @@
...
+        </joint>
+        <!-- end of joint list -->
+
+        <!-- ROS base_link (via base_link) to Motoman Robot (not Base) Frame transform -->
+        <!-- <link name="${prefix}base" />
+        <joint name="${prefix}base_link-base" type="fixed">
+            <origin xyz="0 0 0.410" rpy="0 0 0"/>
+            <parent link="${prefix}base_link"/>
+            <child link="${prefix}base"/>
+        </joint> -->
```
gavanderhoorn: Could you clarify why this is commented?

We should not merge any more macros without the `base` and `tool0` frames defined.

gavanderhoorn (changes requested):
Thanks again for the PR.

I've left some comments.

Overall: could you please check whether all files use spaces for indents, not tabs? Please use 2-space indents per level in all XML-based files.

motoman_sia50_support/package.xml
wxmerkt: I used the [image](https://www.motoman.com/getmedia/3fb46998-8d43-4afc-b6f0-b88cfe21e33b/sia50d_specs.jpg.aspx) on the "Specifications" tab. As thus, I don't have the document title/number/revision/ID. I also used a `.csv` sent by @EricMarcil for limits that did not come with a revision or version.

gavanderhoorn: All the documentation of Yaskawa is available free-of-charge on their website. It should not be difficult to locate the specifications of the SIA50 and mention the document ID here.

motoman_sia50_support/urdf/sia50d_macro.xacro
wxmerkt: It's due to this:

image

image

gavanderhoorn: This is solvable by placing `link_s` in the same ~~place~~ *plane* as `link_l`. That's beneficial, as that would also be the location of the `base` frame.

wxmerkt:
Thank you for the detailed review and feedback @gavanderhoorn. I've addressed the comments

EricMarcil:
Thanks for the PR.  It is greatly appreciated.
I've review it and everything looks good on the robot side.
However there is a "Tools" folder that I think is specific to your application that should be removed from the PR.

wxmerkt:
Thank you for the review and test on hardware. I am happy to remove the `tools` folder - it mirrors the processing scripts from `GP88`. Would you like me to remove it (it might be used to adapt/reproduce the meshes from the originals in the future)?

EricMarcil:
Note, I did not actually test it on hardware.  I just compared the values with my simulator (MotoSim).
As for the Tools folder, I'm wasn't aware of those processing scripts.  I just assumed it was related to your tooling or something.
@gavanderhoorn : When available, should the Tools folder be included in the support package?

gavanderhoorn (changes requested):
Overall this seems OK, but the collision meshes are a bit strange:

rviz_screenshot_2020_07_28-20_30_47

Did you have a particular reason to create them like this @wxmerkt?

If not, I'd recommend to create convex hulls using Meshlab from the `visual` meshes.

Additionally, there are two meshes with missing faces.

`link_l`:

rviz_screenshot_2020_07_28-20_33_59

rviz_screenshot_2020_07_28-20_34_15

and `link_b`:

rviz_screenshot_2020_07_28-20_36_15

Those meshes will have to be fixed before we can merge.

gavanderhoorn:
@EricMarcil wrote:
> As for the Tools folder, I'm wasn't aware of those processing scripts. I just assumed it was related to your tooling or something.
> @gavanderhoorn : When available, should the Tools folder be included in the support package?

We can keep them: they (partially) document how the meshes were generated/converted from the original ones. Assuming @wxmerkt actually used them, of course.

EricMarcil:
@wxmerkt : I've managed to find another model that wasn't missing faces in the stl mesh.  I've regenerated the meshes, so if you want to substitute them to yours and confirm that everything is displaying properly.
[SIA50D STL.zip](https://github.com/ros-industrial/motoman/files/5091949/SIA50D.STL.zip)

gavanderhoorn:
@wxmerkt: friendly ping.

It's ok if you can't spend any more time on this. If you can let us now, I'll take over and fix the meshes.

wxmerkt:
Hi @gavanderhoorn,
Sorry for the silence on this - I can't spend any more time on this right now - thank you for the offer to take over and complete the PR. Regarding the earlier questions: I used the scripts in the processing folder. They use Meshlab under the hood as with the folder I amended them from (GP88). The artefacts in the collision meshes thus must be from needing to amend the settings (I used the very latest Meshlab).

Many thanks for all your input, time and help with this PR

gavanderhoorn:
I've fixed all outstanding issues with the meshes (thanks @EricMarcil for the new set) and added a few more additional commits.

Once Travis gives us the green checkmark, I'll squash-merge.

gavanderhoorn:
And in the end I've removed the `tools` folder, as the meshes we're finally using were not produced using those scripts.

gavanderhoorn (approved):
With all outstanding issues fixed I believe we can merge this.

gavanderhoorn:
@wxmerkt: thanks for the (initial) PR and thanks for iterating.

@EricMarcil: thanks for the meshes and overall checks and guidance.

gavanderhoorn:
@wxmerkt: due to the use of a squash-merge, it doesn't show, but you could remove your `wxm-add-motoman-sia50d-support` branch.

PR #339, opened 2020-06-30
HC20 Support Package

EricMarcil:
Added package for the HC20 (DT/XP) robot model


gavanderhoorn (changes requested):
Overall this looks OK, just a question though: seeing as the robot is almost completely white, why the blue in the xacro macro? Would it make sense to make the robot white instead?

gavanderhoorn:
@EricMarcil: I've added a few fixup commits.

I believe only 9798b9b is really important: previously the `collision` mesh for `base_link` was actually a copy of the `collision` mesh of `link_1_s.stl`. Probably a mistake:

rviz_screenshot_2020_07_28-19_51_33

I created a new convex hull using Meshlab and saved that as `collision/base_link.stl`.

EricMarcil:
Thanks for the review and fixing the base collision model.
I've just push a commit to update the color.
Everything looks good to go.

gavanderhoorn (approved):
Thanks for adding support for the HC20 :+1: 

And thanks for iterating.

gavanderhoorn:
I'll squash-merge this, to prevent my fixes from ending up in the commit history.

gavanderhoorn:
@EricMarcil: make sure to delete the `HC20` branch from your fork. Because of the squash-merge it will look like it hasn't been merged, but it has.

EricMarcil:
@gavanderhoorn Thanks for the review and merge.

PR #344, opened 2020-07-19
GP50 Support Package

EricMarcil:
Added package for GP50 robot model


gavanderhoorn (approved):
This is OK to merge like this.

gavanderhoorn:
@EricMarcil: I've added some fixup commits.

We'll squash-merge this, as fi the duplicated meshes should not become part of the repository history here.

PR #346, opened 2020-07-28
sda10: fix group mapping

gavanderhoorn:
As per subject.

The groups are actually `0, 1, 2, 3`, not `1, 2, 3, 4`.


gavanderhoorn:
@marip8: could you please take a quick look?

marip8 (approved):

PR #348, opened 2020-07-30
ES165 Support

marip8:
Initial commit for the ES165RDII robot with contributions from @gavanderhoorn


motoman_es_support/package.xml
```diff
@@ -0,0 +1,49 @@
...
+      robot model and configuration you intend to use them with.
+    </p>
+  </description>
+  <author>Michael Ripperger</author>
+  <maintainer email="michael.ripperger@swri.org">Michael Ripperger</maintainer>
+  <license>Apache 2.0</license>
+
+  <url type="website">http://wiki.ros.org/motoman_es165rdii_support</url>
+  <url type="bugtracker">https://github.com/ros-industrial/motoman_experimental/issues</url>
+  <url type="repository">https://github.com/ros-industrial/motoman_experimental</url>
```
gavanderhoorn:
```suggestion
  <url type="bugtracker">https://github.com/ros-industrial/motoman/issues</url>
  <url type="repository">https://github.com/ros-industrial/motoman</url>
```

motoman_es_support/package.xml
```diff
@@ -0,0 +1,49 @@
...
+      Before using any of the configuration files and / or meshes included
+      in this package, be sure to check they are correct for the particular
+      robot model and configuration you intend to use them with.
+    </p>
+  </description>
+  <author>Michael Ripperger</author>
+  <maintainer email="michael.ripperger@swri.org">Michael Ripperger</maintainer>
+  <license>Apache 2.0</license>
+
+  <url type="website">http://wiki.ros.org/motoman_es165rdii_support</url>
```
gavanderhoorn:
```suggestion
  <url type="website">http://wiki.ros.org/motoman_es_support</url>
```

gavanderhoorn (changes requested):
Looks OK.

Two in-line suggestions.

motoman_es_support/launch/load_es165rdii.launch
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" ?>
+<launch>
+  <param name="robot_description" command="$(find xacro)/xacro.py --inorder '$(find motoman_es_support)/urdf/es165rdii.xacro'" />
```
gavanderhoorn:
```suggestion
  <param name="robot_description" command="$(find xacro)/xacro --inorder '$(find motoman_es_support)/urdf/es165rdii.xacro'" />
```

marip8:
Given the discussion in #349 which indicates that robot geometry does not typically change between controller variants of a robot, should I re-title these files to something that is not controller specific (i.e. ES165R) and provide controller-specific joint velocity limits?

gavanderhoorn:
> Given the discussion in #349 which indicates that robot geometry does not typically change between controller variants of a robot, should I re-title these files to something that is not controller specific (i.e. ES165R) and provide controller-specific joint velocity limits?

We could, but without using a `.yaml` file I don't see how we would do this in a nice and scalable way. And adding support for `.yaml` files here would be a bit involved.

Would you have any idea immediately?

Or would you want to create separate `_macro.xacro`s? That would only lead to some duplication, but would be straightforward to implement -- and it would align with the "store variants of the same robot model in a single package": the controller would essentially become part of the variant "identification" (which it is, according to @ted-miller in #349).


Otherwise let's not complicate things any further.

marip8:
It wasn't too difficult to implement the YAML files for the specific controller velocity limits. They get loaded into the URDF via launch files where the user specifies the `controller` argument to find the controller-specific file [here](https://github.com/marip8/motoman/blob/12824bfbeb06d40f9a90fd90c073333de3db5261/motoman_es_support/launch/load_es165.launch#L5). Let me know what you think of this approach

marip8:
@gavanderhoorn friendly ping

gavanderhoorn (approved):

gavanderhoorn:
Thanks for the PR @marip8, appreciated.

Apologies for the delayed merge.

It might be nice to add some convenience `.launch` files which provide the correct argument for `controller`. They could be named after the variant. But we don't need to do that now.

PR #349, opened 2020-07-30
ES200-120 Support

marip8:
Adds support for ES200RN-120 robot variant. @EricMarcil @ted-miller can you confirm that RN-120 is the correct name of this variant? I seem to have conflicting documentation from the project in which we purchased this robot where some documents say RN-120 and others that says RDII-120

Note: the URDF for this robot has been tested/verified on real hardware in our lab

Builds on #348


gavanderhoorn (changes requested):
Looks OK.

One comment: 👍 for the `roslaunch` test, but I believe it needs to be added to the `CMakeLists.txt`?

ted-miller:
> can you confirm that RN-120 is the correct name of this variant? I seem to have conflicting documentation from the project in which we purchased this robot where some documents say RN-120 and others that says RDII-120

The RN is for the NX100 controller. The RD is for the DX100 controller. The RDII is for the DX200 controller.

gavanderhoorn:
@ted-miller wrote:
> > can you confirm that RN-120 is the correct name of this variant? I seem to have conflicting documentation from the project in which we purchased this robot where some documents say RN-120 and others that says RDII-120
> 
> The RN is for the NX100 controller. The RD is for the DX100 controller. The RDII is for the DX200 controller.

Are the robots themselves identical?

ted-miller:
No, the robot generations are not identical.

compare

gavanderhoorn:
Hm. Thanks for the information.

What about the geometry of the robot? Has that changed between the generations?

I realise I could check this myself. Let me know when you get *fed up* with our questions :)

marip8:
If the kinematics are the same, I can structure this package to read joint velocities (and joint limits if necessary) from YAML files such that the three variants can use the same URDF but different joint velocity/limit configurations.

ted-miller:
I just realized that my comparison above doesn't use the "-120" variant for the DX200. I cannot find the documentation for the RDII-120. This "probably" means that the RDII is the same as the RD... but I need to check with some colleagues to confirm.

As for the geometry, I'm not entirely certain how to verify this. But, this diagram compares the work envelopes of the RN-120 and RD-120. I believe they are the same.
rn-120 and rd-120

ted-miller:
From a colleague, regarding the RD-120 vs RDII-120:
"Generally the II series is the same as it’s earlier version.  Sometimes the number of manipulator cables were reduced, or RoHS compliance was added reducing some chemicals in use.  The robot performance and dimensions should be the same."

gavanderhoorn (approved):

gavanderhoorn:
Thanks for iterating @marip8 :+1:

PR #355, opened 2020-08-28
Rebase of MotomanIoCtrl on-top of current HEAD - initial IO support (ROS side)

### LINKED ISSUE #313: YRC1000micro with discrete input output gripper
Looking at the [simple_message folder](https://github.com/ros-industrial/motoman/tree/kinetic-devel/motoman_driver/src/simple_message) inside the motoman_driver folder, it appears we can send commands and control a discrete I/O gripper (e.g. [EGP40-N-N-B](https://schunk.com/us_en/gripping-systems/product/28088-0310940-egp-40-n-n-b/)) using these files. Can anyone confirm this posit? Is there ROS documentation that explains this further anyone can please suggest?

steviedale:
This pull request adds support for reading/writing single IO signals for single and multi motion groups.

@marip8 and I were able to test it out on dx200 controller with a single motion group robot.
- Controller software version: DN2.33.00A (JP/US)-00
- MotoROS version: ??
    - @gavanderhoorn @ted-miller how do we determine what version is currently running on the controller?

Notes:
- We were able to read/write universal inputs and outputs, but we were not able to read/write other types (i.e. external, specific, control, network, etc.). @gavanderhoorn is this the intended behavior?
- Leading zeros on addresses were not being interpreted correctly. Example: service request address = 00010 was interpreted as 9. When we put the actual number in there (10 in this case) everything works as expected.

### DIFF
--- a/motoman_msgs/srv/ReadSingleIO.srv
+++ b/motoman_msgs/srv/ReadSingleIO.srv
@@ -1,0 +1,3 @@
+uint32 address
+---
+int32 value
--- a/motoman_msgs/srv/ReadSingleIO.srv
+++ b/motoman_msgs/srv/ReadSingleIO.srv
@@ -1,3 +1,15 @@
+
+# Read (and return) the current value of the IO element at 'address'.
+#
+# Addresses are plain, base-10 integers, as used and displayed by the controller
+# (on the teach pendant for instance).
+#
+# There are no restrictions as to which IO elements can be read, but they have
+# to exist on the controller and be configured correctly.
+#
+# Refer also the Yaskawa Motoman documentation on IO addressing and
+# configuration.
+
 uint32 address
 ---
 int32 value
--- a/motoman_driver/CMakeLists.txt
+++ b/motoman_driver/CMakeLists.txt
@@ -119,7 +119,8 @@ set_target_properties(motoman_robot_state
 add_executable(motoman_motion_streaming_interface
   src/joint_streaming_node.cpp
   src/joint_trajectory_streamer.cpp
-  src/motion_ctrl.cpp)
+  src/motion_ctrl.cpp
+  src/io_ctrl.cpp)
 target_link_libraries(motoman_motion_streaming_interface
   motoman_simple_message
   motoman_industrial_robot_client
@@ -160,7 +161,8 @@ set_target_properties(motoman_robot_state_bswap
 add_executable(motoman_motion_streaming_interface_bswap
   src/joint_streaming_node.cpp
   src/joint_trajectory_streamer.cpp
-  src/motion_ctrl.cpp)
+  src/motion_ctrl.cpp
+  src/io_ctrl.cpp)
 target_link_libraries(motoman_motion_streaming_interface_bswap
   motoman_simple_message_bswap
   motoman_industrial_robot_client_bswap
--- a/motoman_driver/include/motoman_driver/io_ctrl.h
+++ b/motoman_driver/include/motoman_driver/io_ctrl.h
@@ -1,0 +1,101 @@
+﻿/*
+ * Software License Agreement (BSD License)
+ *
+ * Copyright (c) 2016, Delft Robotics Institute
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *  * Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  * Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *  * Neither the name of the Delft Robotics Institute, nor the names
+ *    of its contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * \author G.A. vd. Hoorn (TU Delft Robotics Institute)
+ */
+
+#ifndef MOTOMAN_DRIVER_IO_CTRL_H
+#define MOTOMAN_DRIVER_IO_CTRL_H
+
+#include "simple_message/smpl_msg_connection.h"
+#include "motoman_driver/simple_message/motoman_read_single_io.h"
+#include "motoman_driver/simple_message/motoman_read_single_io_reply.h"
+#include "motoman_driver/simple_message/motoman_write_single_io.h"
+#include "motoman_driver/simple_message/motoman_write_single_io_reply.h"
+
+namespace motoman
+{
+namespace io_ctrl
+{
+using industrial::smpl_msg_connection::SmplMsgConnection;
+using motoman::simple_message::io_ctrl_reply::ReadSingleIOReply;
+using motoman::simple_message::io_ctrl_reply::WriteSingleIOReply;
+
+/**
+ * \brief Wrapper class around Motoman-specific io control commands
+ */
+
+class MotomanIoCtrl
+{
+public:
+  /**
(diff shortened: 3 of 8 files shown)


gavanderhoorn:
Thanks for the PR @steviedale.

> * Leading zeros on addresses were not being interpreted correctly. Example: service request address = 00010 was interpreted as 9. When we put the actual number in there (10 in this case) everything works as expected.

That's ~entirely expected~ possible, as numbers with leading zeros ~are~ could be interpreted as octal numbers. And 10 octal == 8 decimal, so with a 0-based indexing IO system, gets you IO element 9.

We'd have to figure out which part of the infrastructure does that though. I would guess `rosservice call` -- if you used it, that is.

ted-miller:
> how do we determine what version is currently running on the controller?

From the programming pendant touch screen, touch [System Info] > [Version]. On the pendant keypad, press the [PAGE] key two or three times. (I don't remember the exact number)

> We were able to read/write universal inputs and outputs, but we were not able to read/write other types (i.e. external, specific, control, network, etc.)

You should be able to directly write to Network Inputs (#27xxx) (or #25xxx on DX100/FS100) or Universal/General Outputs (#10xxx). Nothing else is directly writable. 

However, there may be a link in your Concurrent I/O Ladder Program (CIOPRG.LST) that links the writable I/O to other signals. So, you can indirectly write to anything.

ted-miller:
The restrictions that I mentioned above only apply to writing. All I/O is readable.

marip8:
> > how do we determine what version is currently running on the controller?
>
> From the programming pendant touch screen, touch [System Info] > [Version]. On the pendant keypad, press the [PAGE] key two or three times. (I don't remember the exact number)

Thanks for the instructions. The controller we tested on is running MotoROS 1.5.0. @gavanderhoorn @ted-miller if you plan on testing this PR with a controller that is running 1.9.0, then we might not update our controller right now. Otherwise, we can go through that process to do testing on the latest version

> >We were able to read/write universal inputs and outputs, but we were not able to read/write other types (i.e. external, specific, control, network, etc.)
>
> You should be able to directly write to Network Inputs (#27xxx) (or #25xxx on DX100/FS100) or Universal/General Outputs (#10xxx). Nothing else is directly writable.
>
> However, there may be a link in your Concurrent I/O Ladder Program (CIOPRG.LST) that links the writable I/O to other signals. So, you can indirectly write to anything.

On our controller, the last number of several of the non-universal I/O was the letter X (i.e. #2700X). I'm not entirely sure what this means, but the I/O services in this PR expect the I/O address to be an integer

gavanderhoorn:
> The controller we tested on is running MotoROS 1.5.0. @gavanderhoorn @ted-miller if you plan on testing this PR with a controller that is running 1.9.0, then we might not update our controller right now.

So I'm running the absolute latest version of MotoROS, but my controller is currently not in a state I can turn it on.

Upgrading really comes down to just replacing the `.out` file, and you can very easily go back. Should not take more than a few controller reboots (you don't need any motion to test the IO integration).

marip8:
> Upgrading really comes down to just replacing the .out file, and you can very easily go back. Should not take more than a few controller reboots (you don't need any motion to test the IO integration).

This should be easy enough to do. We can test early next week with the latest version

gavanderhoorn:
Friendly ping @steviedale and @marip8.

steviedale:
@gavanderhoorn Sorry for the delay.

@marip8 and I just tested out the IO services on a Motoman robot using MotoROS 1.9.1. It looks like the /read_single_io service is working as expected. However, the /write_single_io service returns an error and does not write.

Here's the error message from calling /write_single_io:
```[ERROR] [1601063639.735533492]: Writing IO element 16 failed```

We also noticed a warning that appears when we launch the driver. We think this might be contributing to the writing problem since we don't remember seeing this warning when we were testing on the older version of MotoROS. Here's that warning:
```
[ WARN] [1601063636.675636797]: Tried to connect when socket already in connected state
```

Do you have any ideas about what might be causing the /write_single_io service to return an error?

**Comment by ted-miller** [comment]
> Here's the error message from calling /write_single_io:
[ERROR] [1601063639.735533492]: Writing IO element 16 failed

Can you get a wireshark capture when you attempt to write the signal?

**Comment by gavanderhoorn** [comment]
ping @marip8 / @steviedale : would you happen to have a wireshark trace? It would help diagnose.

In the meantime my controller is back to working order, so I'll try out this PR as well.

**Comment by gavanderhoorn** [comment]
I just tested this and even though I don't fully understand what the controller returns sometimes, reading IOs does seem to work.

```bash
$ rosservice call /read_single_io "address: 10"
value: 0
$ rosservice call /read_single_io "address: 30017"
value: 0
```

etc.

It's not very fast, but that's to be expected (this is not an interface which should be used if determinism and/or certain performance are a requirement).

Writing also works, if I use the addresses that @ted-miller mentions.

@ted-miller wrote:
> The restrictions that I mentioned above only apply to writing. All I/O is readable.

initially I also thought this was true, but the following doesn't work for me:

```bash
$ rosservice call /read_single_io "address: 27001"
ERROR: service [/bla/read_single_io] responded with an error: 
```

The driver logs the following (I've added some extra debug statements):

```
[DEBUG] [1601993320.692144432] [/motion_streaming_interface] [ros.motoman_driver.MotomanIoCtrl]: request to read IO element: 27001
[DEBUG] [1601993320.693725799] [/motion_streaming_interface] [ros.motoman_driver.MotomanIoCtrl]: received reply: result_code: 2, value: 10385
[ERROR] [1601993320.693777762] [/motion_streaming_interface] [ros.motoman_driver]: Reading IO element 27001 failed
```

`result_code` is set to `2`, which is `SmReplyType::ROS_REPLY_FAILURE`.

I'm actually not sure what the `10385` is. It is consistently reported when I attempt to read from that address though and is also shows in the Wireshark trace.

Initially I thought it was a random value in the controllers memory, but repeated calls return the same value.

gavanderhoorn:
@steviedale @marip8: this is with MotoROS `v1.9.1`, which has kept the bw-compatibility wrt where IO requests are being handled.

In the follow-up PRs we should however introduce the stand-alone IO client we've discussed.

At that point we can remove the bw-compatibility infrastructure from the `MotionServer` on the MotoROS side.

ted-miller:
> the following doesn't work for me:
$ rosservice call /read_single_io "address: 27001"
ERROR: service [/bla/read_single_io] responded with an error: 

27001 is not a valid address. The first valid address is 27010.

gavanderhoorn:
Ah, that would explain it.

Is `10385` a specific error code returned by MotoPlus?

ted-miller:
No. I suspect it's just garbage data. The value is not being initialized to zero before the API call, nor is it being modified after the API returns a failure. See 'ioValue':

https://github.com/ros-industrial/motoman/blob/47ef63fe1f30cad1de00ead44a2479f027ecc8b9/motoman_driver/MotoPlus/IoServer.c#L310-L339

gavanderhoorn:
Ok, as I thought initially then.

I'll log it as an issue, so we can track it.

We might have stated somewhere that returned data cannot be trusted whenever the return is a `ROS_REPLY_FAILURE`, but it would still be nice to initialise that variable and prevent potential confusion.

---

Logged #358.

gavanderhoorn:
@marip8 @steviedale: if you could confirm things work for you as well when taking the (expected) limitations into account, we could probably merge.

gavanderhoorn:
@steviedale: it seems the email address you've used for the commits here in this PR is not registered with Github.

Would you want to fix that before merging?

motoman_msgs/srv/ReadSingleIO.srv
```diff
@@ -0,0 +1,3 @@
+uint32 address
+---
+int32 value
```
gavanderhoorn, line 15: It would be nice if we could add some comments to this message (and its write counterpart) which provide info on which addresses can actually be read from (all *valid* addresses) and which can be written to (see @ted-miller's comment).

marip8: Assuming here you mean [this comment](https://github.com/ros-industrial/motoman/pull/355#issuecomment-683132425)? @ted-miller do the valid addresses for all I/O start at `xxx10` rather than `xxxx1`? I seem to remember that being true for the universal inputs on our robot too

ted-miller: > do the valid addresses for all I/O start at xxx10 rather than xxxx1? I seem to remember that being true for the universal inputs on our robot too

Yes.

steviedale:
@gavanderhoorn, @marip8 and I just tested out the read and write services with the latest changes from ```kinetic-devel``` and wrt the address constraints discussed above. Both services now work as expected.

motoman_msgs/srv/ReadSingleIO.srv
steviedale: I just added a comment to the WriteSingleIO.srv definition describing the available addresses (using the information from [this comment](https://github.com/ros-industrial/motoman/pull/355#issuecomment-683132425)). However, I'm having trouble understanding which addresses are valid for reading?

steviedale:
> @steviedale: it seems the email address you've used for the commits here in this PR is not registered with Github.
> 
> Would you want to fix that before merging?

@gavanderhoorn thanks for the heads up! I think I fixed it now.

motoman_msgs/srv/ReadSingleIO.srv
gavanderhoorn: Essentially *all* addresses are readable, as long as they are valid addresses.

Even though that doesn't really say too much, I still feel like a small comment saying something to that effect will help avoid at least *some* confusion. We can always say something like "Refer to the relevant documentation from Yaskawa for information on which addresses are valid addresses on a specific controller with a specific IO configuration".

gavanderhoorn:
I've added a commit expanding the comment(s) a little for both service definitions.

Let's wait on Travis and once we get a green checkmark, we'll merge.

---

Edit: to preserve history, we'll use a regular merge this time.

gavanderhoorn (approved):
Pending Travis this is OK to merge.

We'll improve on it in follow-up PRs.

gavanderhoorn:
Seems Travis is happy, and I'm happy, so we'll merge.

gavanderhoorn:
Thanks @steviedale for iterating, and thanks @marip8 as well.

Thanks @ted-miller for clearing up some things.

@steviedale & @marip8: should we schedule a call to discuss the next MS?

marip8:
> @steviedale & @marip8: should we schedule a call to discuss the next MS?

Sounds good to me. I'll send an email

marip8:
Addresses first milestone in #361

Bdelspi:
Thank you everyone for all your help with this!!!

Issue #361, opened 2020-10-13
I/O Interface Development


marip8:
Opening this thread as a discussion point for the development of I/O interface capabilities from the ROS driver

## Milestones
The tasks below are the currently planned high-level milestones for the I/O interface

- [x] Add I/O communication into joint trajectory streamer node (done: #355)
- [x] Port I/O communication into stand-alone node (done: #365)
- [x] Incorporate better error feedback for incorrect I/O addresses per #360 (started with #366 and #371 (and #369))
- [x] Add support for reading/writing bytes (#435)

## Future Development
- [ ] Implement I/O solution for ROS 2 driver
- [ ] Leverage high-speed ethernet server for I/O communication with complex types (i.e. strings, etc.) and other teach pendant functionality

---

Edit: not on the original list of work, but:

 - [x] Add support for reading/writing M-registers (#439)

gavanderhoorn:
@marip8 wrote:
> Add support for multi-group I/O reading/writing

what is this exactly @marip8 ?

Reading/writing bytes and words, or getting everything to work in both the single and multi-motion group code paths?

marip8:
I meant I/O support for bytes/strings/etc. I just changed the wording above

gavanderhoorn:
Ok.

Suggestion: split this into (at least) two items, as reading/writing bytes is already supported by/through MotoROS (see also [REP-I4: Assigned Message Identifiers for the Simple Message Protocol - Assigned Message Identifiers - Vendor Specific Messages - Motoman](https://github.com/ros-industrial/rep/blob/bbd88b9b1dc9b61be5178c8b3a93b9bec488702a/rep-I0004.rst#motoman)), while variables, strings and others like that would probably need the ethernet server.

cjue:
> Suggestion: split this into (at least) two items, as reading/writing bytes is already supported by/through MotoROS

Yes, just adding read/write byte/word support would be helpful. IMHO having read/write_byte/word_io services would already save time and complexity for simple applications.

gavanderhoorn:
While convenience is why we're adding them here, I'm going to repeat my advice (which I always give when talking about ROS, robot drivers and IO): consider using a proper fieldbus for industrial deployments. Having to add IO support to each and every robot driver doesn't scale very well, and will most likely use custom implementations of protocols which do not have the same level of support, integration nor testing as any of the available (software defined) fieldbuses in existence.

Even a simple wrapper around a modbus/tcp client would be a good option.

[...]

Issue #362, opened 2020-10-15
Consider adding 'success' and 'message' fields to IO services


gavanderhoorn:
This would be similar to [std_srvs/Trigger](http://docs.ros.org/api/std_srvs/html/srv/Trigger.html) and would allow communication of failure causes to service callers.

Ideally MotoROS would also provide more information than a simple failure code, but I'm not sure whether that would be possible.

This could be added both to the read as well as the write service, as both have overlapping failure modes.

gavanderhoorn:
With #366 merged, this can be closed.

We'll improve the error messages returned in a follow-up PR (part of #361).

Issue #360, opened 2020-10-12
Catch writes to 'unwritable' addresses on the ROS side and print informative error


gavanderhoorn:
With #355 basic read/write of individual IOs was added to `motoman_driver`.

As mentioned in the `.srv`, writing is only supported to certain ranges of addresses:

https://github.com/ros-industrial/motoman/blob/591a09c5cb95378aafd02e77e45514cfac3a009d/motoman_msgs/srv/WriteSingleIO.srv#L9-L12

Instead of letting writes requests to other addresses through, the driver should perhaps check which address is targeted and implement a fail-early policy (don't even forward the request to MotoROS).

It could then also print a better error message than just the current:

https://github.com/ros-industrial/motoman/blob/591a09c5cb95378aafd02e77e45514cfac3a009d/motoman_driver/src/joint_trajectory_streamer.cpp#L600

a more informative error would include why the write failed.

gavanderhoorn:
@ted-miller @EricMarcil: would you happen to know the ranges which are writable, or should we look it up in the manual(s)?

gavanderhoorn:
Related ROS Answers Q&A: [motoman_driver external IO writing](https://answers.ros.org/question/363259).

gavanderhoorn:
From the DX200 manual:

![sshot](https://user-images.githubusercontent.com/4550046/95773932-b4690380-0cbf-11eb-9cd4-465506eff721.jpg)

So that would be `[27010; 29567]` for Network Inputs and `[10010; 15127]` for General Outputs.

Difficulty would be to "know" which controller is on the other end. As for FS100/DX100 these numbers are different.

ted-miller:
DX100:
10010 - 12567
25010 - 27567

FS100: 
10010 - 11287
25010 - 26287

DX200: listed in your image above
YRC1000 and micro: Same as DX200.

ted-miller:
Also, keep in mind that the last digit will never exceed 7.

10010 - 10017 = Valid
10018 - 10019 = NOT VALID
10020 - 10027 = Valid

gavanderhoorn:
@ted-miller: thanks.

@marip8 @steviedale: if/when we introduce the stand-alone IO relay we can provide it with the information on the connected controller, so as to filter for the correct ranges.

gavanderhoorn:
@ted-miller: the DX200 *Reference Manual For New Language Environment MotoPlus* (`166580-1CD`) suggests the *interface panel* IOs are also writable:

[...]

PR #365, opened 2020-10-20
Update/separate io control from joint streamer

steviedale:
This pull request addresses the need to move I/O control out of the ```motion_interface``` node and into a new ```io_relay``` node.

This is for #361.

### DIFF
--- a/motoman_driver/include/motoman_driver/io_relay.h
+++ b/motoman_driver/include/motoman_driver/io_relay.h
@@ -54,11 +54,10 @@ public:
   /**
    * \brief Class initializer
    *
-   * \param default_ip
    * \param default_port
    * \return true on success, false otherwise
    */
-  bool init(std::string default_ip, int default_port);
+  bool init(int default_port);
 
 protected:
   io_ctrl::MotomanIoCtrl io_ctrl_;
--- a/motoman_driver/include/motoman_driver/io_relay.h
+++ b/motoman_driver/include/motoman_driver/io_relay.h
@@ -55,10 +55,10 @@ public:
    * \brief Class initializer
    *
    * \param default_ip
-   * \param version_0
+   * \param default_port
    * \return true on success, false otherwise
    */
-  bool init(std::string default_ip, bool version_0);
+  bool init(std::string default_ip, int default_port);
 
 protected:
   io_ctrl::MotomanIoCtrl io_ctrl_;
--- a/motoman_driver/include/motoman_driver/io_relay.h
+++ b/motoman_driver/include/motoman_driver/io_relay.h
@@ -46,7 +46,7 @@ namespace io_relay
 using industrial::tcp_client::TcpClient;
 
 /**
- * \brief Message handler that streams I/O control to the robot controller.
+ * \brief Message handler that sends I/O service requests to the robot controller and receives the responses.
  */
 class MotomanIORelay
 {
--- a/motoman_driver/include/motoman_driver/io_relay.h
+++ b/motoman_driver/include/motoman_driver/io_relay.h
@@ -1,0 +1,82 @@
+﻿/*
+ * Software License Agreement (BSD License)
+ *
+ * Copyright (c) 2013, Southwest Research Institute
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *       * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *       * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *       * Neither the name of the Southwest Research Institute, nor the names
+ *       of its contributors may be used to endorse or promote products derived
+ *       from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef MOTOMAN_DRIVER_IO_RELAY_H
+#define MOTOMAN_DRIVER_IO_RELAY_H
+
+#include "simple_message/socket/tcp_client.h"
+#include "motoman_driver/io_ctrl.h"
+#include "motoman_msgs/ReadSingleIO.h"
+#include "motoman_msgs/WriteSingleIO.h"
+#include <boost/thread.hpp>
+
+namespace motoman
+{
+namespace io_relay
+{
+
+using industrial::tcp_client::TcpClient;
+
+/**
+ * \brief Message handler that streams I/O control to the robot controller.
+ */
+class MotomanIORelay
+{
+public:
+  /**
+   * \brief Class initializer
+   *
+   * \param default_ip
+   * \param version_0
+   * \return true on success, false otherwise
+   */
+  bool init(std::string default_ip, bool version_0);
+
+protected:
+  io_ctrl::MotomanIoCtrl io_ctrl_;
+
+  ros::ServiceServer srv_read_single_io;   // handle for read_single_io service
+  ros::ServiceServer srv_write_single_io;   // handle for write_single_io service
+
+  ros::NodeHandle node_;
+  boost::mutex mutex_;
+  TcpClient default_tcp_connection_;
+
+  bool readSingleIoCB(motoman_msgs::ReadSingleIO::Request &req,
+                            motoman_msgs::ReadSingleIO::Response &res);
+  bool writeSingleIoCB(motoman_msgs::WriteSingleIO::Request &req,
+                            motoman_msgs::WriteSingleIO::Response &res);
+};
+
+}  // namespace io_relay
+}  // namespace motoman
+
+#endif  // MOTOMAN_DRIVER_IO_RELAY_H
--- a/motoman_driver/include/motoman_driver/io_relay.h
+++ b/motoman_driver/include/motoman_driver/io_relay.h
@@ -1,7 +1,7 @@
 ﻿/*
  * Software License Agreement (BSD License)
  *
- * Copyright (c) 2013, Southwest Research Institute
+ * Copyright (c) 2020, Southwest Research Institute
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
--- a/motoman_driver/launch/io_relay.launch
+++ b/motoman_driver/launch/io_relay.launch
@@ -5,8 +5,15 @@
   <!-- IP of robot (or PC running simulation) -->
   <arg name="robot_ip" doc="IP of controller" />
 
+  <!-- Load the byte-swapping version of io_relay if required -->
+  <arg name="use_bswap" doc="If true, robot driver will byte-swap all incoming and outgoing data" />
+
   <!-- put them on the parameter server -->
   <param name="robot_ip_address" type="str" value="$(arg robot_ip)" />
 
-  <node name="io_relay" pkg="motoman_driver" type="io_relay" />
+  <!-- load the correct version of the io relay node -->
+  <node if="$(arg use_bswap)" name="io_relay"
+        pkg="motoman_driver" type="io_relay_bswap" />
+  <node unless="$(arg use_bswap)" name="io_relay"
+        pkg="motoman_driver" type="io_relay" />
 </launch>
--- a/motoman_driver/launch/io_relay.launch
+++ b/motoman_driver/launch/io_relay.launch
@@ -2,11 +2,11 @@
   Wrapper launch file for I/O relay node.
 -->
 <launch>
-	<!-- IP of robot (or PC running simulation) -->
-	<arg name="robot_ip" doc="IP of controller" />
+  <!-- IP of robot (or PC running simulation) -->
+  <arg name="robot_ip" doc="IP of controller" />
 
-	<!-- put them on the parameter server -->
-	<param name="robot_ip_address" type="str" value="$(arg robot_ip)" />
+  <!-- put them on the parameter server -->
+  <param name="robot_ip_address" type="str" value="$(arg robot_ip)" />
 
-        <node name="io_relay" pkg="motoman_driver" type="io_relay" />
+  <node name="io_relay" pkg="motoman_driver" type="io_relay" />
 </launch>
--- a/motoman_driver/launch/io_relay.launch
+++ b/motoman_driver/launch/io_relay.launch
@@ -1,0 +1,12 @@
+<!-- 
+  Wrapper launch file for I/O relay node.
+-->
+<launch>
+	<!-- IP of robot (or PC running simulation) -->
+	<arg name="robot_ip" doc="IP of controller" />
+
+	<!-- put them on the parameter server -->
+	<param name="robot_ip_address" type="str" value="$(arg robot_ip)" />
+
+        <node name="io_relay" pkg="motoman_driver" type="io_relay" />
+</launch>
--- a/motoman_driver/src/io_relay.cpp
+++ b/motoman_driver/src/io_relay.cpp
@@ -46,7 +46,6 @@ bool MotomanIORelay::init(int default_port)
   std::string ip;
   int port;
 
-  ros::param::get("robot_ip_address", ip);
   // override port with ROS param, if available
   const std::string port_param_name = "~port";
   // TODO: should really use a private NodeHandle here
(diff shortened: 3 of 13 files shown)


motoman_driver/include/motoman_driver/io_relay.h
```diff
@@ -0,0 +1,82 @@
+﻿/*
+ * Software License Agreement (BSD License)
+ *
+ * Copyright (c) 2013, Southwest Research Institute
```
gavanderhoorn: I would suggest to update the copyright here.

motoman_driver/include/motoman_driver/io_relay.h
```diff
@@ -0,0 +1,82 @@
...
+
+namespace motoman
+{
+namespace io_relay
+{
+
+using industrial::tcp_client::TcpClient;
+
+/**
+ * \brief Message handler that streams I/O control to the robot controller.
```
gavanderhoorn:
```suggestion
 * \brief Message handler that sends I/O service requests to the robot controller and receives the responses.

```

motoman_driver/include/motoman_driver/io_relay.h
```diff
@@ -0,0 +1,82 @@
...
+{
+public:
+  /**
+   * \brief Class initializer
+   *
+   * \param default_ip
+   * \param default_port
+   * \return true on success, false otherwise
+   */
+  bool init(std::string default_ip, int default_port);
```
gavanderhoorn: Suggestion: remove the `default_ip`. Not sure why we have that, as it makes no sense (in contrast to the `default_port`).

motoman_driver/launch/io_relay.launch
```diff
@@ -0,0 +1,12 @@
...
+-->
+<launch>
+	<!-- IP of robot (or PC running simulation) -->
+	<arg name="robot_ip" doc="IP of controller" />
+
+	<!-- put them on the parameter server -->
+	<param name="robot_ip_address" type="str" value="$(arg robot_ip)" />
+
+        <node name="io_relay" pkg="motoman_driver" type="io_relay" />
+</launch>
```
gavanderhoorn, line 19: Please check indentation in this file: 2-spaces per level, no tabs.

motoman_driver/launch/io_relay.launch
```diff
@@ -0,0 +1,12 @@
+<!-- 
+  Wrapper launch file for I/O relay node.
+-->
+<launch>
+	<!-- IP of robot (or PC running simulation) -->
+	<arg name="robot_ip" doc="IP of controller" />
+
+	<!-- put them on the parameter server -->
+	<param name="robot_ip_address" type="str" value="$(arg robot_ip)" />
```
gavanderhoorn: Suggestion: make this a private parameter of the node instead.

I know that's not how the trajectory streamer and the state node work, but it would be better.

motoman_driver/src/io_relay.cpp
```diff
@@ -0,0 +1,127 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ * Copyright (c) 2013, Southwest Research Institute
```
gavanderhoorn: Update copyright here as well.

And as this includes code moved from the original location, it'd be nice to add Delft Universities copyright as well

motoman_driver/src/io_relay.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+using industrial::shared_types::shared_int;
+
+namespace motoman
+{
+namespace io_relay
+{
+
+// TODO: is this needed?
+#define ROS_ERROR_RETURN(rtn,...) do {ROS_ERROR(__VA_ARGS__); return(rtn);} while(0)
```
gavanderhoorn:
```suggestion
```

This is not used anywhere in this file, so remove.

motoman_driver/src/io_relay.cpp
```diff
@@ -0,0 +1,127 @@
...
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "motoman_driver/io_relay.h"
+#include <string>
+#include <ros/ros.h>
+
+using industrial::shared_types::shared_int;
```
gavanderhoorn: Could you move this as close as possible to the scope in which it is actually used?

motoman_driver/src/io_relay.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+namespace motoman
+{
+namespace io_relay
+{
+
+// TODO: is this needed?
+#define ROS_ERROR_RETURN(rtn,...) do {ROS_ERROR(__VA_ARGS__); return(rtn);} while(0)
+
+bool MotomanIORelay::init(std::string default_ip, int default_port)
```
gavanderhoorn: See earlier comment on removing `default_ip`).

motoman_driver/src/io_relay.cpp
```diff
@@ -0,0 +1,127 @@
...
+{
+  ROS_INFO("MotomanIORelay: init");
```
gavanderhoorn: Good for debugging/reviewing, but please change to `DEBUG`.

Preferably even `ROS_DEBUG_NAMED`.

motoman_driver/src/io_relay.cpp
```diff
@@ -0,0 +1,127 @@
...
+    return false;
+  }
+
+  char* ip_addr = strdup(ip.c_str());  // connection.init() requires "char*", not "const char*"
+  ROS_INFO("I/O relay connecting to IP address: '%s:%d'", ip_addr, port);
+  default_tcp_connection_.init(ip_addr, port);
+  free(ip_addr);
+  default_tcp_connection_.makeConnect();
+
+  io_ctrl_.init(&default_tcp_connection_);
```
gavanderhoorn: Can `init(..)` fail? If so: should we check its return value?

motoman_driver/src/io_relay.cpp
```diff
@@ -0,0 +1,127 @@
...
+  shared_int io_val= -1;
+
+  // send message and release mutex as soon as possible
+  this->mutex_.lock();
+  bool result = io_ctrl_.readSingleIO(req.address, io_val);
+  this->mutex_.unlock();
+
+  if (!result)
+  {
+    ROS_ERROR("Reading IO element %d failed", req.address);
```
gavanderhoorn: Please use `ROS_ERROR_NAMED` here as well.

motoman_driver/src/io_relay.cpp
```diff
@@ -0,0 +1,127 @@
...
+    return false;
+  }
+  res.value = io_val;
+  ROS_INFO("Element %d value: %d", req.address, io_val);
```
gavanderhoorn: Suggestion: make these `DEBUG` (same in `writeSingleIoCB(..)`).

motoman_driver/src/io_relay_node.cpp
```diff
@@ -0,0 +1,50 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "motoman_driver/io_relay.h"
+
+using motoman::io_relay::MotomanIORelay;
+
+int main(int argc, char** argv)
+{
+  const int io_port = 50242;
```
gavanderhoorn:
```suggestion
  const int default_io_port = 50242;
```

motoman_driver/src/io_relay_node.cpp
```diff
@@ -0,0 +1,50 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ * Copyright (c) 2013, Southwest Research Institute
```
gavanderhoorn: Copyright again.

gavanderhoorn (changes requested):
Overall this looks OK, some comments.

Would you want to include some changes to address #360 and #362, or would you prefer to do that in separate PRs?

motoman_driver/launch/io_relay.launch
gavanderhoorn: Hm, on second thought: perhaps changing this now (as in: this late) would not be a good idea.

@marip8: thoughts?

marip8:
> Would you want to include some changes to address #360 and #362, or would you prefer to do that in separate PRs?

I would suggest separate PRs just so we don't introduce too many changes for review at once

motoman_driver/launch/io_relay.launch
marip8: For consistency, we should probably leave this the same as the trajectory streamer. I would prefer privately namespaced parameters, though. Maybe we could make a separate PR that changes both the trajectory streamer and IO relay nodes together to pull the required parameters from a private namespace. It would touch a large number of files because we would need to modify the launch files for all robots in this and the experimental repository to load the parameters correctly

motoman_driver/src/io_relay.cpp
steviedale: @gavanderhoorn Should I convert all DEBUG, ERROR, and INFO macros in this file to named versions? Or just the ones that you specified?

Also, should the name argument be the name of the class, `MotomanIORelay`?

motoman_driver/src/io_relay_node.cpp
steviedale: Addressed in 8c1757d

motoman_driver/src/io_relay.cpp
steviedale: Addressed in 8c1757d

motoman_driver/include/motoman_driver/io_relay.h
steviedale: Addressed in 8c1757d

motoman_driver/include/motoman_driver/io_relay.h
steviedale: Addressed in bf9fb1c

motoman_driver/include/motoman_driver/io_relay.h
steviedale: Addressed in e6cab50

motoman_driver/src/io_relay.cpp
steviedale: Addressed in e6cab50

motoman_driver/src/io_relay_node.cpp
steviedale: Addressed in e6cab50

motoman_driver/launch/io_relay.launch
steviedale: Addressed in f0834eb

motoman_driver/src/io_relay.cpp
steviedale: Addressed in da21862

motoman_driver/src/io_relay.cpp
steviedale: Addressed in 8729f66

motoman_driver/src/io_relay.cpp
gavanderhoorn: Would be nice to use `NAMED` everywhere, yes.

> Also, should the name argument be the name of the class, `MotomanIORelay`?

I wouldn't use the name of the class. We already know the node, the file and the line nr (via [this](http://wiki.ros.org/ROS/EnvironmentVariables#Console_Output_Formatting)), and the class doesn't add to much imo.

Suggestion: group log lines based on 'theme'. So in this case it would be `read`, earlier logs might be `init`, etc.

motoman_driver/src/io_relay.cpp
steviedale: It looks like the `MotomanIOCtrl::init` function is currently set to always return true. Should I change the function type to void? Or should I leave it how it is and add in a check here for the bool (in case we decide to add error checking to the `init` function later)?

https://github.com/steviedale/motoman/blob/8729f66b5fccb1784e7c7c62706ffb7a978f0f32/motoman_driver/src/io_ctrl.cpp#L62-L66

motoman_driver/src/io_relay.cpp
```diff
@@ -0,0 +1,124 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ * Copyright (c) 2020, Southwest Research Institute
```
gavanderhoorn, line 5:
```suggestion
 * Copyright (c) 2016, Delft Robotics Institute
 * Copyright (c) 2020, Southwest Research Institute
```

seeing as this code is largely copied, would be nice to mention them (they pay my salary, so :))

motoman_driver/src/io_relay.cpp
gavanderhoorn: As the method returns a `bool`, that's what we should check. Fact it cannot be anything but `true` right now doesn't change that.

motoman_driver/src/io_relay.cpp
steviedale: Hahaha fair point. I need to watch out for the copyrights (I get a little bit copy and paste happy)

Addressed in 001d34e and 4ec0248

gavanderhoorn: Thanks.

Do you not like the *Commit suggestion* feature?

motoman_driver/src/io_relay.cpp
steviedale: Addressed in a6abe75

Sounds good. Let me know if any of these names should be changed.

motoman_driver/src/io_relay.cpp
steviedale: Addressed in a6abe75

motoman_driver/src/io_relay.cpp
steviedale: Addressed in a6abe75

motoman_driver/src/io_relay.cpp
steviedale: Addressed in abed8fd

Sounds good. I also added a similar check to the `TcpClient.init` call.

gavanderhoorn:
Nice. Thanks for iterating.

I'll test this tomorrow and if everything checks out we can merge it.

steviedale:
> Nice. Thanks for iterating.
> 
> I'll test this tomorrow and if everything checks out we can merge it.

No problem, thank you for the feedback! I'll start working on the pull requests for #360 and #362

motoman_driver/src/io_relay.cpp
```diff
@@ -0,0 +1,134 @@
...
+  ros::param::get("robot_ip_address", ip);
+  // override port with ROS param, if available
+  ros::param::param<int>("~port", port, default_port);
+
+  // check for valid parameter values
+  if (ip.empty())
+  {
+    ROS_ERROR_NAMED("init", "No valid robot IP address found.  Please set ROS 'robot_ip_address' param");
+    return false;
+  }
```
marip8:
```suggestion
  // override port with ROS param, if available
  const std::string port_parameter = "port";
  if(!ros::param::param<int>(port_parameter, port, default_port))
  {
    ROS_WARN_STREAM_NAMED("init", "Failed to get '" << port_parameter << "' parameter; setting to default value");
  }
  
  const std::string robot_ip_parameter = "robot_ip";
  if(!ros::param::get(robot_ip_parameter, ip) || ip.empty())
  {
    ROS_ERROR_NAMED_STREAM("init", "Invalid IP address; please set the " << robot_ip_parameter << "' parameter");
    return false;
  }
```

marip8: It would be nice to know when parameters aren't found or a default value is used.

motoman_driver/launch/io_relay.launch
```diff
@@ -0,0 +1,12 @@
+<!-- 
+  Wrapper launch file for I/O relay node.
+-->
+<launch>
+  <!-- IP of robot (or PC running simulation) -->
+  <arg name="robot_ip" doc="IP of controller" />
+
+  <!-- put them on the parameter server -->
+  <param name="robot_ip_address" type="str" value="$(arg robot_ip)" />
```
marip8, line 12: I would change the `robot_ip_address` parameter to `robot_ip` to be consistent with the argument

motoman_driver/include/motoman_driver/io_relay.h
```diff
@@ -0,0 +1,81 @@
...
+class MotomanIORelay
+{
+public:
+  /**
+   * \brief Class initializer
+   *
+   * \param default_port
+   * \return true on success, false otherwise
+   */
+  bool init(int default_port);
```
marip8, line 60: Is there really a need for a separate `init` function? I would prefer to load the `port` and `robot_ip` parameters in the node executable and pass their values to the constructor of this class instead of having a separate `init` function.

motoman_driver/launch/io_relay.launch
gavanderhoorn: This would be a similar change to the private vs global parameters. Probably better to do this in a follow-up PR.

motoman_driver/include/motoman_driver/io_relay.h
gavanderhoorn: Need: no. But this is almost a straight refactor taking the code out of the trajectory relay.

We can handle this in a follow-up PR.

gavanderhoorn:
I've added some commits. Mostly cosmetics and a few logging changes.

gavanderhoorn:
@steviedale: I've had to fix up b3a2658 and forced pushed the branch there. Not sure whether you'll have to `reset` your local copy (probably not, as you probably do not have those commits yet).

motoman_driver/src/io_relay.cpp
gavanderhoorn: You're right, that is nicer. Implemented in 08cbb8e:

```
[ WARN] [1603449484.869552736] [/my_ns/io_relay] [ros.motoman_driver.io.init]: Failed to get '~port' parameter: using default (50242)
```

gavanderhoorn:
And a few more commits.

Base `roslaunch` testing and make sure the new node gets installed.

And we want to also support FS100, so I've added the byte-swapped version of the IO relay as well.

gavanderhoorn:
I've tested this on my YRC1000 and it (still) works (as in: it already worked before, integrated into the trajectory relay, and still does).

Could you check my additions @steviedale and let me know whether you're OK with them?

steviedale:
> Could you check my additions @steviedale and let me know whether you're OK with them?

@gavanderhoorn: Your changes look good to me.

gavanderhoorn (approved):
Looks good now.

Thanks @steviedale :+1: 

We'll iterate further on this in follow-up PRs.

gavanderhoorn:
With the note that this PR will "break" IO support for those users on *old* MotoROS versions, we can merge this.

Note: before #355 we had no support for the IO svcs available in MotoROS, so we're breaking those users which were using this in the past 14 days only. Seeing as we haven't (yet) announced the added support, I deem this breakage acceptable.

This is breaking, as we're moving the service servers from the trajectory relay to the stand-alone IO relay, and the latter will now connect to the stand-alone IO server in MotoROS, instead of abusing the trajectory task for IO services. So only those users with a MotoROS old enough to *only* have the IO services in the trajectory task *and* having used #355 the past few weeks would notice this breakage.

gavanderhoorn:
Thanks again @steviedale and @marip8.

PR #366, opened 2020-10-22
Add success and message fields to IO services

steviedale:
This pull request addresses issue ros-industrial#362, adding success and message fields to the Motoman IO services

### DIFF
--- a/motoman_driver/src/io_relay.cpp
+++ b/motoman_driver/src/io_relay.cpp
@@ -114,11 +114,14 @@ bool MotomanIORelay::readSingleIoCB(
 
   if (!result)
   {
-    std::stringstream message;
-    message << "Reading IO element " << req.address << " failed";
     res.success = false;
+
+    // provide caller with failure indication (not very informative yet)
+    std::stringstream message;
+    message << "Reading IO element " << req.address << " failed.";
     res.message = message.str();
     ROS_ERROR_STREAM_NAMED("io.read", res.message);
+
     return true;
   }
 
@@ -145,11 +148,14 @@ bool MotomanIORelay::writeSingleIoCB(
 
   if (!result)
   {
+    res.success = false;
+
+    // provide caller with failure indication (not very informative yet)
     std::stringstream message;
     message << "Writing to IO element " << req.address << " failed";
-    res.success = false;
     res.message = message.str();
     ROS_ERROR_STREAM_NAMED("io.write", res.message);
+
     return true;
   }
 
--- a/motoman_driver/src/io_relay.cpp
+++ b/motoman_driver/src/io_relay.cpp
@@ -122,12 +122,11 @@ bool MotomanIORelay::readSingleIoCB(
     return true;
   }
 
-  std::stringstream message;
-  message << "Element " << req.address << " value: " << io_val;
+  ROS_DEBUG_STREAM_NAMED("io.read", "Address " << req.address << ", value: " << io_val);
+
+  // no failure, so no need for an additional message
   res.value = io_val;
   res.success = true;
-  res.message = message.str();
-  ROS_DEBUG_NAMED("io.read", message.str().c_str());
   return true;
 }
 
@@ -154,11 +153,10 @@ bool MotomanIORelay::writeSingleIoCB(
     return true;
   }
 
-  std::stringstream message;
-  message << "Element " << req.address << " set to: " << req.value;
+  ROS_DEBUG_STREAM_NAMED("io.write", "Element " << req.address << " set to: " << req.value);
+
+  // no failure, so no need for an additional message
   res.success = true;
-  res.message = message.str();
-  ROS_DEBUG_NAMED("io.write", message.str().c_str());
   return true;
 }
 
--- a/motoman_driver/src/io_relay.cpp
+++ b/motoman_driver/src/io_relay.cpp
@@ -118,7 +118,7 @@ bool MotomanIORelay::readSingleIoCB(
     message << "Reading IO element " << req.address << " failed";
     res.success = false;
     res.message = message.str();
-    ROS_ERROR_NAMED("io.read", message.str().c_str());
+    ROS_ERROR_STREAM_NAMED("io.read", message.str());
     return true;
   }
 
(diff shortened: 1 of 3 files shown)


steviedale:
@gavanderhoorn I created this branch off of the `steviedale:update/separate_io_from_joint_streamer` branch (from this [PR](https://github.com/ros-industrial/motoman/pull/365)). I figured this would avoid merge conflicts, but if this is against best practice I can rebase off of the `ros-industrial:kinetic-devel` branch instead.

gavanderhoorn:
You'll probably want to `rebase` this after we merge #365.

gavanderhoorn:
Could you clarify why you added c80c351? `std::to_string(..)` looks like it makes sense where you used it.

marip8:
A couple general comments:

- I don't find it useful to print error messages returned by services to the console. If you call a service from the command line this message will be printed automatically when the service completes, and if a caller of the service wants to log the error then they can do so in the calling code. To me, this additional logging just adds more code and clutters the terminal output unnecessarily. I would recommend removing the ROS logging from the service callbacks
- I prefer to make make service callback functions always return true and return the logical success in the service response instead. When a service returns false, it's ambiguous whether the service was unreachable (which also causes `ros::service::call(...)` to return false) or if there was a logical failure.

gavanderhoorn:
>  * I don't find it useful to print error messages returned by services to the console. If you call a service from the command line this message will be printed automatically when the service completes, and if a caller of the service wants to log the error then they can do so in the calling code. To me, this additional logging just adds more code and clutters the terminal output unnecessarily. I would recommend removing the ROS logging from the service callbacks

~Are you thinking of ROS services?~

~Afaict, @steviedale kept the logging which was already there in the callbacks of the Simple Message service.~

~It seems like it would be informative to keep those, as it allows to quickly see whether at least the Simple Message part of the communication succeeded or not.~

I misread the diff.

I'm not against printing the message to the console, as it's a nice way first step in debugging or diagnosing a "why does it not work" situation. It's for humans only though, so @marip8 is right in that callers could do this themselves.

> * I prefer to make make service callback functions always return true and return the logical success in the service response instead. When a service returns false, it's ambiguous whether the service was unreachable (which also causes `ros::service::call(...)` to return false) or if there was a logical failure.

Assuming you are referring to ROS services here: there are very specific constraints on how services should behave: [roscpp/Overview/Services: Callback Signature](http://wiki.ros.org/roscpp/Overview/Services#Callback_Signature).

`false` should be returned when the server failed to process the service request.

So we should probably be returning `true` here always, as long as were able to forward the request to the MotoROS side. Then communicate success of the read/write operation using the `result` and `message` fields.

That was the rationale for adding the `result` and `message` fields (also to `std_srvs/Trigger`), as the `bool` return from the callback should not be used for this.

steviedale:
> Could you clarify why you added [c80c351](https://github.com/ros-industrial/motoman/commit/c80c351138beaff915f7b1b555d899aafcae99eb)? `std::to_string(..)` looks like it makes sense where you used it.

I changed this because the kinetic CI was failing (the melodic passes though). It looks like maybe the version of C++ being used for kinetic does not have std::to_string. I believe this function was added in C++11. I'm not sure what version the kinetic CI is set up for or if it can be easily changed, but I figured an easy solution would be to use a more backwards compatible method for int to string conversion.

steviedale:
Apologies for the force commit and any confusion that might bring. I messed up a rebase and needed to fix it. The changes are the same as before with the addition of always returning `true` for io service calls.

gavanderhoorn:
@steviedale: with #365 merged I believe this PR contains some duplicated commits?

steviedale:
@gavanderhoorn I just fixed the commit history. Are there any changes that still need to be addressed for this pull request?

motoman_driver/src/io_relay.cpp
```diff
@@ -113,12 +114,20 @@ bool MotomanIORelay::readSingleIoCB(
 
   if (!result)
   {
-    ROS_ERROR_NAMED("io.read", "Reading IO element %d failed", req.address);
-    return false;
+    std::stringstream message;
+    message << "Reading IO element " << req.address << " failed";
```
marip8: Please update the messages here with a single quote or parentheses around the streamed variables like `req.address`. In the case that a user specifies a blank address, this error message would be somewhat confusing and unhelpful: `Reading IO element  failed` vs `Reading IO element () failed`

gavanderhoorn: Isn't `address` an integer?

marip8:
```suggestion
    message << "Reading IO element (" << req.address << ") failed";
```

gavanderhoorn: @marip8: if `address` is an integer, can it ever be empty?

marip8: Good point; I would think it probably couldn't be empty (opposed to strings that are blank by default). This change is probably superfluous then

gavanderhoorn: I fully agree with you in case this was a `string` btw. Nothing more perplexing than seeing weird error messages.

gavanderhoorn:
@steviedale: I've added some commits. Some cosmetic, but I also wanted to avoid having to pass `char*` to `ROS_ERROR_NAMED(..)` (as we do actually have a `std::string`), and we don't need to add a "success message" to the service response. `success` is already `true`, so it's clear things worked.

Perhaps the last change also connects to what @marip8 wrote in his earlier comment.

The `message` is there to help the caller (ie: human in this case still) determine what happened. But in case of success, we already know, so no need for the additional message.

gavanderhoorn:
@steviedale: could you let me know whether you are OK with my additions?

I've tested them with our YRC1000, and on invalid writes I now get:

```
$ rosservice call /write_single_io "{ address: 10010, value: 2 }"
success: False
message: "Writing to IO element 10010 failed"
```

instead of `service returned an error:` (and then nothing, due to use returning `false`).

steviedale:
@gavanderhoorn I just looked over your changes, looks good to me.

gavanderhoorn (approved):
Thanks @steviedale for the contribution :+1:

Issue #358, opened 2020-10-06
io server can return uninitialised memory in case of failure


gavanderhoorn:
As described by @ted-miller in https://github.com/ros-industrial/motoman/pull/355#issuecomment-704349823, `Ros_IoServer_ReadIOBit(..)` may be returning uninitialised memory as part of a `ROS_MSG_MOTO_READ_IO_BIT_REPLY`:

https://github.com/ros-industrial/motoman/blob/47ef63fe1f30cad1de00ead44a2479f027ecc8b9/motoman_driver/MotoPlus/IoServer.c#L335

Those variables should be initialised, and even in case of failure should not return garbage data.

gavanderhoorn:
With #369 merged, I'm going to close this.

PR #369, opened 2020-11-10
MotoROS v1.9.2

ted-miller:
- Add validation to the input data for reading/writing I/O.
- Add new result codes to the I/O reply message.
- Add support for reading/writing M-Registers.


gavanderhoorn:
Apologies for leaving this sitting so long @ted-miller. I'll review this Friday.

@steviedale: could you also test this on your controller/robot?

Should make the work needed for #360 much simpler.

gavanderhoorn:
This also seems to address #358.

motoman_driver/MotoPlus/IoServer.c
```diff
@@ -324,26 +328,30 @@ int Ros_IoServer_ReadIOBit(SimpleMsg* receiveMsg, SimpleMsg* replyMsg)
...
-	apiRet = mpReadIO(&ioReadInfo, &ioValue, 1);
+	if (Ros_IoServer_IsValidReadAddress(receiveMsg->body.readIOBit.ioAddress, IO_ACCESS_BIT))
+	{
+		ioReadInfo.ulAddr = receiveMsg->body.readIOBit.ioAddress;
+		apiRet = mpReadIO(&ioReadInfo, &ioValue, 1);
 
-	if (apiRet == OK)
-		resultCode = ROS_REPLY_SUCCESS;
+		replyMsg->body.readIOBitReply.value = ioValue;
+		replyMsg->body.readIOBitReply.resultCode = (apiRet == OK) ? IO_RESULT_OK : IO_RESULT_READ_API_ERROR;
```
gavanderhoorn, line 337: This seems to now return `0` when `apiRet == OK` in `resultCode`. It used to return `ROS_REPLY_SUCCESS` (which `== 1`).

I'm not necessarily against this change, just wanted to note this somewhere.

motoman_driver/MotoPlus/IoServer.c
```diff
@@ -324,26 +328,30 @@ int Ros_IoServer_ReadIOBit(SimpleMsg* receiveMsg, SimpleMsg* replyMsg)
...
+		replyMsg->header.replyType = (apiRet == OK) ? ROS_REPLY_SUCCESS : ROS_REPLY_FAILURE;
```
gavanderhoorn, line 338: Technically, `reply_code` should only encode the success or failure of the reception and processing of the message, not of the execution of the service.

I know this isn't documented very explicitly though. See [this related discussion](https://github.com/ros-industrial/rep/pull/20#discussion_r173632236) with @shaun-edwards.

To summarise:

 - `replyType` should tell clients whether the service request itself was received correctly, and whether the server supports the request.
 - the `resultCode` should then be used to determine whether the action associated with the request was successfully executed.

This is a direct analog of the way ROS services behave.

However, I'm not entirely sure whether we've adhered to this with the other implementations of service servers in MotoROS. Would you know @ted-miller ?

motoman_driver/MotoPlus/IoServer.c
```diff
@@ -357,34 +365,40 @@ int Ros_IoServer_ReadIOGroup(SimpleMsg* receiveMsg, SimpleMsg* replyMsg)
...
 
-	for (i = 0; i < 8; i += 1)
+	if (Ros_IoServer_IsValidReadAddress(receiveMsg->body.readIOGroup.ioAddress, IO_ACCESS_GROUP))
 	{
-		ioReadInfo[i].ulAddr = (receiveMsg->body.readIOGroup.ioAddress * 10) + i;
-	}
-	apiRet = mpReadIO(ioReadInfo, ioValue, 8);
+		for (i = 0; i < 8; i += 1)
+		{
+			ioReadInfo[i].ulAddr = (receiveMsg->body.readIOGroup.ioAddress * 10) + i;
```
gavanderhoorn, line 372: I may have asked this before, but why the `* 10`?

motoman_driver/MotoPlus/IoServer.c
```diff
@@ -427,19 +453,184 @@ int Ros_IoServer_WriteIOGroup(SimpleMsg* receiveMsg, SimpleMsg* replyMsg)
...
+
+	// set prefix: length of message excluding the prefix
+	replyMsg->prefix.length = sizeof(SmHeader) + sizeof(SmBodyMotoReadIOMRegisterReply);
+
+	// set header information of the reply
+	replyMsg->header.msgType = ROS_MSG_MOTO_READ_MREGISTER_REPLY;
+	replyMsg->header.commType = ROS_COMM_SERVICE_REPLY;
+
+	if (receiveMsg->body.readRegister.registerNumber < 1000000)
+		receiveMsg->body.readRegister.registerNumber += 1000000;
```
gavanderhoorn, line 501: Would it perhaps be better to check this and fail if `1e6` was added by the client?

If I understand the comments in `SimpleMessage.h` correctly, clients should specify the register index directly (ie: 1, 2, 3, etc), instead of adding `1e6` themselves.

For consistency, it may be better to reject `registerNumber > whatever_the_normal_max_is`.

motoman_driver/MotoPlus/IoServer.c
```diff
@@ -324,26 +328,30 @@ int Ros_IoServer_ReadIOBit(SimpleMsg* receiveMsg, SimpleMsg* replyMsg)
 	replyMsg->header.msgType = ROS_MSG_MOTO_READ_IO_BIT_REPLY;
 	replyMsg->header.commType = ROS_COMM_SERVICE_REPLY;
 	
-	ioReadInfo.ulAddr = receiveMsg->body.readIOBit.ioAddress;
-	apiRet = mpReadIO(&ioReadInfo, &ioValue, 1);
+	if (Ros_IoServer_IsValidReadAddress(receiveMsg->body.readIOBit.ioAddress, IO_ACCESS_BIT))
+	{
+		ioReadInfo.ulAddr = receiveMsg->body.readIOBit.ioAddress;
+		apiRet = mpReadIO(&ioReadInfo, &ioValue, 1);
```
gavanderhoorn: Is `1` the number of addresses to read? If so: could we avoid the magic nr and use a named constant?

motoman_driver/MotoPlus/IoServer.h
```diff
@@ -34,10 +34,240 @@
...
+	IO_ACCESS_GROUP,
+	IO_ACCESS_REGISTER
+} IoAccessSize;
+
+extern BOOL Ros_IoServer_IsValidReadAddress(UINT32 address, IoAccessSize size);
+extern BOOL Ros_IoServer_IsValidWriteAddress(UINT32 address, IoAccessSize size);
+extern BOOL Ros_IoServer_IsValidWriteValue(UINT32 value, IoAccessSize size);
+
+//**********************************************************
+#if DX100
```
gavanderhoorn, line 61: These constants seem like they overlap in many places.

Would it be worth it to reuse them if they have the same values? Or would you want to keep the duplicated ones?

motoman_driver/MotoPlus/SimpleMessage.h
```diff
@@ -69,6 +69,10 @@ typedef enum
 	ROS_MSG_MOTO_WRITE_IO_GROUP = 2009,
 	ROS_MSG_MOTO_WRITE_IO_GROUP_REPLY = 2010,
 	ROS_MSG_MOTO_IOCTRL_REPLY = 2011,
+	ROS_MSG_MOTO_READ_MREGISTER = 2012,
+	ROS_MSG_MOTO_READ_MREGISTER_REPLY = 2013,
+	ROS_MSG_MOTO_WRITE_MREGISTER = 2014,
+	ROS_MSG_MOTO_WRITE_MREGISTER_REPLY = 2015,
```
gavanderhoorn: This will require a corresponding update to [REP-I4: Assigned Message Identifiers for the Simple Message Protocol](https://github.com/ros-industrial/rep/blob/bbd88b9b1dc9b61be5178c8b3a93b9bec488702a/rep-I0004.rst), specifically the *Vendor Specific Ranges: Motoman* section.

gavanderhoorn: Additionally: there should be no need for separate IDs for requests and replies. Requests and replies share the same ID, so `2012` would be read and `2013` would be write of M registers.

motoman_driver/MotoPlus/SimpleMessage.h
```diff
@@ -285,7 +299,7 @@ typedef struct _SmBodyMotoReadIOBit SmBodyMotoReadIOBit;
 struct _SmBodyMotoReadIOBitReply
 {
 	UINT32 value;
-	UINT32 resultCode;
+	IoResultCodes resultCode;
```
gavanderhoorn, line 300: Is this always correct? I'm not completely up-to-speed on how VxWorks maps enums to integer types, but afaik the standard does not necessarily dictate that 4-byte integers must be used for all enums.

motoman_driver/MotoPlus/SimpleMessage.h
```diff
@@ -298,33 +312,33 @@ typedef struct _SmBodyMotoWriteIOBit SmBodyMotoWriteIOBit;
...
 {
-	UINT32 resultCode;
+	IoResultCodes resultCode;
 } __attribute__((__packed__));
 typedef struct _SmBodyMotoWriteIOBitReply SmBodyMotoWriteIOBitReply;
 
 struct _SmBodyMotoReadIOGroup
 {
-	UINT32 ioAddress;
+	UINT32 ioAddress; //Group address. Example: '1001' will read OG#1 (10010 - 10017)
```
gavanderhoorn:
```suggestion
	UINT32 ioAddress; //Group address. Example: '1001' will read OG#1 (IOs / bits 10010 to 10017)
```
The `-` confused me at first, I thought there was a subtraction there.

gavanderhoorn: And the same below.

gavanderhoorn (changes requested):
I've tested this on our YRC1000 and it works. Especially with #371 this results in informative error messages when using the IO interface incorrectly.

It would have been nice to have split this PR into (at least) two commits though: the addition of M-register read/write seems to be fairly stand-alone functionality.

gavanderhoorn:
Thanks for the PR @ted-miller :+1: 

I've left some in-line comments.

motoman_driver/MotoPlus/IoServer.c
ted-miller: > However, I'm not entirely sure whether we've adhered to this with the other implementations of service servers in MotoROS. Would you know @ted-miller ?

This has not been adhered to throughout MotoROS. Generally speaking, if anything fails, the replyType will not indicate 'success'. Here's an example in the motion server:
https://github.com/ros-industrial/motoman/blob/4b0a030de3917aa13ab1407f2dcf131d4f0ffd27/motoman_driver/MotoPlus/MotionServer.c#L608-L636

motoman_driver/MotoPlus/IoServer.c
ted-miller: When reading a whole 'group', it is expecting the group-number, which is the first four digits of an I/O address.
For example, for the signal 12345, it is Group 1234, Bit 5.

motoman_driver/MotoPlus/IoServer.c
ted-miller: > If I understand the comments in SimpleMessage.h correctly, clients should specify the register index directly (ie: 1, 2, 3, etc), instead of adding 1e6 themselves.

Yes. But, I put this catch in there in case someone doesn't read the comment. 
I think it's better to allow for more flexible input format, rather than return an error code and force someone to waste time debugging just because they didn't read a comment in a header file. This extra check eliminates confusion and headache in the future.

motoman_driver/MotoPlus/IoServer.c
ted-miller: But magic is fun...

motoman_driver/MotoPlus/IoServer.h
ted-miller: I don't understand your comment. Can you elaborate?

If you're referring the usage of the controller-model constant, this is something that is defined in the build system for the MotoPlus SDK.

motoman_driver/MotoPlus/IoServer.c
gavanderhoorn: Hm. That's not nice, but also not something we can change at this point I believe.

motoman_driver/MotoPlus/SimpleMessage.h
ted-miller: I'm fine with removing the 'reply' message ID's. However, this is the method that is also used for motion-control messages.
https://github.com/ros-industrial/motoman/blob/4b0a030de3917aa13ab1407f2dcf131d4f0ffd27/motoman_driver/MotoPlus/SimpleMessage.h#L60-L61

motoman_driver/MotoPlus/IoServer.c
gavanderhoorn: > I think it's better to allow for more flexible input format, rather than return an error code and force someone to waste time debugging just because they didn't read a comment in a header file. 

that's not the only option we have: we can create an appropriate result_code and return it to clients.

motoman_driver/MotoPlus/IoServer.c
gavanderhoorn: Certainly :)

motoman_driver/MotoPlus/SimpleMessage.h
ted-miller: I'm not sure if that is dictated by the build system or the OS. But, I can confirm with certainty that an enum is a 4 byte int in this environment.

motoman_driver/MotoPlus/IoServer.h
gavanderhoorn: Example: `IFPANELMIN` and `IFPANELMAX` seem to be the same across all supported controller models.

Instead of defining the values 5 times, we could also do it once, and only provide `#define`s for the values which are different.

motoman_driver/MotoPlus/SimpleMessage.h
gavanderhoorn: > However, this is the method that is also used for motion-control messages.

I know, but that was also unnecessary.

I'm not sure why/how that got through the reviews at the time.

motoman_driver/MotoPlus/IoServer.h
ted-miller: Yeah, I originally tried to determine what all the differences were. But, it started to get confusing and difficult to read the code due to all the similarities/differences between controllers.

I think it's easier to read if they are all grouped nicely in a block in numerical order. It will also be easier to update in the future, if this code is around for the next controller generation.

motoman_driver/MotoPlus/SimpleMessage.h
ted-miller: I guess my only concern is breaking existing functionality.

There are users who are currently using the I/O interface, today. I assume that their code will be looking for the "_REPLY" message ID. If we remove these messages, will it break existing installations? (assuming they upgrade their MotoROS at some point)

https://github.com/ros-industrial/motoman/blob/4b0a030de3917aa13ab1407f2dcf131d4f0ffd27/motoman_driver/MotoPlus/SimpleMessage.h#L63-L71

Thoughts?

motoman_driver/MotoPlus/IoServer.h
gavanderhoorn: Ok. Makes sense.

motoman_driver/MotoPlus/SimpleMessage.h
gavanderhoorn: > I assume that their code will be looking for the "_REPLY" message ID.

could you elaborate how? The numbers you refer to are numeric IDs of message structures. I would expect code to work with the numbers, but not necessarily (de)serialise to different classes/structures.

I'm not suggesting we remove the ones which are already there. It's unfortunate we have those, but removing them may break code outside of this repository.

I'm suggesting to not introduce *more* unnecessary ID allocations.

motoman_driver/MotoPlus/SimpleMessage.h
gavanderhoorn: Hm. I don't personally like to assume things like this (as they tend to change and break things while I'm not looking), but if you're comfortable with this, then we can keep it.

motoman_driver/MotoPlus/IoServer.c
ted-miller: This has been updated

motoman_driver/MotoPlus/SimpleMessage.h
ted-miller: This has been updated. Please note that I also changed the value for ROS_MSG_MOTO_WRITE_MREGISTER. So, that will need to be updated on the motomand-driver side too.

motoman_driver/MotoPlus/SimpleMessage.h
ted-miller: This has been updated.

motoman_driver/MotoPlus/SimpleMessage.h
gavanderhoorn: Support for those things need to be added still, so there is nothing to update yet.

motoman_driver/MotoPlus/IoServer.c
```diff
@@ -427,19 +453,184 @@ int Ros_IoServer_WriteIOGroup(SimpleMsg* receiveMsg, SimpleMsg* replyMsg)
...
+	bValueOk = Ros_IoServer_IsValidWriteValue(receiveMsg->body.writeIOGroup.ioValue, IO_ACCESS_GROUP);
+
+	if (bAddressOk && bValueOk)
+	{
+		for (i = 0; i < 8; i += 1)
+		{
+			ioWriteData[i].ulAddr = (receiveMsg->body.writeIOGroup.ioAddress * 10) + i;
+			ioWriteData[i].ulValue = (receiveMsg->body.writeIOGroup.ioValue & (1 << i)) >> i;
+		}
+		apiRet = mpWriteIO(ioWriteData, 8);
```
gavanderhoorn: Should this be updated as well to use the constant(s)?

ted-miller: You're right. I just missed that.

motoman_driver/MotoPlus/IoServer.c
gavanderhoorn: Would returning an error result code be an option @ted-miller? Seems anything out of the range [0, 1000)` would be illegal.

ted-miller: I don't see any reason to return an error. If the value is between 0 - 999 or 1000000 - 1000999, then we know what they're trying to do. We might as well honor the request message. There's no benefit to making it more difficult for the user.

gavanderhoorn: Reason I suggest returning an error is because it's very valuable to have interfaces which do one thing well, without possibility for ambiguity or allowing unneeded variation.

Allowing both `[0, 999]` and `[1000000, 1000999]` is such unneeded variation. Only by reading documentation would users know which ranges are accepted, and it would only be the documentation here iiuc, as the MotoPlus documentation actually only allows `[1000000, 1000999]`.

Seeing as that's a likely source of information for users / customers of Yaskawa robots, it would seem good to actually only accept `[1000000, 1000999]` and nothing else.

This would make it consistent with available documentation and the the rest of of the IO services we expose here.

gavanderhoorn: But I don't feel strongly enough about it to hold up this PR, so if you'd like this merged as-is now, we can do that.

gavanderhoorn:
@steviedale @marip8: would you have a chance to test this on your controller before the end of this week?

It's OK of course if you can't, just let me know and I'll try to test it myself.

steviedale:
@gavanderhoorn Sorry for the delayed response regarding the testing. I won't be in the office until Monday. I can test it first thing on Monday if that isn't too late?

motoman_driver/MotoPlus/IoServer.c
ted-miller: I've had too many occurrences of supporting a user over the course of multiple days, only to find that the root cause was a simple ID or value being misinterpreted. It rightfully upsets the user, and wastes both our time.

Anything that makes an interface more accepting so that it "just works" is a good thing. There is no downside to making something easier for different users. (Especially since this interface is being separated from the motoman-driver so that it can be used independently.)

gavanderhoorn: Ok. Let's agree to disagree. Personally I find clear error messages and clearly defined interfaces a good thing, especially if those interfaces adhere to what is documented (in documentation available elsewhere), but I can see how what you describe can be valuable.

gavanderhoorn:
I've tested the YRC1000 binary on our controller today.

Everything seems to be in order.

As such, I'm merging this.

CI is not cooperating due to some changes at Travis. I'll use admin privileges to ignore that and merge the PR.

gavanderhoorn:
Note: the code for reading/writing M-registers has not been tested by me.

That will have to wait until we have client code in `motoman_driver`.

I do not want to block this PR on that becoming available.

gavanderhoorn (approved):

gavanderhoorn:
Actually, I'll wait for @marip8's approval of #371.

gavanderhoorn:
Thanks @ted-miller for the changes.

PR #371, opened 2020-11-20
Forward MotoROS (1.9.2+) IO service results to ROS clients

### LINKED ISSUE #367: Extend MotoROS to return informative errors from IO read/write services
Similar to #360, the `io_relay` could check for invalid values being passed with write requests.

For single digital IOs, anything but `0` or `1` is illegal.

For group IO, there is a specific maximum value that could be written.

We could consider checking for illegal values in the `io_relay` and not pass those requests on to MotoROS. This would also allow us to return meaningful error messages to users.

gavanderhoorn:
This PR uses the updated `result_code` values added in #369 (which hasn't been merged yet) to provide ROS service clients with more descriptive error messages in case something goes wrong.

Examples of messages returned:

```shell
$ rosservice call /read_single_io "address: 10018"
message: "Read failed (address: 10018): Illegal address for read: outside permitted range on this controller, see documentation (1001)"
success: False
value: 0

$ rosservice call /write_single_io "{ address: 0, value: 0 }"
message: "Write failed (address: 0): Illegal address for write: outside permitted range on this controller, see documentation (1002)"
success: False

$ rosservice call /write_single_io "{ address: 10010, value: 2 }"
message: "Write failed (address: 10010): Illegal value for the type of IO element addressed (1003)"
success: False
```

Failed `mpReadIO` and `mpWriteIO` have similar messages, but I can't seem to trigger those :)

--

Forgot to mention I believe this would address / fix #367.

### DIFF
--- a/motoman_driver/include/motoman_driver/simple_message/motoman_read_single_io_reply.h
+++ b/motoman_driver/include/motoman_driver/simple_message/motoman_read_single_io_reply.h
@@ -57,15 +57,15 @@ namespace io_ctrl_reply
 /**
  * \brief Enumeration of Read Single IO reply result codes.
  */
-namespace ReadSingleIOReplyResults
+namespace ReadSingleIOReplyResultCodes
 {
-enum ReadSingleIOReplyResult
+enum ReadSingleIOReplyResultCode
 {
   FAILURE    = 0,
   SUCCESS    = 1
 };
 }
-typedef ReadSingleIOReplyResults::ReadSingleIOReplyResult ReadSingleIOReplyResult;
+typedef ReadSingleIOReplyResultCodes::ReadSingleIOReplyResultCode ReadSingleIOReplyResultCode;
 
 /**
  * \brief Class encapsulated read single io reply data.  These messages are sent
@@ -111,7 +111,7 @@ public:
    *
    */
   void init(industrial::shared_types::shared_int value,
-            ReadSingleIOReplyResult result_code);
+            ReadSingleIOReplyResultCode result_code);
 
   /**
    * \brief Sets value
--- a/motoman_driver/include/motoman_driver/simple_message/motoman_read_single_io_reply.h
+++ b/motoman_driver/include/motoman_driver/simple_message/motoman_read_single_io_reply.h
@@ -61,8 +61,12 @@ namespace ReadSingleIOReplyResultCodes
 {
 enum ReadSingleIOReplyResultCode
 {
-  FAILURE    = 0,
-  SUCCESS    = 1
+  SUCCESS               =    0,
+  READ_ADDRESS_INVALID  = 1001, // The ioAddress cannot be read on this controller
+  WRITE_ADDRESS_INVALID = 1002, // The ioAddress cannot be written to on this controller
+  WRITE_VALUE_INVALID   = 1003, // The value supplied is not a valid value for the addressed IO element
+  READ_API_ERROR        = 1004, // mpReadIO returned -1
+  WRITE_API_ERROR       = 1005, // mpWriteIO returned -1
 };
 }
 typedef ReadSingleIOReplyResultCodes::ReadSingleIOReplyResultCode ReadSingleIOReplyResultCode;
--- a/motoman_driver/src/io_ctrl.cpp
+++ b/motoman_driver/src/io_ctrl.cpp
@@ -65,7 +65,7 @@ bool MotomanIoCtrl::init(SmplMsgConnection* connection)
   return true;
 }
 
-bool MotomanIoCtrl::readSingleIO(shared_int address, shared_int &value)
+bool MotomanIoCtrl::readSingleIO(shared_int address, shared_int &value, std::string &err_msg)
 {
   ReadSingleIOReply reply;
 
@@ -78,11 +78,15 @@ bool MotomanIoCtrl::readSingleIO(shared_int address, shared_int &value)
   value = reply.getValue();
 
   bool read_success = reply.getResultCode() == ReadSingleIOReplyResultCodes::SUCCESS;
+  if (!read_success)
+  {
+    err_msg = reply.getResultString();
+  }
 
   return read_success;
 }
 
-bool MotomanIoCtrl::writeSingleIO(shared_int address, shared_int value)
+bool MotomanIoCtrl::writeSingleIO(shared_int address, shared_int value, std::string &err_msg)
 {
   WriteSingleIOReply reply;
 
@@ -93,6 +97,10 @@ bool MotomanIoCtrl::writeSingleIO(shared_int address, shared_int value)
   }
 
   bool write_success = reply.getResultCode() == WriteSingleIOReplyResultCodes::SUCCESS;
+  if (!write_success)
+  {
+    err_msg = reply.getResultString();
+  }
 
   return write_success;
 }
--- a/motoman_driver/src/io_ctrl.cpp
+++ b/motoman_driver/src/io_ctrl.cpp
@@ -41,8 +41,8 @@
 #include <string>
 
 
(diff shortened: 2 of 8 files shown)


motoman_driver/include/motoman_driver/simple_message/motoman_read_single_io_reply.h
```diff
@@ -57,15 +57,19 @@ namespace io_ctrl_reply
...
  */
-namespace ReadSingleIOReplyResults
+namespace ReadSingleIOReplyResultCodes
 {
-enum ReadSingleIOReplyResult
+enum ReadSingleIOReplyResultCode
 {
-  FAILURE    = 0,
-  SUCCESS    = 1
+  SUCCESS               =    0,
```
gavanderhoorn, line 64: It seems MotoROS, prior to #369, returned `0` for failure instead of success.

@ted-miller: did you change that on purpose? Could be we made an incorrect assumption in this code here though.

motoman_driver/src/io_relay.cpp
```diff
@@ -106,19 +106,21 @@ bool MotomanIORelay::readSingleIoCB(
...
+  bool result = io_ctrl_.readSingleIO(req.address, io_val, err_msg);
   this->mutex_.unlock();
 
   if (!result)
   {
     res.success = false;
 
-    // provide caller with failure indication (not very informative yet)
+    // provide caller with failure indication
+    // TODO: should we also return the result code?
```
gavanderhoorn, line 121: This would require extending the message.

motoman_driver/src/io_ctrl.cpp
```diff
@@ -65,7 +65,7 @@ bool MotomanIoCtrl::init(SmplMsgConnection* connection)
   return true;
 }
 
-bool MotomanIoCtrl::readSingleIO(shared_int address, shared_int &value)
+bool MotomanIoCtrl::readSingleIO(shared_int address, shared_int &value, std::string &err_msg)
```
gavanderhoorn, line 68: Not too happy with adding yet another *out* argument, but for now this works.

We'll probably want to refactor the IO client (to add support for IO groups and registers), so I did not want to change too much here in this PR.

gavanderhoorn:
@steviedale @marip8: would you have time to take a look at this?

motoman_driver/include/motoman_driver/simple_message/motoman_read_single_io_reply.h
ted-miller: Prior to 369, the resultCode was being assigned with an enum of type SmReplyType. 
https://github.com/ros-industrial/motoman/blob/kinetic-devel/motoman_driver/MotoPlus/SimpleMessage.h#L95

I believe this was wrong and should have actually been using SmResultType, which has a different value for success.
https://github.com/ros-industrial/motoman/blob/kinetic-devel/motoman_driver/MotoPlus/SimpleMessage.h#L123

For other functions in the application, a status of '0' indicates success. Any other value is an error code.

ted-miller: (But no, it wasn't on purpose. I just now discovered the different enum.)

motoman_driver/src/io_relay.cpp
marip8: Do you mean providing something more than stating the result code in the error message?

motoman_driver/src/simple_message/motoman_read_single_io_reply.cpp
```diff
@@ -80,9 +80,17 @@ std::string ReadSingleIOReply::getResultString(shared_int result_code)
...
+  case ReadSingleIOReplyResultCodes::READ_ADDRESS_INVALID:
+     return "Illegal address for read: outside permitted range on this controller, see documentation (1001)";
+  case ReadSingleIOReplyResultCodes::WRITE_ADDRESS_INVALID:
+     return "Illegal address for write: outside permitted range on this controller, see documentation (1002)";
+  case ReadSingleIOReplyResultCodes::WRITE_VALUE_INVALID:
+     return "Illegal value for the type of IO element addressed (1003)";
+  case ReadSingleIOReplyResultCodes::READ_API_ERROR:
+     return "The MotoPlus function MpReadIO returned -1. No further information is available (1004)";
+  case ReadSingleIOReplyResultCodes::WRITE_API_ERROR:
+     return "The MotoPlus function MpWriteIO returned -1. No further information is available (1005)";
```
marip8: It might be worth changing the strings to use the value directly from the enumeration rather than hard-coding the numbers, just in case they change in the future, i.e.:

```c++
return "Illegal address for read: outside permitted range on this controller, see documentation (" + std::to_string(ReadSingleIoReplyResultCodes::READ_ADDRESS_INVALID) + ")";
```

motoman_driver/include/motoman_driver/simple_message/motoman_read_single_io_reply.h
```diff
@@ -57,15 +57,19 @@ namespace io_ctrl_reply
...
+  READ_ADDRESS_INVALID  = 1001, // The ioAddress cannot be read on this controller
+  WRITE_ADDRESS_INVALID = 1002, // The ioAddress cannot be written to on this controller
+  WRITE_VALUE_INVALID   = 1003, // The value supplied is not a valid value for the addressed IO element
+  READ_API_ERROR        = 1004, // mpReadIO returned -1
+  WRITE_API_ERROR       = 1005, // mpWriteIO returned -1
 };
 }
-typedef ReadSingleIOReplyResults::ReadSingleIOReplyResult ReadSingleIOReplyResult;
+typedef ReadSingleIOReplyResultCodes::ReadSingleIOReplyResultCode ReadSingleIOReplyResultCode;
```
marip8, line 72: Out of curiosity, why are the result code classes wrapped in a namespace if we are just going to `typedef` the namespace away?

marip8 (dismissed):
I think this looks good and should help make the I/O functionality more usable. I made a few small suggestions, but nothing that should hold up this PR

motoman_driver/include/motoman_driver/simple_message/motoman_read_single_io_reply.h
gavanderhoorn: This is essentially cargo-cult. I'm just maintaining/duplicating what was there already.

Not sure why it was done this way, perhaps that was the best available at the time.

motoman_driver/src/io_relay.cpp
gavanderhoorn: Yes. We could add the code itself in a `uint32` field fi.

That would make more automated handling of errors possible. The current message (with a `str` result message) is nice for humans, not so much for machines.

motoman_driver/src/simple_message/motoman_read_single_io_reply.cpp
gavanderhoorn: I'll see if I change this. We're not really using C++11 anywhere in this codebase, this would be a divergence, but that could be OK.

gavanderhoorn: Decided it was worth it. Driver now builds with C++11 enabled, so we get to use `std::to_string(..)`.

thanks for the suggestion.

motoman_driver/src/io_relay.cpp
gavanderhoorn: Let's do this in a follow-up PR.

gavanderhoorn:
I've tested this again with the changes I added in the last two commits.

As this is my own PR: could you review @marip8 and :+1:  if OK?

Then we can merge this today.

marip8 (approved):
Looks good. Thanks for adding this feature

Issue #370, opened 2020-11-16
Xacro mh12 correction 


leitoxve:
in 
https://github.com/ros-industrial/motoman/blob/kinetic-devel/motoman_mh12_support/urdf/mh12_macro.xacro

inverse the flench translate values in xacro 
line 144 with 149
**ORIGINAL**
144-origin xyz="0 0 0" rpy="0 0 0" 
149-origin xyz="0.100 0 0" rpy="3.1415926 -1.570796 0"
to 
**MODIFICATION**
144-origin xyz=**"0.10 0 0**" rpy="0 0 0" 
149-origin xyz=**"0.000 0 0"** rpy="3.1415926 -1.570796 0"

gavanderhoorn:
It would be good if you could provide a little more context and explanation here.

Do you have a particular reason you want this change applied?

EricMarcil:
The standard is that Tool0 represent the flange and it should not be edited.
If you are adding a tool to the robot then you should create a new joint with Tool0 as the parent.
Technically, the change that you are proposing shouldn't change the result.  
The way it is currently done, matches the way Yaskawa does its modeling in our simulator (MotoSim) except that we call Flange what Ros calls Tool0.

gavanderhoorn:
I'll try not to confuse everyone by mentioning the `flange` frame (I'll mark my comment as off-topic), but we do have a `flange` frame. It just wouldn't be aligned with any TCP, but with REP-103 standard orientation of frames. This would be for connection of EEF geometry to xacro macros from robot support packages.

See [REP-199](http://gavanderhoorn.github.io/rep/rep-0199.html) for a WIP proposal (I really need to finish that). Ignore the *Vendor Nomenclature Mapping*, that will be removed.

For additional tool frames, they should not be made children of `tool0`, but siblings, and then have `flange` as their parent.

`tool0` must not be changed, and the difference between `tool0` and `flange` would be the orientation only on most robots.

gavanderhoorn:
I'm going to close this, as I believe @EricMarcil's response should clear this up.

@leitoxve: please feel free to keep commenting on the issue.

If it turns out we should actually change or fix something here, we'll re-open.

leitoxve:
what I mean is that in real life the robot does not have a washer in the last joint  in the middle of nowhere
![Screenshot from 2020-11-27 15-45-33](https://user-images.githubusercontent.com/44784957/100460532-a16f8e80-30c7-11eb-9a31-d22d7133c096.png)

[...]

PR #372, opened 2020-11-30
MH12 T-axis Mesh Correction

EricMarcil:
Updated the meshes of the MH12 t-axis link to correct the mesh origin so that the link displays correctly at the end of the robot.


Gaspard-Bourgeois (dismissed):
I checked first that the URDF was matching the MotoSim model as told in [https://github.com/ros-industrial/motoman/issues/370](https://github.com/ros-industrial/motoman/issues/370). It is alright.

The new STL for both visualisation and collision are well located according to the URDF. Good job.

gavanderhoorn (approved):
Looks correct.

gavanderhoorn:
Thanks @EricMarcil for the fix.

Issue #375, opened 2020-12-05
GP180 support


Doomerdinger:
There is support for the GP180-120 but not the base GP-180.

PR #376, opened 2020-12-05
Added gp180 support.

Doomerdinger:
Added support for GP180.

See #375


Doomerdinger:
Going to make some improvements before reopening.

gavanderhoorn:
@Doomerdinger: would you be willing to resubmit this?

Even if the improvements you refer to would be an approach to resolve #390, it would be good to first merge the basic model here.

Solving #390 might take some time.

Doomerdinger:
There were a couple of things, but yes that was one of them.

I can reopen this and a couple other arms and such that I have going in the next couple days. My arms do rely on changes I made in [PR378](https://github.com/ros-industrial/motoman/pull/378) so if I could ask you to take another look at that it'd smooth things out.

Issue #314, opened 2020-02-09
error material 'yaskawa_blue' is not unique when using multiple robots


dignakov:
When using multiple robots each seems to add a definition for yaskawa_blue, causing this error:

`[ERROR] [1581285016.618736830]: material 'yaskawa_blue' is not unique.`

System: Ubuntu 18.04, ROS-melodic, motoman package from git.

example xacro (my_robot_workspace.xacro):
```
<?xml version="1.0" ?>

<robot name="my_robot_workspace" xmlns:xacro="http://ros.org/wiki/xacro">
    <xacro:include filename="$(find motoman_gp25_support)/urdf/gp25_macro.xacro" />

    <xacro:motoman_gp25 prefix="first_"/>
    <xacro:motoman_gp25 prefix="second_"/>

    <link name="world" />

    <joint name="world-first_base_link" type="fixed">
        <origin xyz="0 0 0" rpy="0 0 0"/>
        <parent link="world"/>
        <child link="first_base_link"/>
    </joint>

    <joint name="world-second_base_link" type="fixed">
        <origin xyz="0 1.5 0" rpy="0 0 0"/>
        <parent link="world"/>
        <child link="second_base_link"/>
    </joint>

</robot>

```

load_my_robot.launch
```
<launch>
	<param name="robot_description" command="$(find xacro)/xacro -\-inorder '$(find my_robot)/urdf/my_robot_workspace.xacro'" />
</launch>
```

test_robot.launch
```
<?xml version="1.0" ?>
<launch>
	<include file="$(find my_robot)/launch/load_my_robot.launch" />
	<node name="joint_state_publisher" pkg="joint_state_publisher" type="joint_state_publisher" >
		<param name="use_gui" value="true" />
	</node>
	<node name="robot_state_publisher" pkg="robot_state_publisher" type="robot_state_publisher" />
    <node name="rviz" pkg="rviz" type="rviz" args="-d $(find my_robot)/config/rviz_config.rviz" required="true" />

</launch>
```

gavanderhoorn:
Yes, this is an unfortunate side-effect of all macros redefining `yaskawa_blue`.

We could introduce a `motoman_resources` package (similar to `fanuc_resources`, `abb_resources`, etc) which provides this definition once and then reuse it everywhere.

[...]

PR #378, opened 2020-12-07
Added motoman_resources package.

Doomerdinger:
Made robots use materials from that package so they do not conflict when using multiple arms.

Also added a common macro file that can be used to load kinematic parameters and other values.

Related to #314

### DIFF
--- a/motoman_resources/urdf/common_colors.xacro
+++ b/motoman_resources/urdf/common_colors.xacro
@@ -1,0 +1,6 @@
+<?xml version="1.0" ?>
+<robot xmlns:xacro="http://ros.org/wiki/xacro">
+  <xacro:property name="color_yaskawa_blue" value="${33/255} ${38/255} ${191/255} 1.0"/>
+
+  <xacro:property name="color_yaskawa_white" value="0.99 0.99 0.99 1"/>
+</robot>
--- a/motoman_gp12_support/package.xml
+++ b/motoman_gp12_support/package.xml
@@ -43,6 +43,7 @@
 
   <exec_depend>joint_state_publisher</exec_depend>
   <exec_depend>motoman_driver</exec_depend>
+  <exec_depend>motoman_resources</exec_depend>
   <exec_depend>robot_state_publisher</exec_depend>
   <exec_depend>rviz</exec_depend>
   <exec_depend>xacro</exec_depend>
--- a/motoman_gp12_support/urdf/gp12_macro.xacro
+++ b/motoman_gp12_support/urdf/gp12_macro.xacro
@@ -2,10 +2,7 @@
 <robot xmlns:xacro="http://ros.org/wiki/xacro">
 <xacro:macro name="motoman_gp12" params="prefix">
 
-    <!-- Properties -->
-    <material name="yaskawa_blue">
-          <color rgba="0.12941 0.14902 0.74902 1" />
-    </material>
+    <xacro:include filename="$(find motoman_resources)/urdf/common_materials.xacro"/>
 
     <!-- link list -->
     <link name="${prefix}base_link">
@@ -13,7 +10,7 @@
             <geometry>
                 <mesh filename="package://motoman_gp12_support/meshes/visual/gp12_base_link.stl"/>
             </geometry>
-            <material name="yaskawa_blue"/>
+            <xacro:material_yaskawa_blue/>
         </visual>
         <collision>
             <geometry>
@@ -26,7 +23,7 @@
             <geometry>
                 <mesh filename="package://motoman_gp12_support/meshes/visual/gp12_link_1_s.stl"/>
             </geometry>
-            <material name="yaskawa_blue"/>
+            <xacro:material_yaskawa_blue/>
         </visual>
         <collision>
             <geometry>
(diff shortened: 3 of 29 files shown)


motoman_resources/urdf/common_macros.xacro
```diff
@@ -0,0 +1,234 @@
+<?xml version="1.0"?>
```
gavanderhoorn: I would recommend not using this macro here.

There are other ways to load `.yaml` data which are nicer, more maintainable and easier to understand.

Doomerdinger: Fair enough, any suggestions?

I was just using this because in the back of my mind I am considering trying to do something akin to what the UR descriptions package has going on, with one file for all the arms then config files to make all the different robots. This seemed like a decent way to consistently load up the values, dictating names in the configs with the xacro soruce.
Obviously it won't work for all the Yaskawa arms, but it seems like many of them fall into the standard 6 dof arm category.

Doomerdinger: If you don't think there's any value in a future shift like that, I can abandon this configuration idea altogether. Not hard set on it, was just a thought.

I am currently working on loading kinematic parameters for some arms, and this was just an extension of that.

gavanderhoorn: It's a nice idea, but I'm not sold on it (yet).

There are a couple of reasons we duplicate so much in these repositories:

 - it's easier to grok what is going on (I know I helped rewrite the `universal_robot` xacros to rely on `.yaml` for just about everything, but tbh it doesn't really get any clearer)
 - easier to make changes to individual models (ie: things which are not part of the `.yaml`)
 - less moving parts
 - packages have less dependencies
 - not all xacro parsers actually support loading `.yaml` files (several are in use in external projects, which will not be able to load the models any more)

Especially the first bullet is not something to underestimate. Cognitive overload is easy to achieve, and with everything going in ROS, adding more complexity isn't always a good thing.

As I wrote: I'm not against it per se, I just think we should be a bit careful about importing the way it was done in `universal_robot`.

An alternative could be to pass entire dicts from the `.yaml` to `xacro:macro`s, as Robert describes here: https://github.com/ros/xacro/issues/240#issuecomment-636302854.

gavanderhoorn:
I'll do a proper review later, some high-level comments now:

 - please don't manually bump versions of packages. This will all be done automatically by the release pipeline. Please revert all those changes to the manifests
 - same goes for the `CHANGELOG` file: it's added automatically, so please revert the addition
 - please add at least one ROS-I maintainer (me). See the other packages for the metadata to add to the manifest
 - and it's a minor point, but I believe you've copied the structure/approach from `fanuc_resources` or one of the other similar packages. I realise we're in the same 'context' here (ie: ROS-Industrial), but technically, I believe it would've been good to at least mention that somewhere, as the license of those packages requires

motoman_resources/urdf/common_colors.xacro
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" ?>
+<robot xmlns:xacro="http://ros.org/wiki/xacro">
+    <xacro:property name="color_yaskawa_blue" value="${33/2355} ${38/255} ${191/255} 1.0"/>
```
gavanderhoorn:
```suggestion
    <xacro:property name="color_yaskawa_blue" value="${33/255} ${38/255} ${191/255} 1.0"/>
```

Doomerdinger:
> I'll do a proper review later, some high-level comments now:
> 
> * please don't manually bump versions of packages. This will all be done automatically by the release pipeline. Please revert all those changes to the manifests
> * same goes for the `CHANGELOG` file: it's added automatically, so please revert the addition
> * please add at least one ROS-I maintainer (me). See the other packages for the metadata to add to the manifest
> * and it's a minor point, but I believe you've copied the structure/approach from `fanuc_resources` or one of the other similar packages. I realise we're in the same 'context' here (ie: ROS-Industrial), but technically, I believe it would've been good to at least mention that somewhere, as the license of those packages requires

Didn't realize the release pipeline took care of versions/changelog, I'll be sure to not do that moving forward.
I'll add you to the packages, and mention the fanuc package.

motoman_resources/urdf/common_macros.xacro
Doomerdinger: I agree with the complexity notion.

Alright, for my purposes I only _need_ the kinematics as input at the moment.

What do you think of leaving the nominal kinematic values in the URDF, and load offsets from nominal from a yaml file for the purpose of calibration? This wouldn't involve a separate macro, the dictionary containing the values would be the xacro argument.

Doesn't have the complexity and flexibility of the other approach, still allows for calibration, but makes the nominal values clear for users that do not care about that, and the loading of the yaml optional (can fallback to all 0s for offsets if no dictionary is passed in)

Edit:
Perhaps I am missing something, bu it does not appear that there is a good way to emulate loading a yaml (to get the `${param['x']['y']}` syntax) without actually loading a yaml. So for the above approach to work correctly with parsers that cannot load yaml I would need either a bunch of `if else` statements to check and then only add 0 to the offset, or a bunch of parameters that are set to 0 if no offsets are found, or to the actual offsets if they are.

Edit2:
I'm just blind.
`<xacro:property name="props" value="${dict(a=1, b=2, c=3)}"/>`
unless that is also not supported by the parsers in question.

gavanderhoorn:
Replacement in #405 (with additional commits migrating more support packages to use the resources package).

Thanks for the initial PR @Doomerdinger :+1:

Issue #377, opened 2020-12-06
MH5 Support Update


acbuynak:
I am working on an update to the MH5 support package.
Primarily changing things to follow ros-i conventions, ie [REP-0199](http://gavanderhoorn.github.io/rep/rep-0199.html).

I have access to a [MH5LSII](https://www.motoman.com/en-us/products/robots/industrial/assembly-handling/mh-series/mh5ls-ii) variant.

The current MH5 support package matches the MH5F variant from what I can tell (c72c98c)(it's not notated specifically in package.xml). AFAIK, the MH5F variant is now [archived](https://www.motoman.com/en-us/products/product-archive) and not sold new.

There are physical differences in link (L&R) shape and length between the "F" and "LSII/SII" variants.

I think it would be helpful to new users to change the default support model to match what Motoman sells, the MH5LSII or MH5SII (differs by long or short variant). If not I'm happy to push to motoman_experimental or otherwise.

Looking for guidance. Thanks!

EricMarcil:
Hi,

The MH5F (FS100) and MH5SII (DX200) and MH5S(DX100) are the same (physically and parameters)
The MH5LF (FS100) and MH5SLII (DX200) and MH5SL(DX100) are the same (physically and parameters)
For the long model, it has a longer L-axis (different part) and an extension to the U-axis (front of U-axis the arm is the same).

I took a look at the current URDF on github, and I can't the reference to the current values.  The physical model matches the later version with the slim (tapered) R-axis.  But the parameters are probable from the initial model (none tapered R-axis) for which there were a lot of variants.

The values that I currently have for the short: 
![image](https://user-images.githubusercontent.com/4691677/101783448-49258d00-3ac8-11eb-88e4-19196d7509dd.png)

and for the long model:
![image](https://user-images.githubusercontent.com/4691677/101784167-28116c00-3ac9-11eb-87d0-07e5bb67cd4a.png)

Below are the spec files for each models:
[MH5-series.zip](https://github.com/ros-industrial/motoman/files/5673410/MH5-series.zip)

So I would tend to agree that the model could be updated and we could make a variant for the long model.

acbuynak:
Thanks for the info! I'm pulling together an updated short "MH5" and a new long "MH5L" support package. I'll get those into a PR soon.

> I can't the reference to the current values. The physical model matches the later version with the slim (tapered) R-axis. But the parameters are probably from the initial model (none tapered R-axis) for which there were a lot of variants.

[...]

PR #380, opened 2021-01-05
Update MH5: Add Short & Long Variants

acbuynak:
Overhaul of MH5 Support package. 

Defined two physically and parametrically unique variants (short & long) of the MH5 and stored within a single MH5 package. Kept support files for each variant separate excepting the `joint_names_mh5.yaml`. Some of the mesh files are the same/similar, but I kept the meshes grouped by variant for overall clarity.

Initial work tracked in Issue #377 

Prior to merging a release date needs assigned in the drafted `changelog.rst`


EricMarcil:
@acbuynak: Thanks for the contribution.  Here are a few things to review:
- I don't think we need to repeat all the meshes that are identical between the MH5 and MH5L.  You can probably just keep the link_3_u.stl file in the MH5L and make reference to the MH5 for the other models.
- Some of the visual models seems a bit large.  I normally try to get them down below 1 MB but we don't want to loose to much detail.  So it's a question of balancing the two.
- For the collision models, we usually do a convex hull around the model to reduce the size as much as possible.  You models seem more detailled but when I checked the model size, they were not that bad.  So they are probably fine.

EricMarcil:
I've complete checking the motion direction and range by comparing with our simulator (MotoSim).  It's all matching up.

acbuynak:
@EricMarcil - Thanks for checking the PR.

Responses to your above bullet points...
1 - Repeated Meshes. I understand the need to reduce overall package size, but wanted to emphasize that this package contained two models by making a distinction in the meshes folder.
2 - Mesh Size. Agreed, I missed that. Thanks. I will simplify (quadratic edge collapse decimation in MeshLab) to something more reasonable that retains sufficient fidelity.
3 - Collision Mesh Detail. When testing the models, I discovered that a convex hull caused a false collision of links L & R at the extreme joint limit of U. The correction was 22f11d6 to allow the "V" cutout in the L link to be slotted with the R link.

gavanderhoorn:
Thanks for the PR.

I'll try to take a look at it today.

At a high-level and as a first comment: I do agree with @EricMarcil that we'd rather reuse meshes as much as possible. This reduces total size of the package and maintenance as well (as we don't have to keep two sets of meshes in sync).

The fact the package supports multiple variants should already be clear from the `_macro.xacro` files, the package manifest and the various `.launch` files.

acbuynak:
Sounds good. I've drafted a commit to consolidate meshes. I'm planning to keep the 'unique' MH5L mesh files in their own folder, the rest will reference to the MH5.

Just a minor note...
I did a side-by-side comparison of each short & long variant mesh. I found the file I have for the U-joints to be different. Long variant meshes were sourced from Motoman Tech Support.


**I'm leaning to use the short variant (right image) for both URDF's.** They're close enough. Collision mesh varies by ~1cm at abs max. ~~I note this because my actual robot (long variant) matches the short mesh file shown.~~ _(edit. error, I have the left shown u-joint)_ This may be a feature option to the robot, rather than short/long variant.
Thoughts?

Summary btwn Short/Long meshes:
-Unique: L, U, R
-Common: base, S, B, T 

mh5_u_link

EricMarcil:
That's interesting.  All the models (long and short) that I have in MotoSim matches your left picture (MH5L).  But I know that the MH5 and MH5L have a lot of variations.  So it is quite possible that you have the right variant.
But, I believe that the left model is the most common.
I would recommend putting the right model, in a folder MH5_other or something similar with the name: link3_u_old.  Then maybe in the urdf, have a commented reference to it that people can easily customize depending on their needs.

Out of curiousity, can you send me the exact model type (ex: YR-MH0005S-C00) that you have on the name plate on the robot arm.  And confirm that your robot is a DX200.  Maybe also send me your all.prm file, I'll double check that is is matching the standard values that I had sent you before.  You can use my work e-mail: eric.marcil@motoman.com

acbuynak:
@EricMarcil - Emailed you my parameters file.

For transparency, I caught a mistake above. My physical MH5L U-Link matches the long variant shown on the left as EricMarcil noted is common.

I like the suggestion of retaining a `link_3_u_old` files w/ commented option in URDF. I will add this in next commit.

gavanderhoorn:
Closing as we now have #620.

Thanks again @acbuynak for your work :+1: We'll merge it as part of #620.

And apologies for leaving it open for this long :disappointed:

PR #381, opened 2021-01-06
Add CMake support for MotoPlus builder

Gaspard-Bourgeois:
Well done Julien. This file is working great in our lab, I support the pull request.

### DIFF
--- a/motoman_driver/MotoPlus/CMakeLists_with_vs_support.txt
+++ b/motoman_driver/MotoPlus/CMakeLists_with_vs_support.txt
@@ -1,3 +1,6 @@
+#With Visual Studio support: existing .vcxproj (created by Visual Studio) and linked mps files (created by hand?) in git repository
+#Note: remove "_with_vs_support" in filename to normally use this file with CMake
+
 cmake_minimum_required(VERSION 3.0)
 
 project(MpRosAllControllers NONE)
@@ -33,6 +36,7 @@ macro(PROJECT_CONFIG name controller suffix)
                       ${name}${suffix}${PROJECT_VERSION}.out)
 endmacro(PROJECT_CONFIG)
 
+#Adding support for the application on several controllers
 project_config(MotoRos DX100 DX1_)
 project_config(MotoRos DX200 DX2_)
 project_config(MotoRos FS100 FS_)
--- a/motoman_driver/MotoPlus/CMakeLists_without_vs_support.txt
+++ b/motoman_driver/MotoPlus/CMakeLists_without_vs_support.txt
@@ -1,0 +1,159 @@
+#Without Visual Studio support: .vcxproj and linked mps files are automatically generated on the fly by CMake macro
+#Note: remove "_without_vs_support" in filename to normally use this file with CMake
+
+cmake_minimum_required(VERSION 3.0)
+
+project(MotoRos NONE)
+
+set(SRCS
+    ./Controller.c
+    ./CtrlGroup.c
+    ./IoServer.c
+    ./MotionServer.c
+    ./mpMain.c
+    ./SimpleMessage.c
+    ./StateServer.c
+    )
+
+set(HEADERS
+    ./Controller.h
+    ./CtrlGroup.h
+    ./IoServer.h
+    ./MotionServer.h
+    ./MotoROS.h
+    ./ParameterExtraction.h
+    ./ParameterTypes.h
+    ./SimpleMessage.h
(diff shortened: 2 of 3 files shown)


gavanderhoorn:
Thanks for the PR.

I'm wondering though: this seems to duplicate quite a bit of information which is already part of the various `.mps` files and the existing visual studio solution file.

I'm not against merging this, but duplication would be something to avoid.

Could you not have reused the `.mps` files?

Julien-Livet:
Thanks for the feed back !

Originally, I wrote this file to use another IDE than MotoPlus IDE. I talked about that with Yaskawa France and it was possible with MpBuilder, and possibly without having Visual Studio IDE but using Visual Studio file project. So the purpose of my original commit.

But for this project, it's more interesting to use each mps file and existing vxproj file, so the updated commit.

gavanderhoorn:
So to get it clear: this adds a CMake macro which generates command lines to instruct `mpbuilder.exe` (from the MotoPlus SDK) to build MotoROS. Correct?

> Originally, I wrote this file to use another IDE than MotoPlus IDE. I talked about that with Yaskawa France and it was possible with MpBuilder, and possibly without having Visual Studio IDE but using Visual Studio file project. So the purpose of my original commit.
> 
> But for this project, it's more interesting to use each mps file and existing vxproj file, so the updated commit.

thanks for this comment, I now understand better what you're trying to do.

I believe it would be fine to have some infrastructure to build MotoROS with `mpbuilder.exe` but without Visual Studio. If that was your original intent, we could merge that here, but we'd probably have to a) make that very clear and b) not name it `CMakeLists.txt`, but give it some sort of suffix to clearly indicate the use-case.

Julien-Livet:
> So to get it clear: this adds a CMake macro which generates command lines to instruct mpbuilder.exe (from the MotoPlus SDK) to build MotoROS. Correct?

Exactly! Thanks for your answer!

> I believe it would be fine to have some infrastructure to build MotoROS with mpbuilder.exe but without Visual Studio. If that was your original intent, we could merge that here, but we'd probably have to a) make that very clear and b) not name it CMakeLists.txt, but give it some sort of suffix to clearly indicate the use-case.

I agree with your comment, I will see to re-think the content on a clear way and propose an appropriated name for it. Thank you very much!

gavanderhoorn (changes requested):
Marking as *Changes Requested* so we don't accidentally merge.

Julien-Livet (changes requested):
Update CMake support

Gaspard-Bourgeois:
it seems OK to me

gavanderhoorn:
I'm not sure I follow @Julien-Livet and @Gaspard-Bourgeois: why did you close the PR?

dhled:
Hey @gavanderhoorn . I think @Gaspard-Bourgeois  didn't know that he was closing the PR. 
I think he is new to open source community :). We will reach him. 

Sorry for that !

Gaspard-Bourgeois:
I'm sorry, I did'nt know closing meant that

motoman_driver/MotoPlus/CMakeLists_with_vs_support.txt
```diff
@@ -0,0 +1,44 @@
...
+                      ${path}\\${name}${suffix}.out
+                      ${name}${suffix}${PROJECT_VERSION}.out)
+endmacro(PROJECT_CONFIG)
+
+#Adding support for the application on several controllers
+project_config(MotoRos DX100 DX1_)
+project_config(MotoRos DX200 DX2_)
+project_config(MotoRos FS100 FS_)
+project_config(MotoRos YRC1000 YRC1u_)
+project_config(MotoRos YRC1000u YRC1_)
```
cjue, line 44: The controller names and suffixes are mixed up here

motoman_driver/MotoPlus/CMakeLists_without_vs_support.txt
```diff
@@ -0,0 +1,159 @@
+#Without Visual Studio support: .vcxproj and linked mps files are automatically generated on the fly by CMake macro
```
cjue, line 1: I don't understand the purpose of this file.

The other file uses the information already present in the mps file and avoids any duplication. It allows generating build configuration for VS, Makefiles or any other CMake target, that will use the mpbuilder.exe from `MotoPlus for Visual Studio`, which is useful.

This file just produces the same results, only with the difference of spelling everything out explicitly?

Issue #338, opened 2020-06-25
Robot status motion possible flag incorrect when robot enable called during PFL stop


CameronDevine:
I am controlling a Yaskawa HC10 with the power force limiting features enabled from ROS. I have found that if the robot is in  PFL stop, and the `/robot_enable` service is called, the `motion_possible` flag in the `/robot_state` messages is set to true, but the robot cannot be moved.

I am using my own branch of `motoman_driver` (with the master branch recently merged in) available here, https://github.com/CameronDevine/motoman/tree/point-streaming-with-io

I believe I am using MotoROS v3.6.0, but I can check early next week. I am also planning to test this with the current master branch.

gavanderhoorn:
> I believe I am using MotoROS v3.6.0

that would be interesting, as we're only at `v1.9.1` with MotoROS.

> I am using my own branch of `motoman_driver` (with the master branch recently merged in) available here, https://github.com/CameronDevine/motoman/tree/point-streaming-with-io

Diff [here](https://github.com/ros-industrial/motoman/compare/kinetic-devel...CameronDevine:point-streaming-with-io
).

I see no changes to MotoROS here, only to the `motoman_driver` side.

> I am controlling a Yaskawa HC10 with the power force limiting features enabled from ROS. 

What do you mean exactly by "enabled from ROS"? There is no explicit support for any PFL features in `motoman_driver`, so I'm not sure I understand what you write here.

> I have found that if the robot is in PFL stop, and the `/robot_enable` service is called, the `motion_possible` flag in the `/robot_state` messages is set to true, but the robot cannot be moved.

This sounds a bit like #287.

@ted-miller @EricMarcil: could it be that PFL status is not taken into account when calculating the values for those fields?

CameronDevine:
I'm sorry, when I wrote this up I was a little hurried.

I believe I am running MotoROS v1.9.0. Is there any way to check which version is installed on the controller?

I am using the HC10 robot with the PFL feature turned on and sending motion commands to the robot from ROS.

gavanderhoorn:
>  Is there any way to check which version is installed on the controller?

I don't believe so (not until we get #193 going).

What you could do is copy the `.out` file to a USB stick, generate the `md5sum` and compare that to the hashes shown [here](http://wiki.ros.org/motoman_driver/Tutorials/indigo/InstallServer#Verifying_integrity_of_the_download) (you'd be essentially following the instructions there).

For `v1.9.0` it would be:

```
[...]

PR #382, opened 2021-01-06
MotoRos 1.9.3 - PFL Handling Improvement

EricMarcil:
These modifications should only affect HC-series robot which are collaborative robots with the PFL (Power Force Limiting) function: 
- Fixed motion_possible status when PFL is active
- Added PFL Avoidance signal
- Added PFL flag to prevent continuing previous motion after PFL activation


EricMarcil:
### MotoROS PFL Handling

Following the bug report #338 "Robot status motion possible flag incorrect when robot enable called during PFL stop" on the github motoman repo, the MotoROS PFL handling was review.
It was found that if the robot is in PFL stop, and the `/robot_enable` service is called, the `motion_possible` flag in the `/robot_state` messages is set to `true`, but the robot cannot be moved.
Furthermore, the PFL Avoidance mode was not originally covered.
Under some PFL modes, some motion could occur after PFL activation completed but that motion was no longer synchronized with the desired ROS motion.


### PFL (Power Force Limiting) Overview

The PFL (Power Force Limiting) function of the HC-series robots has 3 modes of operations:

- Transient Contact: Temporary contact caused by the robot hits something, or something hits the robot. Robot stops and the amber light button on the upper arm turns on. When the amber light button is press, the light turns off and the motion is resumes.

- Quasi-Static Contact: Similar to the transient contact but the contact is maintained, like hitting a table or a wall. After stopping, if an external force is maintained, the robot will move away from the exerted force to release the pressure and then stop again. The amber light button on the upper arm turns back on. When the amber light button is press, the light turns off and the motion is resumed.

- Avoidance: A low external force is applied to move the robot from its current trajectory. Once the force is removed, the robot will move back to the position where the external force was originally applied. Then the robot resumes its original path.

Note that some modes may be disabled on some systems.


### ROS Handling of PFL

The PFL function, being a safety feature, supersedes any motion from ROS.
The PFL normal behavior should be allowed to take place and ROS should not interfere with it by sending stop motion while the PFL function is active.
When ROS is moving the robot, since the motion path is coming from an external source and the position of the robot may have changed following the activation of PFL, MotoROS will cancel the current trajectory following the PFL activation.
Following the PFL activation any joint trajectory message sent to the controller will return the `ROS_RESULT_NOT_READY_PFL_ACTIVE` code `5011`.
The status bit `in_motion` will be `false` when the robot is not moving.
It may temporarily turn on in PFL's *Quasi-Static Contact* mode as the robot moves away to release pressure on the impact point.
The status bit `motion_possible` (ready) will be `false` and remain `false` until the `/robot_enable` service is called (Start trajectory mode message `ROS_MSG_MOTO_MOTION_CTRL` message `2001` with the `ROS_CMD_START_TRAJ_MODE` command `200121`).
When `motion_possible` is `true` again, the current robot position can be acquired by ROS and a new trajectory to resume motion can be planned and sent to the controller.


### Code Change:

Code change essentially consists of adding new I/O signals to capture the avoidance mode:

- `IO_ROBOTSTATUS_PFL_AVOIDING` (15120)
- `IO_ROBOTSTATUS_PFL_AVOID_JOINT` (15124)
- `IO_ROBOTSTATUS_PFL_AVOID_TRANS` (15125)

The `IO_ROBOTSTATUS_PFL_AVOIDING` signal indicates that a force above the avoidance threshold was detected.
The signals `IO_ROBOTSTATUS_PFL_AVOID_JOINT` and `IO_ROBOTSTATUS_PFL_AVOID_TRANS` only indicate that the avoidance mode is enabled.
So, the `IO_ROBOTSTATUS_PFL_AVOIDING` signal should only be considered if one of the other two signals is on.

And a new flag `bool bPFLduringRosMove` was added to the `Controller` class to track that PFL was activated.
Finally, logic to manage the detection and reset of the PFL flag was added.

CameronDevine:
@EricMarcil This is working quite well. I did notice a small issue, if you call the `/robot_enable` service while a PFL stop is active, the service returns false. However, if immediately after it returns you press the PFL resume button, the robot will be re-enabled without calling the service a second time.

I will continue to use this version and let you know if I find any other issues.

Also, for some applications it might be nice to provide the return value of the `Ros_Controller_IsPflActive` function to the client system. This may be beyond the scope of the PR as it would (I believe) require a new simple message to be defined.

EricMarcil:
@CameronDevine: Thanks for the feedback.  I'll try to see if I can improve on the small issue you mentionned but it might be difficult.  If I can't, I don't think it is a big problem.  The main objective still being that /robot-enable needed to be sent at least once and the previous motion can't be continued without replanning.

I agree that having separate signal for PFL would make things clearer, but currently we are using the standard ROBOT_STATUS message (13).  So I'm not sure if we could add to that message or if we would need change to a Motoman specific message.  What are the implication for current users if we make that change?
@gavanderhoorn: Any thoughts on this?

CameronDevine:
@EricMarcil I agree that it isn't a major issues; however I wanted to mention it just in case. I don't think I would have noticed the issue at all if I didn't have a second PFL resume button at my workstation, so it is unlikely anyone else will ever notice it.

As for the second issue, would it be possible for the MotoPlus application to detect if the robot is part of the HC series, and if so publish a Motoman specific simple_message including the PFL status? The motoman_driver could then subscribe to both the universal robot status message, as well as the Motoman specific version with PFL status. When a simple_message with PFL status is received, both the standard ROS industrial_msgs robot status message and a Motoman specific ROS message including PFL status could be published for backwards compatibility.
I don't have a good understanding of the simple_message and MotoRos systems, so I will absolutely defer design choices to @gavanderhoorn and others.

gavanderhoorn:
Just tested this on our YRC1000.

I can't identify any regressions, but I can't test the PFL side to this at all.

gavanderhoorn:
@EricMarcil: I'm going to update the commit comment of 177af0d to include the information you posted in https://github.com/ros-industrial/motoman/pull/382#issuecomment-755495018, as that seems like it is very valuable information on how PFL is integrated into MotoROS.

gavanderhoorn:
And @EricMarcil: I'd suggest we merge this update first, and discuss making the PFL status available via `motoman_driver` later.

motoman_driver/MotoPlus/Controller.c
```diff
@@ -538,7 +543,13 @@ BOOL Ros_Controller_IsMotionReady(Controller* controller)
 BOOL Ros_Controller_IsPflActive(Controller* controller)
 {
 #if (YRC1000||YRC1000u)
-	return ((controller->ioStatus[IO_ROBOTSTATUS_PFL_STOP] | controller->ioStatus[IO_ROBOTSTATUS_PFL_ESCAPE]) != 0);
+	//return ((controller->ioStatus[IO_ROBOTSTATUS_PFL_STOP] | controller->ioStatus[IO_ROBOTSTATUS_PFL_ESCAPE]) != 0);
+	if (controller->bPFLduringRosMove | controller->ioStatus[IO_ROBOTSTATUS_PFL_STOP] | controller->ioStatus[IO_ROBOTSTATUS_PFL_ESCAPE] |
+		( controller->ioStatus[IO_ROBOTSTATUS_PFL_AVOIDING]
+		&& (controller->ioStatus[IO_ROBOTSTATUS_PFL_AVOID_JOINT] | controller->ioStatus[IO_ROBOTSTATUS_PFL_AVOID_TRANS])) != 0)
```
gavanderhoorn: is this using binary OR on purpose (ie: `|`) or should it be `||` instead? It seems to mix booleans and bit manipulation.

In the end it probably all gets processed correctly by the compiler, but it would be good to be careful to use the correct operators.

motoman_driver/MotoPlus/MotionServer.c
```diff
@@ -1751,6 +1755,13 @@ void Ros_MotionServer_IncMoveLoopStart(Controller* controller) //<-- IP_CLK prio
 			{
 				if(ret == -3)
 					printf("mpExRcsIncrementMove returned: %d (ctrl_grp = %d)\r\n", ret, moveData.ctrl_grp);
+#if (YRC1000||YRC1000u)
+				else if (ret == -1 || ret == -19)
```
gavanderhoorn: I realise there is already one on line 1756, but could we avoid adding more magic numbers? If possible, please add named constants for `-1` and `-19`.

motoman_driver/MotoPlus/MotionServer.c
```diff
@@ -1751,6 +1755,13 @@ void Ros_MotionServer_IncMoveLoopStart(Controller* controller) //<-- IP_CLK prio
 			{
 				if(ret == -3)
 					printf("mpExRcsIncrementMove returned: %d (ctrl_grp = %d)\r\n", ret, moveData.ctrl_grp);
+#if (YRC1000||YRC1000u)
+				else if (ret == -1 || ret == -19)
+				{
+					printf("mpExRcsIncrementMove returned: %d (TFS Active)\r\n", ret, moveData.ctrl_grp);
```
gavanderhoorn: Is TFS another name for PFL?

gavanderhoorn (changes requested):
A few minor comments in-line.

motoman_driver/MotoPlus/MotionServer.c
EricMarcil: No, that's for the source control in Visual Studio.  I got my accronyms mixted up.  I'll make the correction.

motoman_driver/MotoPlus/MotionServer.c
EricMarcil: I'll clarify.

motoman_driver/MotoPlus/Controller.c
EricMarcil: I'll make the correction

motoman_driver/MotoPlus/MotionServer.c
gavanderhoorn: `-1` appears to be `E_EXRCS_IMOV_UNREADY`.

The YRC1000 API Spec I have does not list `-19`?

motoman_driver/MotoPlus/MotionServer.c
gavanderhoorn: I've pushed a commit which fixes this.

EricMarcil: @gavanderhoorn 
Thanks for the update.  I had made the changes locally but I was waiting for confirmation on the -19 return value which is actually undocumented. 
-3 is E_EXRCS_CTRL_GRP
-1 is E_EXRCS_IMOV_UNREADY
-19 is I believe is PFL related but I don't know exactly in which context it occurs.  Normally I see -1.  So, I'm trying to get the confirmation from Japan and an update on the header files that should define it.

gavanderhoorn: > I had made the changes locally but I was waiting for confirmation

yes, I should probably have done a PR against your branch.

Apologies.

gavanderhoorn:
@EricMarcil: have you heard anything from Japan already about the missing constant declaration?

EricMarcil:
@gavanderhoorn: Sorry about the delay.  I've restested and the -19 return doesn't seem to be caused by PFL so I removed it.  We should be good to merge and close this.

motoman_driver/MotoPlus/Controller.c
EricMarcil: I've pushed the correction in 1fd459a and b7ed970

motoman_driver/MotoPlus/MotionServer.c
EricMarcil: I've removed the -19 of the condition.  It is not a valid return value.  I'm must have accidently mixted it up with some other function return value.

EricMarcil (commented):
I think we are good to merge

gavanderhoorn (approved):
Approving based on my own testing (no regressions with a YRC1000) and the comments by @CameronDevine and @EricMarcil.

gavanderhoorn:
Thanks @EricMarcil for fixing this.

Thanks @CameronDevine for testing it.

gavanderhoorn:
Updated the wiki: [diff](http://wiki.ros.org/action/info/motoman_driver/Tutorials/indigo/InstallServer?action=diff&rev2=62&rev1=61).

CameronDevine:
> And @EricMarcil: I'd suggest we merge this update first, and discuss making the PFL status available via `motoman_driver` later.

@gavanderhoorn and @EricMarcil thanks for all your work on this. Should a new issue be opened about making the PFL status available in ROS?

gavanderhoorn:
Yes, that would be good.

PR #389, opened 2021-02-17
rebased 319: serialise SmplMsg connection access in MJTS

gavanderhoorn:
As per subject.

Commit provenance retained.

Rebased to run with new CI, and added an additional commit which uses C++11 `mutex` instead of Boost.

### DIFF
--- a/motoman_driver/src/joint_trajectory_streamer.cpp
+++ b/motoman_driver/src/joint_trajectory_streamer.cpp
@@ -119,8 +119,10 @@ bool MotomanJointTrajectoryStreamer::init(SmplMsgConnection* connection, const s
 
 MotomanJointTrajectoryStreamer::~MotomanJointTrajectoryStreamer()
 {
+  // SmplMsgConnection is not thread safe, so lock first
+  // NOTE: motion_ctrl_ uses the SmplMsgConnection here
+  const std::lock_guard<std::mutex> lock{smpl_msg_conx_mutex_};
   // TODO( ): Find better place to call StopTrajMode
-  boost::mutex::scoped_lock lock(connection_mutex_);
   motion_ctrl_.setTrajMode(false);   // release TrajMode, so INFORM jobs can run
 }
 
@@ -129,12 +131,12 @@ bool MotomanJointTrajectoryStreamer::disableRobotCB(std_srvs::Trigger::Request &
 {
   trajectoryStop();
 
-  bool motion_ctrl_result = false;
   {
-    boost::mutex::scoped_lock lock(connection_mutex_);
-    motion_ctrl_result = motion_ctrl_.setTrajMode(false);
+    // SmplMsgConnection is not thread safe, so lock first
+    // NOTE: motion_ctrl_ uses the SmplMsgConnection here
+    const std::lock_guard<std::mutex> lock{smpl_msg_conx_mutex_};
+    res.success = motion_ctrl_.setTrajMode(false);
   }
-  res.success = motion_ctrl_result;
 
   if (!res.success)
   {
@@ -152,12 +154,12 @@ bool MotomanJointTrajectoryStreamer::disableRobotCB(std_srvs::Trigger::Request &
 
 bool MotomanJointTrajectoryStreamer::enableRobotCB(std_srvs::Trigger::Request& req, std_srvs::Trigger::Response& res)
 {
-  bool motion_ctrl_result = false;
   {
-    boost::mutex::scoped_lock lock(connection_mutex_);
-    motion_ctrl_result = motion_ctrl_.setTrajMode(true);
+    // SmplMsgConnection is not thread safe, so lock first
+    // NOTE: motion_ctrl_ uses the SmplMsgConnection here
+    const std::lock_guard<std::mutex> lock{smpl_msg_conx_mutex_};
+    res.success = motion_ctrl_.setTrajMode(true);
   }
-  res.success = motion_ctrl_result;
 
   if (!res.success)
   {
@@ -380,7 +382,9 @@ bool MotomanJointTrajectoryStreamer::send_to_robot(const std::vector<SimpleMessa
 {
   bool motion_ctrl_result = false;
   {
-    boost::mutex::scoped_lock lock(connection_mutex_);
+    // SmplMsgConnection is not thread safe, so lock first
+    // NOTE: motion_ctrl_ uses the SmplMsgConnection here
+    const std::lock_guard<std::mutex> lock{smpl_msg_conx_mutex_};
     motion_ctrl_result = motion_ctrl_.controllerReady();
   }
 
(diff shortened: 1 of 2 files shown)


gavanderhoorn:
I've tested this with our YRC1000 and wasn't able to break it, but seeing as this has the potential potential (yes, twice) for deadlocks I'd like a second pair of eyes and some tests.

@marip8 @steviedale: would you have a chance of testing this?

gavanderhoorn:
Note: the reason this is necessary is because the MJTS starts a separate thread to send trajectories to the controller. This thread uses the same `SmplMsgConnection` object as "the rest" of the node.

The node itself is single-threaded (simple `ros::spin()`), but because of the extra thread interacting with a shared resource, locking is necessary.

gavanderhoorn:
An alternative solution could be to implement a (private) getter for the connection object. This could then protect it with a mutex. All users of the connection object would have to use the getter, instead of a bare pointer.

That would require additional changes in all users of the connection object though.

Another alternative would be to make `simple_message` thread-safe itself, but that would be rather involved.

gavanderhoorn:
@steviedale @marip8: could you take a look?

riv-rbush:
Tested this on our [local branch](https://github.com/additiveautomations/motoman/commits/point-streamer-hotfix). Haven't encountered any issues over the past few days. Some of the changes we've made compared to `kinetic-devel` branch:
- `CONTROLLER_STATUS_UPDATE_PERIOD = 4`
- Point Streamer
- IO server is turned OFF
- `MotoROS` and `motoman_driver` `sleep`s have been reduced

Is there anything specific you want us to test?

gavanderhoorn:
This has been tested on real HW by @rr-robert for some time now, and I've personally also run this locally.

I'd like to merge this before merging #435, #436 and #439, as it makes sure we avoid concurrent write problems with the SimpleMessage connection.

motoman_driver/src/joint_trajectory_streamer.cpp
```diff
@@ -380,6 +399,8 @@ bool MotomanJointTrajectoryStreamer::send_to_robot(const std::vector<SimpleMessa
 void MotomanJointTrajectoryStreamer::streamingThread()
 {
   int connectRetryCount = 1;
+  bool is_connected = false;
+  bool is_msg_sent = false;
```
marip8, line 403: Personal preference, but I think it would improve readability to declare these where they are used below

motoman_driver/src/joint_trajectory_streamer.cpp
```diff
@@ -390,11 +411,25 @@ void MotomanJointTrajectoryStreamer::streamingThread()
...
+        const std::lock_guard<std::mutex> lock{smpl_msg_conx_mutex_};
+        this->connection_->makeConnect();
+      }
       ros::Duration(0.250).sleep();  // wait for connection
 
-      if (this->connection_->isConnected())
+      is_connected = false;
+      {
+        // SmplMsgConnection is not thread safe, so lock first
+        // TODO(gavanderhoorn): not sure this needs to be protected by a mutex
```
marip8, line 424: Theoretically it's a good idea to keep this around since it appears that the [`SimpleSocket` implementation of `isConnected`](https://github.com/ros-industrial/industrial_core/blob/9bc930ce9be0f828f550047767343ec4b5fb88ff/simple_message/include/simple_message/socket/simple_socket.h#L168-L171) does not use an atomic bool for the connection status.

motoman_driver/src/joint_trajectory_streamer.cpp
```diff
@@ -403,6 +438,7 @@ void MotomanJointTrajectoryStreamer::streamingThread()
       continue;
     }
 
+    // this does not lock smpl_msg_conx_mutex_, but the mutex from JointTrajectoryStreamer
     this->mutex_.lock();
```
marip8, line 442: Is it worth switching this underlying mutex to a `std::mutex` and using the same lock semantics introduced in this PR for consistency?

marip8 (dismissed):
It looks like you captured all of the places where `connection_` and `motion_ctrl_` were being used and it sounds like it's been tested on hardware sufficiently, so I think this is good to go

gavanderhoorn:
Apologies @marip8, I appear to have force-pushed just as you submitted your review.

Your comments are still relevant though, so I'll respond to them.

motoman_driver/src/joint_trajectory_streamer.cpp
gavanderhoorn: It could be nice, but that would require changes to `JointTrajectoryStreamer` in `industrial_robot_client` ([here](https://github.com/ros-industrial/industrial_core/blob/ea3d805e1271550be312409d6474177e67796e15/industrial_robot_client/include/industrial_robot_client/joint_trajectory_streamer.h#L120)), it's not part of `motoman_driver`.

gavanderhoorn: To clarify: I'm not against updating `industrial_robot_client`, but that would need additional testing, further delaying this PR.

gavanderhoorn:
An alternative implementation could be to create either a wrapper class or a subclass of `SmplMsgConnection` and make that wrapper/subclass thread-safe.

motoman_driver/src/joint_trajectory_streamer.cpp
marip8: Okay; I thought it might have been limited to the `motoman_driver/industrial_robot_client` directory. We can address this in a separate PR as necessary then

motoman_driver/src/joint_trajectory_streamer.cpp
gavanderhoorn: I'd like to do that, but I believe we can't. At least `is_connected` needs to be in the scope which contains the `switch (this->state_)`. So it can be moved to be inside the outer `while (ros::ok())`, but that doesn't really change things I believe.

marip8:
I was thinking about that earlier. I don't know much about the existing `SimpleSocket` implementation or socket messaging in general, but would it be as simple as using atomic variables and putting a mutex in the `sendBytes` and `receiveBytes` functions (or around where they're called in the `SmplMessageConnection` base class)? I can't imagine it would be that easy, but I don't know what else it would take

gavanderhoorn:
We'd have to make sure to protect access to the underlying socket, yes. That'd basically be it.

It's been a while since I've looked at the implementation though, so there could be some gotchas hidden there.

It would be more fundamental than what we're doing here though. So without testing it myself I wouldn't feel comfortable making that change and just merging it.

---

Edit: I wanted to get this in as it's a user-contributed change that functionally achieves the goal: prevent concurrent writes to the underlying `SmplMsgConnection`.

---

Edit 2: actually, it's probably mostly needed in `sendAndReceive(..)`. The issue is that without sequence nr tracking and book keeping of outstanding responses to match them against requests, that method will just assume that whatever message comes in is the response to what it sent out. Which of course doesn't have to be the case, but is an OK assumption if message interleaving is impossible due to serialised access / single thread of control. Now that we are breaking that assumption (by allowing ROS services to be called while the connection is being used for other things), we run into problems.

gavanderhoorn:
@marip8: my force-push dismissed your review. If you are still OK with the changes proposed, could you resubmit your review?

marip8 (approved):

gavanderhoorn:
Thank you @marip8 for the review, and thank you @rr-robert for testing this out.

And finally thank you @yszheda for reporting the original problem and providing the initial fix (which I based my modifications on).

PR #266, opened 2019-02-19
Add ma2010, motopos d500 support and ma2010 moveit_config

tdl-ua:
Added support packages for both MA2010 manipulator and D500 positioner, and a movit_config package for the MA2010 manipulator. 

Motopos D500 support package is called `motoman_motopos_d500_support`. Is this correct or should it be renamed to `motopos_d500_support`? Am assuming that only manipulators are called "motoman".

Testing:
* both ma2010 and d500 support packages have been tested with `test_[robot/positioner model].launch`: all joints move as intended
* `motoman_ma2010_moveit_config`: 
  * `moveit_planning_execution.launch sim:=true`: planning and execution
  working
  * `moveit_planning_execution.launch` has not been tested on a physical robot

PR #391, opened 2021-02-18
rebased 266: MA2010 and D500 support

gavanderhoorn:
Rebased version of #266.

I did not want to push commits directly to the `Adamsua-lab/motoman@kinetic-devel`, as I don't know whether that branch is actively being used or not.

Commit provenance is retained.

We'll let CI run, and afterwards I'll squash my fixup commits into the three initial commits by @tdl-ua, so as to not pollute the commit history.

I'm still not sure about the MoveIt configuration (it duplicates files and information from the other packages, among other things), but we'll merge it, as it may help other users.


gavanderhoorn:
Thanks again @tdl-ua.

If you're using your `kinetic-devel` right now, pulling from `upstream` here may result in conflicts (as you already have the packages these commits add), but I did not see a way to avoid that without overwriting your `kinetic-devel` (and I did not want to do that).

gavanderhoorn:
Note: I've corrected the `effort` limits in the `motoman_ma2010_support` package you contributed.

@EricMarcil provided me with the correct limits.

tdl-ua:
@gavanderhoorn Thank you for merging and thanks for the considerations. For future consideration, I'm assuming it would be better to not create a PR based on the default branch (e.g., `kinetic-devel`) but to create a dedicated branch for PRs?

gavanderhoorn:
> For future consideration, I'm assuming it would be better to not create a PR based on the default branch (e.g., `kinetic-devel`) but to create a dedicated branch for PRs?

yes, indeed.

that way we don't run the risk of breaking things for you when we push some fixup commits to your PR branch.

PR #395, opened 2021-03-04
Add gp180 support

Doomerdinger:
Added support for GP180.

See #375

Requires #378.


gavanderhoorn:
@Doomerdinger: it appears this model has been constructed such that all `joint`s rotate around the Z axis.

I assume that's because Path Robotics is using a calibration procedure which is based on D-H?

Unfortunately we don't use that in ROS, and I'd really rather see we keep the models compliant with REP 103 (and all other support packages in our repositories), which would mean X+ 'forward', Y+ 'left', Z+ 'up'.

If that wouldn't be something you'd want to do, I'll change the `joint`s myself, but I at least wanted to ask you.

gavanderhoorn:
(note: the conflicts are because I merged the rebase of #378 in #405; simplest fix is to `rebase` and only keep 7d08b75)

Doomerdinger:
Yes we use a calibration procedure based on DH.

I based on the wording in REP103 I did not think that standard necessarily applied to joints, and was more for things like base links or flange, transforms with respect to some body. I could've sworn I saw some examples in this repository with joint rotating around the Z.

Regardless, if this is not possible by default do you think there's some way this could be customized? I'd prefer to not have a fork for every robot we create/use, it will make it harder to make contributions back.

Whatever the case I can switch them easily, but if you're already on it be my guest. I won't get to it until sometime next week.

Doomerdinger:
@gavanderhoorn Expanding on my last comment after looking at some robot models.

It seems I may have misunderstood the meaning of REP-103. Based on the way the robots in other support packages are configured, it seems that that axis are oriented in such a way that "x forward" is always with respect to the robot base.

I had assumed that "x forward" applied to whatever body that the transform represented. With that interpretation the "forward" direction for each part of the robot is a bit more ambiguous. In my models the X axis points "forward" and Y "left" with Z "up" in accordance with the right hand rule with respect to the axis of orientation of the joint. So positive rotation of the joint is defined by the rotation from positive/forward X, to positive/forward Y.

I was actually a bit confused as to why the support packages in this project didn't use the Z axis to rotate around, since it seemed to comply with REP-103 as well as allow for usage of things like DH.

Is there a systematic way determine which "body" a particular transform is relative to so that "x forward" can be set correctly?

Doomerdinger:
@gavanderhoorn if you have time, you mentioned you could do what is necessary to bring this up to snuff to get it into the main branch?

I've had slim to no time to work on this stuff in quite some time.

gavanderhoorn:
I'll try to get to it. Thanks for letting us know @Doomerdinger.

gavanderhoorn:
Replacement: #479.

Thanks for the initial work @Doomerdinger :+1:

gavanderhoorn:
And apologies for taking so long to get this merged in.

PR #397, opened 2021-03-05
Add motopos mh1655 support

Doomerdinger:
Added support for the mh1655 positioner.

Also included is the option to put the positioner up on a pedestal (one of the official motoman ones) - this functionality is disabled by default. I can remove this if desired.

Requires #378.


gavanderhoorn:
Closing this in favour of #621.

Thanks @Doomerdinger for the initial PR :+1: 

And apologies (again) for letting this sit here for so long :disappointed:

PR #402, opened 2021-04-19
MotoROS v1.9.4

ted-miller:
Fix lockup if network connection was broken without properly disconnecting.
Update Visual Studio project file to be compatible with VS2019.


ted-miller:
This update has been tested by @EricMarcil

gavanderhoorn:
High-level comment: according to what I can find, `TCP_KEEPCNT` et al. should be supported on the VxWorks versions MotoROS targets. That should make it possible to reduce the "2 hours" idle timeout and configure it on a per `fd` level.

Using TCP idle detection/heartbeats could make the use of `PING` redundant and reduce the complexity of the implementation.

Was that considered?

---

Edit: additionally, could you add a short description of the approach taken to implement lost connections? `SO_KEEPALIVE` is enabled, but there also appear to be structural/control flow changes in addition to the use of the `PING` message. It's not entirely clear to me how those all interfact (if they do).

ted-miller:
The mpSetSockOpt api does not support additional configuration options. Also, I dug through the vxworks header files and couldn't find the applicable options in there either. So, I'm curious what documentation you found.

The code to send a ping command was actually removed. I left the code to receive/handle a ping though. I've seen an instance where someone was using it and they got a bunch of unnecessary error messages.

Basically, all this new code does is allow a new connection to replace an old connection. If a new one comes in, then the old one is dropped. In the event that motion is in process, then I assume that the old connection is still in control and I reject the new connection.

gavanderhoorn:
I won't have time to test this on real hw until Friday.

In the meantime, I'm going to accept the PR in preparation of merging it.

It could be that using `TCP_KEEPCNT` et al. is possible on VxWorks, but that could be added later.

gavanderhoorn (approved):
Approving provisionally. Real HW test on Friday (in addition to tests by @EricMarcil).

gavanderhoorn:
I've not been able to test this today as our YRC1000 has developed some problems with the external axis, blocking all use of the robot.

Based on @EricMarcil's tests we'll merge this.

Thanks @ted-miller for the PR and addressing the reconnection issue.

gavanderhoorn:
Updated the wiki: [diff](http://wiki.ros.org/action/info/motoman_driver/Tutorials/indigo/InstallServer?action=diff&rev2=63&rev1=62).

Issue #410, opened 2021-05-27
Error 5011 after EStop (ROS_RESULT_NOT_READY_PFL_ACTIVE)


ted-miller:
@EricMarcil: Can you look at this issue? I don't understand all the signals you're monitoring in Ros_Controller_IsPflActive.

Steps to reproduce:
- Start a motion trajectory
- EStop during motion
- Release the EStop
- Send Start-Trajectory-Mode
- Check Motion Ready returns 5011

The only way to clear the 5011 is to explicitly send a Stop-Trajectory-Mode message and then try again.
Note that this was performed on a GP-12 that does not have FSU. So... there's definitely nothing going on with PFL.

EricMarcil:
@ted-miller: Do you have anything on your system interacting with outputs 15120, 15124 or 15125 on that system?  Those are outputs used by the Avoidance function that affect that logic.
We should probably figure out a way to identify if the robot is a HC robot or not, and put an overall condition around all the PFL related code.  I'll try to look into it tomorrow or early next week.

ted-miller:
> Do you have anything on your system interacting with outputs 15120, 15124 or 15125 on that system?

Not that I'm aware of. I can double check later.

> We should probably figure out a way to identify if the robot is a HC robot or not

In UWI, I look at the PFL function parameter. I could bake that into the parameter extraction library.

EricMarcil:
I've confirmed the issue.  When the PFL activates, the mpExRcsIncrementMove returns E_EXRCS_IMOV_UNREADY and I had put a flag to mark that state because I had found a mode where other signals couldn't be relied on.  It logs:
    mpExRcsIncrementMove returned UNREADY: %d (Could be PFL Active)
But the E-Stop causes the same state (and maybe the hold, I'll check).
I'll add a check for E-Stop and Hold to distinguish these cases.  I'll also try to add a check to see if the PFL is even active on the system or not.

EricMarcil:
The should be corrected by PR #412

gavanderhoorn:
#412 should fix this.

PR #412, opened 2021-06-04
MotoRos v1.9.5

EricMarcil:
- Fixed message reporting PFL being enabled after E-Stop.
- Added check on startup to identify if PFL is enabled, and remove PFL related checks if disabled.
- Added MotionStop if controller is unable to process incremental motion to prevent inaccurate incomplete motion.
- Improved handling of waking robot from energy saving mode (Eco Mode) in relation with improvement  made on controller software YAS4.13 (YRC1000) or later and YBS2.43 (YRC1000micro) or later.


gavanderhoorn:
Tested on our YRC1000 (`YAS4.24`) -- with an FSU, but no PFL.

Sequence described by @ted-miller in #410 is no longer problematic (ie: traj mode can be enabled).

No others problems encountered, but I'll have to rely on someone else to test the PFL side to this.

gavanderhoorn:
For future PRs: would it be possible to use separate commits instead of a single one? That would make reviews a little easier.

EricMarcil:
@gavanderhoorn Thanks for the testing this.  I tested on my HC10 with PFL.  I've also test on YRC1000 and DX200 without PFL.
Sorry for the single commit. It started as a single issue and I was doing testing, I kept finding other small issues to fix.  I'll try to split them in the future.

gavanderhoorn (approved):

motoman_driver/MotoPlus/SimpleMessage.h
```diff
@@ -159,7 +159,8 @@ typedef enum
 	ROS_RESULT_NOT_READY_NOT_STARTED,
 	ROS_RESULT_NOT_READY_WAITING_ROS,
 	ROS_RESULT_NOT_READY_SKILLSEND,
-	ROS_RESULT_NOT_READY_PFL_ACTIVE
+	ROS_RESULT_NOT_READY_PFL_ACTIVE,
+	ROS_RESULT_NOT_READY_INC_MOVE_ERROR
```
gavanderhoorn, line 163: We need to update the `NotReadyCode` enum:

https://github.com/ros-industrial/motoman/blob/9ade90acfe5835ca38fddcf6ac9cf589a452f449/motoman_driver/include/motoman_driver/simple_message/motoman_motion_reply.h#L100-L116

gavanderhoorn: We'll do that in a separate PR.

gavanderhoorn:
@EricMarcil: will you update the wiki page (ie: the links to the `.out` files and the versions)?

gavanderhoorn:
Thanks for the update @EricMarcil :+1:

EricMarcil:
@gavanderhoorn: Yes, we'll update wiki page.  I've requested edit access.  If I still don't have access on Monday, I'll ask Ted (He usually the one that updates the wiki...  But I should probably get access to do it for my changes.)

gavanderhoorn:
This is the page that needs to be updated: [motoman_driver/Tutorials/indigo/InstallServer](http://wiki.ros.org/motoman_driver/Tutorials/indigo/InstallServer). The last update changed this: [diff](http://wiki.ros.org/action/info/motoman_driver/Tutorials/indigo/InstallServer?action=diff&rev2=63&rev1=62).

That includes the version nrs, links to the binaries and the `md5sum`s.

EricMarcil:
Wiki was updated

gavanderhoorn:
thank you

EricMarcil:
I just realized that I forgot to change the version number in the MotoPlus code (in MotoROS.h) but I don't think it actually displays anywhere.

gavanderhoorn:
Hm. It should actually. IIRC, @ted-miller added a bit of code which registers MotoROS with the system to show versioning info on the TP *somewhere*.

I'll try to find it.

gavanderhoorn:
Here:

https://github.com/ros-industrial/motoman/blob/0786691078e75e5b3a5b17543ee6013fdf40d568/motoman_driver/MotoPlus/Controller.c#L77-L93

ted-miller:
Yeah, the version number displays on the pendant and is also exported to SYSTEM.SYS and PANELBOX.LOG when saving a robot backup. It's helpful for troubleshooting to know the exact version the user is using.

gavanderhoorn:
@EricMarcil: if you could submit a fixup-PR with the version nr updated and new binaries, I'll review it quickly.

EricMarcil:
@gavanderhoorn I just pushed PR #413 to fix it.

motoman_driver/MotoPlus/SimpleMessage.h
gavanderhoorn: See #438.

PR #413, opened 2021-06-14
Version reference fixed to 1.9.5

EricMarcil:
Fixed version reference that was forgotten in the PR #412


gavanderhoorn:
I realise FS100 and DX100 don't show these versions anywhere, but it would be good to include the complete set of updated binaries in this PR.

That'll make it that much easier to reason about what changed in the future, and easier to match binaries to Git hashes.

EricMarcil:
Yes, I saw that the FS100 and DX100 binary were not included but I actually compiled them.  GitHub won't pick them up because the resulting compiled binary doesn't actually change.

gavanderhoorn:
> GitHub won't pick them up because the resulting compiled binary doesn't actually change.

Ok, thanks for confirming that.

gavanderhoorn (approved):
I can't test this on real hw, but this rectifies #412 and is the same pattern used in earlier PRs which update MotoROS version numbers.

gavanderhoorn:
Unfortunately this will mean you'll have to update the wiki again @EricMarcil. At least for the affected binaries.

EricMarcil:
@gavanderhoorn I expected that.  It's done and I've confirmed that the downloaded .out was matching the MD5 hash.
Thanks for the quick turn around.

PR #415, opened 2021-07-11
Added variant model HC10DT-B10

EricMarcil:
Added variant HC10DT-B10 to the HC10 Support package.
Also made some modification to existing HC10DT variant to optimize reuse of HC10 base model files.


EricMarcil:
@gavanderhoorn Can you please review.  

I've tried to optimized the reused of the `joint_names_hc10.yaml` and `robot_interface_streaming_hc10.launch` for the variant model.

Also in the `roslaunch_test_hc10dt.xml` and `roslaunch_test_hc10dt_b10.xml`, should the `<group ns=>` make reference to the generic model:

```xml
<group ns="load_hc10">
```

or the variant model:

```xml
<group ns="load_hc10dt">
```

I've also acquired a new model convertion software that will allow me to directly convert MotoSim models to dae mesh.  The Motosim model are already optimized for size and contain different textures with color information which gives nicer models.

gavanderhoorn:
Closing/re-opening to trigger new CI run.

gavanderhoorn:
I've opened an alternative version of this: #434.

@EricMarcil: could you please check that one and if you agree close this one?

EricMarcil:
@gavanderhoorn I was trying to avoid duplicating all the files that have the exact same content.
But if you think it is better to duplicate everything with the DT_B10 at the end then that's fine.

gavanderhoorn:
> I was trying to avoid duplicating all the files that have the exact same content. 

Yes, I understood.

> But if you think it is better to duplicate everything with the DT_B10 at the end then that's fine.

deduplication (or prevention of duplication) is typically a good idea. However, there is quite some cognitive load already when trying to figure out which support package to use, and then which files in that package to use for new users.

Without reading the contents of the files, the filename is the "only" information users have to figure out which to use for their specific variant. If we then don't include certain files for certain variants, but do include other files (such as meshes and some `.launch` files), it gets confusing really quickly (is support for a variant incomplete? Were the files not included on purpose? The name of the file states it's for the HC10, can I also use it for my HC10DT? etc).

Coupled with the fact these files are basically small, simple text files, the cost of duplication isn't high enough to be a real problem.

gavanderhoorn:
Closing this as we're continuing in #434

PR #417, opened 2021-07-24
Fix memset call to proper size

lights0123:
`sizeof(torqueValues)` is 4 (because `double torqueValues[MAX_PULSE_AXES]` as a parameter in C is equivalent to `double *torqueValues`, so the size of the pointer), while we actually want to clear the data behind the pointer. Modern compilers warn:

> Sizeof on array function parameter will return size of `double *` instead of `double [8]` (-Wsizeof-array-argument)

This does prompt a question on whether this call should exist at all, because no problems occurred when it was incorrectly setting data. Just to be safe, I'm leaving it in.

Incidentally, this function is never called, and can probably be removed entirely.

### DIFF
--- a/motoman_driver/MotoPlus/CtrlGroup.c
+++ b/motoman_driver/MotoPlus/CtrlGroup.c
@@ -454,7 +454,7 @@ BOOL Ros_CtrlGroup_GetTorque(CtrlGroup* ctrlGroup, double torqueValues[MAX_PULSE
   	LONG status = 0;
   	int i;
 
-	memset(torqueValues, 0, sizeof(torqueValues)); // clear result, in case of error
+	memset(torqueValues, 0, sizeof(double [MAX_PULSE_AXES])); // clear result, in case of error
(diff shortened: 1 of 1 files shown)


EricMarcil (approved):
Good catch. 
Since that function is not currently used in the standard driver, we don't need to release a new version but the change can be merge for future release.
Thanks

gavanderhoorn:
>  we don't need to release a new version but the change can be merge for future release.

I would actually suggest we do increase the version.

It makes debugging much easier.

gavanderhoorn (approved):

gavanderhoorn:
I've asked @ted-miller to bump and build `1.9.7` in https://github.com/ros-industrial/motoman/pull/422#issuecomment-911517080.

gavanderhoorn:
> Modern compilers warn

yes, they do.

But MotoPlus does not include a recent compiler at this time.

gavanderhoorn:
We'll also need to update the wiki again.

ted-miller:
#423

PR #422, opened 2021-09-01
Fix alarm code in robot status

ted-miller:
If there is an alarm of any type other than "user alarm", then the `alarmCode` will be zeroed out. 
This fixes the `alarmCode` so that it is populated correctly.


ted-miller:
Awaiting test results from the person who reported the problem.

EricMarcil (approved):

AGregg:
Just to record what we did to test.  We had been triggering the shock sensor on a robot while monitoring the robot status ros topic.  Before installing this update on the robot controller, the alarm code would not populate in the error_code field of the topic message, but after installing the updated driver the alarm code came through.

PXL_20210901_185549419~2

AGregg (approved):

gavanderhoorn (approved):
No regressions from my testing (YRC1000).

gavanderhoorn:
@ted-miller: if we merge #417, could you build `1.9.7`?

Issue #425, opened 2021-09-15
Stopping robot during planned moves fails occasionally


riv-dmason:
I have been working on a problem where the robot needs to stop and then plan a different move based on the status of a sensor.

Expected sequence:

- robot starts at point A
- robot moves to point B (planned path, then executed)
- Sensor triggers
- Robot stop issued (before getting to B)
- robot moves to point A (planned path, then executed)
- sensor trigger
- robot stop issued (before getting to A)
- robot moves to point A (planned path, then executed)
- robot gets to point A

my python code commanding this sequence completes as expected however when running the sequence for 10 cycles in a row for example it will trigger an error (see below) on some of the stop commands but not all 20 (2 stops per cycle). The robot stops at the point expected (visually checked, over 1cm before the target position) when the sensor is triggered. The next move is also planned fine and the whole sequence continues regardless of the error.

![image](https://user-images.githubusercontent.com/73720167/133431101-85bbbb37-10cd-4648-9157-f5936bc4764c.png)

I am using the code below to achieve the stops. `moveit` is our own python module for interacting with moveit. Within that is a standard `movegroup` which is defined like this `move_group = moveit_commander.MoveGroupCommander(self.planning_group, ns=f"/{self.ns}")`

```
self.moveit.move_group.stop()
self.moveit.move_group.set_start_state_to_current_state()
self.moveit.move_group.clear_pose_targets()
time.sleep(0.5)
```

Here is a Wireshark dump of the communication and the package numbers of interest;

- 61040 (control message that worked)
- 61563 (reply saying it worked)
- 67677 (control message that failed)
- 67736 (reply saying it failed)

[wiresharkdump.zip](https://github.com/ros-industrial/motoman/files/7169917/wiresharkdump.zip)

gavanderhoorn:
Could you locally merge #389 and see whether that prevents the error?

riv-dmason:
We will give it a go today

riv-dmason:
@gavanderhoorn we merged that PR branch into our local version and tested it this afternoon and were able to replicate the same issue. It appeared within 10 of the cycles described above.

Can I ask if we are best continuing with that PR merged in anyway/wondering why it has not been merged yet?

EricMarcil:
@David-Mason-1992 When this occurs and you get the error, does the motion still stops?  If it continues, do it reach its fully reach its target position?  
[...]

PR #426, opened 2021-09-16
Fix StopMotion delayed reaction when queue is full

EricMarcil:
Added condition to stop AddPulseIncPointToQ loop when StopMotion is requested.


riv-rbush:
Tested this on our [local branch](https://github.com/additiveautomations/motoman/commits/point-streamer-hotfix). It fixed issue #425.

ted-miller:
@EricMarcil : wouldn't it make more sense to put `controller->bStopMotion` inside of `Ros_Controller_IsMotionReady`? This way the flag gets checked in other locations that look for motion-ready.

ted-miller:
We need to figure something out for version numbering. Using the current scheme, we've only got one more update before v2.0.0. I think this would be confusing since all of the ROS 2 stuff seems to be named with a 2. 

Just something to think about moving forward. Maybe we just use more than one digit, like 1.9.11.

gavanderhoorn:
After `9` comes `10`, not `0`. So we have plenty of minor nrs to go.

EricMarcil:
@ted-miller Yes, it does make more sense.  I've made the changes.  If you could review it again.

@robertjbush If you have a chance, please grab the new binary and update your driver.  Let us know that the issue is still fixed with this new version.

ted-miller (approved):

riv-rbush:
@EricMarcil I will test your latest changes this week.

gavanderhoorn:
@robertjbush: have you had a chance to test this?

riv-rbush:
@EricMarcil @gavanderhoorn Tested this on our Yaskawa Motoman GP8 with the commit 2f46450 and 19f459d cherry picked to our local branch. Test was a success, no error messages appeared on the console.

gavanderhoorn:
thanks for reporting.

PS: the changes are on the MotoROS side, so I'm not exactly sure why you cherry-picked those commits (unless you are building your own MotoROS binaries).

EricMarcil:
@robertjbush Thanks for the final check. 
I've squashed and merged the changes.  I'll update the binary on the wiki shortly.

PR #432, opened 2021-11-01
Add opw parameters for MH12

simonschmeisser:
This adds OPW parameters for the OPW inverse kinematics solver as used by https://github.com/JeroenDM/moveit_opw_kinematics_plugin and others

Parameters were verified against the URDF using the self-verification mechanism of above plugin


gavanderhoorn:
Has this been used with HW as well?

"Verifying against the URDF" seems like we're setting up a cycle of sorts ;)

simonschmeisser:
No, this has not been tested against hardware and yes, the data is definitely redundant. But at least I confirmed it's consistent.

gavanderhoorn (approved):

gavanderhoorn:
Thanks @simonschmeisser :+1:

Issue #374, opened 2020-12-05
Robots should have flange


Doomerdinger:
Each robot should have flange, and tool0 should be a child of flange not the last robot link.
See [REP-0199](http://gavanderhoorn.github.io/rep/rep-0199.html).

Related: #21

gavanderhoorn:
I agree with this, but note: the REP you mention is not an official REP.

So we can decide we want to adhere to it here, but we can't "fault" the current models for not having those frames.

Doomerdinger:
Yup, I'm already throwing something together for it and since things like FANUC already do it (thanks to you I believe) I figured I'd open something up.

PR #433, opened 2021-11-02
GP4 Support Package

EricMarcil:
Adding support package for the Yaskawa Motoman GP4 (1-06vx4-A0*) model


gavanderhoorn:
Could I ask you to please not use `.dae` for collision meshes?

It may be superstition almost nowadays, but we've had some pretty bad problems with certain motion planning frameworks not liking Collada and crashing big robots into walls and other immovable objects ..

Please convert them to `.stl`. That should avoid any potential problems.

EricMarcil:
@gavanderhoorn I've changed te models to stl as per your request. But it would be nice to validate if dae are truly a problem,  I find the dae are nicer because the can keep my original model colors which allow mixted colors on some of the links and our Yaskawa label on the model.
Actually, as I'm reading this, I'm realizing that you mentionned only the collision model.  Would it be fine to keep dae for the visual model?

gavanderhoorn:
My apologies, I may not have been clear.

Only the `collision` meshes needed to be converted.

The `visual` meshes can be as elaborate and colourful as you'd like :)

---

Edit:

> Actually, as I'm reading this, I'm realizing that you mentionned only the collision model. Would it be fine to keep dae for the visual model?

just saw your edit.

So: yes.

EricMarcil:
@gavanderhoorn My bad for not reading your e-mail more closely.  I just pushed a version to make everyone happy.  Thanks.

motoman_gp4_support/launch/load_gp4.launch
```diff
@@ -0,0 +1,3 @@
+<launch>
+  <param name="robot_description" command="$(find xacro)/xacro --inorder '$(find motoman_gp4_support)/urdf/gp4.xacro'" />
```
simonschmeisser, line 2:
```suggestion
  <param name="robot_description" command="$(find xacro)/xacro '$(find motoman_gp4_support)/urdf/gp4.xacro'" />
```
inorder was only needed on ROS `indigo`

cjue:
To my eyes the visual meshes look a bit "defective", as the Yaskawa label is mangled and especially at the base there are a lot of weird polygons. @EricMarcil Did you perform the decimation with Meshlab?

gp4_Screenshot_20211110_132619

EricMarcil:
@cjue I was using Blender in which the label Yaskawa was displaying properly but I can see in Mashlab that it is not.  I believe that the issue is with the faces orientation.  Half the faces are pointing in one direction and the others in the opposite one.  I'll see if I can fix it.
For the model quality, it is a new process that I was using.  I used the model from MotoSim which is already simplified and smaller.  I converted directly the HSF format to DAE.  It look like whoever did the model in Japan pushed a bit to much the decimation on the model.
I'll try to get the original (large) model and do the decimation myself but it might take a couple of weeks to get to it.

motoman_gp4_support/launch/load_gp4.launch
EricMarcil: @gavanderhoorn Can you confirm that we no longer need the --inorder, or are we keeping for backward compatibility to indigo?

cjue:
In my opinion just having a readable logo would be fine. Improving the quality of the other parts of the model would just be a nice to have, not worth delaying the pull request.

cjue:
@gavanderhoorn is there interest in adding an opw parameters file to some or all of the support packages, like this for the gp4?

https://github.com/isys-vision/motoman/blob/opw-gp4/motoman_gp4_support/config/opw_parameters_gp4.yaml

EricMarcil:
@cjue I just push a corrected model with the Yaskawa label that should display correctly.  Let me know if it is correct in your application.

cjue:
Thanks, looks much more presentable now!

In your commit you also added a yaskawa.dae file to the meshes folder, was that on purpose?

gavanderhoorn:
@cjue wrote:
> @gavanderhoorn is there interest in adding an opw parameters file to some or all of the support packages, like this for the gp4?
> 
> https://github.com/isys-vision/motoman/blob/opw-gp4/motoman_gp4_support/config/opw_parameters_gp4.yaml

yes, of course.

motoman_gp4_support/launch/load_gp4.launch
gavanderhoorn: We don't support Indigo any more, but Kinetic is in extended support.

We can keep the `--inorder` for now.

cjue:
> @cjue wrote:
> 
> > @gavanderhoorn is there interest in adding an opw parameters file to some or all of the support packages, like this for the gp4?
> > https://github.com/isys-vision/motoman/blob/opw-gp4/motoman_gp4_support/config/opw_parameters_gp4.yaml
> 
> yes, of course.

@EricMarcil could you then maybe add the `opw_parameters_gp4.yaml` file I linked to to your PR? I tested the parameters with a real robot, and it's consistent with the URDF.

gavanderhoorn:
I'm confused.

Why is there https://github.com/isys-vision/motoman/pull/1 and the PR here?

cjue:
The referenced isys-vision#1 is our temporary internal version we are using to prepare a demo with the GP4. it's not meant as a PR towards this ros-industrial repo.

cjue:
Sorry for my messy link to the opw file in isys-vision#1. I now created a clean PR with just that file added towards Eric's repository GP4-Support branch https://github.com/EricMarcil/motoman/pull/9

EricMarcil:
@cjue I've pushed some new model that I've regenerated from the original 3D model.  If you could do a final review and then we should be able to squash and merge into the main branch.

cjue:
Looks good to me:

gp4_Screenshot_20211126_103837

EricMarcil:
@ted-miller @gavanderhoorn Can one of you do the review so we can merge?

EricMarcil:
@gavanderhoorn Could you approve the review so we can merge.  I think @cjue has already reviewed it but doesn't have the rights to approve.  Thanks

motoman_gp4_support/urdf/gp4_macro.xacro
```diff
@@ -0,0 +1,159 @@
...
+      <parent link="${prefix}link_2_l"/>
+      <child link="${prefix}link_3_u"/>
+      <origin xyz="0 0 0.260" rpy="0 0 0" />
+      <axis xyz="0 -1 0" />
+      <limit lower="${radians(-65)}" upper="${radians(200)}" effort="95.55" velocity="${radians(525)}"/>
+    </joint>
+    <joint name="${prefix}joint_4_r" type="revolute">
+      <parent link="${prefix}link_3_u"/>
+      <child link="${prefix}link_4_r"/>
+      <origin xyz="0.290 0 0.015" rpy="0 0 0" />
```
gavanderhoorn, line 125: This joint has translations in more than a single dimension.

We typically try to avoid that. One way to do that is to move the X translation to the next joint).

gavanderhoorn: This also makes the `origin` of `joint_4_r` coincident with `joint_5_b`. In most other models we don't do this.

motoman_gp4_support/urdf/gp4_macro.xacro
```diff
@@ -0,0 +1,159 @@
...
+        </geometry>
+        <xacro:material_yaskawa_silver/>
+      </visual>
+      <collision>
+        <geometry>
+          <mesh filename="package://motoman_gp4_support/meshes/gp4/collision/link_6_t.stl"/>
+        </geometry>
+      </collision>
+    </link>
+    <link name="${prefix}tool0"/>
```
gavanderhoorn: I would suggest to move this `link` down to where its `joint` is defined (so after line 143), similar to how this is done with `base`.

That would nicely separate the base kinematic chain from the additional frames we define.

gavanderhoorn (changes requested):
I've left some in-line comments.

We might also want to add a `flange` frame here, to work towards #374. Refer to [fanuc_m10ia_support/urdf/m10ia_macro.xacro](https://github.com/ros-industrial/fanuc/blob/74d7d441311590d3950b03da9a1a86409332c2e8/fanuc_m10ia_support/urdf/m10ia_macro.xacro#L165-L171) for an example.

motoman_gp4_support/urdf/gp4_macro.xacro
EricMarcil: Actually, all our models are build this way.  GP8, MH12...  It is based on how Yaskawa builds there internal model.  So it makes it easier to create the model and the meshes orginal (imported from the 3D model) matches the joint_4_r coincident with the joint_5_b.

gavanderhoorn: Avoiding the translation in two dimensions has helped with generating things like IKFast plugins in the past.

Is that also from the DH parameters?

EricMarcil: Yes, it is from the HD parameters.
If you think that it will help with the integration with other tools, I can make the change and make it the standard for future models.

gavanderhoorn: Do the DH params then specify both a `d` and an `a` for that joint?

It has helped in the past, I don't know what the current state is of those code generators.

It would be good to find a compromise / workflow which works well for you, as you are the one who'd have to make manual changes all the time.

EricMarcil:
I completely agree with adding the flange.  I've never like having that last offset in the Tool0.  I was just the way it was done early on, I'm not sure for what reason.

gavanderhoorn:
> I completely agree with adding the flange. I've never like having that last offset in the Tool0. I was just the way it was done early on, I'm not sure for what reason.

Just how things developed over time.

See [here](http://gavanderhoorn.github.io/rep/rep-0199.html#flange) for some more info, especially the *Rationale* paragraph (and the one below, for `tool0`).

And note: `tool0` was never intended to be used as a tool attachment point. It's a TCP frame. Those are two different things, but people started using it to attach EEF models. Hence the introduction of `flange`.

motoman_gp4_support/urdf/gp4_macro.xacro
EricMarcil: Yes, the DH for that joint has both d and a values.  Making the change isn't a big amount of work.  What I'm more concern about is that (even if we don't currently publish the values), I believe that data for CG and Inertia of the links from our standard model is relative that joint location.  I'm not sure exactly how that normal converts to applications like Gazebo, but it might create problems down the line.

motoman_gp4_support/urdf/gp4_macro.xacro
```diff
@@ -141,13 +140,22 @@
...
+    <joint name="${prefix}joint_6_t-flange" type="fixed">
       <origin xyz="0.072 0 0" rpy="${radians(180)} ${radians(-90)} 0"/>
       <parent link="${prefix}link_6_t"/>
-      <child link="${prefix}tool0"/>
+      <child link="${prefix}flange"/>
     </joint>
     <!-- end of joint list -->
-
+	
+	<!-- default tool -->
```
cjue: looks like these two lines have tabs instead of spaces

EricMarcil: Thanks for letting me know.  I just pushed a correction.

cjue:
@EricMarcil: It appears that the collision mesh of link2 restricts the range of motion of link4 significantly:

Screenshot_20211208_093542
Screenshot_20211208_093512
Screenshot_20211208_093415
Screenshot_20211208_093443

In these images joint_3_u is at minimum position (-65°), which in reality is (for some configurations) collision free. The collision mesh of link2 fills in all of the space within link2 that's "hollowed out" especially for link 3 and 4 to allow this range of motion. 

For my application just disabling collisions between link2 and link4 would be a risky workaround. Could you create a smaller collision model for link2?

simonschmeisser:
If `joint_3_u` is already at the minimum joint limit then it should be save to ignore that collision?

cjue:
> If `joint_3_u` is already at the minimum joint limit then it should be save to ignore that collision?

Not in all circumstances: when joint_3_u is at -65° and joint_4_r is not near 0 or +/-180, the rectangular shape of link4 could touch link2 (which the controller will prohibit).

EricMarcil:
@cjue Please test with the new link_2_l collision model to see if that fixes your problem.

cjue:
@EricMarcil Thank you very much, with the new collision model the collision checks allow for the full range of motion of joint_3_u (as stated above: only in certain configurations of joint_4_u). 
The collision model is still a little bit more restrictive than reality, but it's a good solution in my eyes.

EricMarcil:
@gavanderhoorn If you can review this PR, I believe it is ready to merge.

gavanderhoorn:
I've pushed two small commits which resolve some minor issues with the `flange` frame `joint`s and `link`s.

gavanderhoorn (approved):

gavanderhoorn:
We'll wait for CI to become green, then we can merge.

I'll squash-merge this, as there are quite a few fixup commits *and* we don't want/need the Colllada collision meshes to end up in the history.

gavanderhoorn:
Thanks @EricMarcil :+1: 

and thanks @cjue and @simonschmeisser as well.

PR #434, opened 2021-12-03
Alternative for: Added variant model HC10DT-B10

gavanderhoorn:
This is an alternative PR for #415.

It keeps changes to existing variants separate from the addition of the B10 variant, and does not delete any existing files.

It also changes the collision geometry to use STL instead of Collada (for the same reasons as stated in #433.

Commit provenance has been retained where it made sense.


gavanderhoorn:
@EricMarcil: if you could check this and see whether it works for you, I suggest we continue the review here.

gavanderhoorn:
Some additional points for discussion:

 - this variant is called HC10DT IP67 ([here](https://www.yaskawa.eu.com/products/robots/collaborative/productdetail/product/hc10dt-ip67_10516)), with two sub types B10 and B12. Should we follow that naming here?
 - the Collada meshes seem to have some strange colouring (but could be just how this is supposed to look)
 - this seems to have the same 'problem' with the Yaskawa logo/lettering as in #433
 - is the robot in the correct zero position?
   - compare HC10DT:

     hc10dt_zero

     and HC10DT B10:

     hc10dt_b10_zero

   - is the joint transform for `joint_4_r` correct? It seems to be the negative version of the HC 10 DT macro
   - `joint_4_r` has a positive transform here, but for both the base HC 10 and the DT variant it's negative
 - how similar are the three DT variants? Could they potentially share meshes?
 - is the `joint_6_t-tool0` orientation correct? The other variants here have `rpy="${pi} 0 0"`

Additionally some comments:

 - please do not delete files which were there already. We gain little by removing them, and they are part of the standard "structure". We 'suffer' a little duplication, but gain a lot (as it's immediately clear which file is responsible for what, and for which variant). I've reverted most of the changes to existing files, and added new files for the B10 which were missing.
 - we still need the `blue` material on the `visual` elements, otherwise the collision meshes are red (which is the default colour). This does not interfere with the vertex colours in the Collada files. I've already done this.

EricMarcil:
> this variant is called HC10DT IP67 (here), with two sub types B10 and B12. Should we follow that naming here?

Normally, the last digit is not significant from a parameter or mechanical point of view.  In the controller configuration, the last digits is usually replaced by the "*" character (HC10DT-B1*) but that wouldn't work for the file naming.  So I think we can just make reference to the -B10 and if someone ask for a variation B12 variation, we can point them to this one.

> the Collada meshes seem to have some strange colouring (but could be just how this is supposed to look)

Those are the colors that came in with the model.  There are not quite as bright.  I'll check with the GP4 colors to see if they are different.

> this seems to have the same 'problem' with the Yaskawa logo/lettering as in GP4 Support Package #433

Yes, I've figured out how to fix this but forgot that it was also on this PR.  I'll update it shortly.

> is the robot in the correct zero position?
> - is the joint transform for joint_4_r correct? It seems to be the negative version of the HC 10 DT macro
> - joint_4_r has a positive transform here, but for both the base HC 10 and the DT variant it's negative

Yes, the home position has changed between the HC10-DT and HC10-DT-B10.  The changes on the joint_4_r is a result of this change.  The transform is going from Z=-0.500 (arm pointing down) to X=0.500 (arm being forward)

> how similar are the three DT variants? Could they potentially share meshes?

The HC10 and HC10DT already share the same meshes except for the link_6_t.  The HC10DT-B1* is a complete redesign, so all the components are different.  Between the HC10DT-B1* variant, there are physically all the same.

> is the joint_6_t-tool0 orientation correct? The other variants here have rpy="${pi} 0 0"

Yes, that is correct.  It is again a result of the arm pointing forward instead of down.

Comments noted.

EricMarcil:
@gavanderhoorn I suggest that we add the flange link to the HC10DT-B10 since it hasn't been merged to the main repo yet.  What do you think?

gavanderhoorn:
> @gavanderhoorn I suggest that we add the flange link to the HC10DT-B10 since it hasn't been merged to the main repo yet. What do you think?

yes, that would make sense to me.

EricMarcil:
@gavanderhoorn I've pushed an update adding the flange link to the model.  I think with that we should be good to merge.

motoman_hc10_support/urdf/hc10dt_b10_macro.xacro
```diff
@@ -141,12 +140,23 @@
       <axis xyz="-1 0 0" />
       <limit lower="${radians(-180)}" upper="${radians(180)}" effort="31.36" velocity="${radians(250)}"/>
     </joint>
-    <joint name="${prefix}joint_6_t-tool0" type="fixed">
+    <!-- end of joint list -->
+
+    <!-- ROS-Industrial 'flange' frame: attachment point for EEF models -->
+    <link name="${prefix}flange"/>
+    <joint name="${prefix}joint_6_t-flange" type="fixed">
       <origin xyz="0.170 0 0" rpy="${radians(180)} ${radians(-90)} 0"/>
```
gavanderhoorn: Is this correct?

`flange` should have X+ forward. Is that the case?

(I haven't checked in RViz)

EricMarcil: The HC10DT-B10 changed the home position as shown in picture further up.  The flange is pointing forward, so the Z+ is pointing forward.  This should allow to add tools (end-effectors) with angles that matches the values in controller tool defintions.

EricMarcil: It is the same as the GP4
Screenshot from 2022-01-06 14-31-22
4

EricMarcil (dismissed):
I think everything is good to merge.

EricMarcil:
@gavanderhoorn Are you OK with merging this?

gavanderhoorn:
I was actually *just* looking at this.

For some reason, the model renders like this for me right now:

rviz_screenshot_2022_02_02-14_34_33

as only `link_2` and `link_4` render in this strange way, I'm trying to figure out whether dae5eef did something to the normals of the meshes, or changed it in some other way.

---

Edit: as soon as I update RViz (from what's in `osrf/ros:melodic-desktop`), rendering of those meshes changes. Same things with the GP4 model.

I'm wondering whether the export/conversion method you're using results in meshes which RViz doesn't like somehow.

motoman_hc10_support/urdf/hc10dt_b10_macro.xacro
gavanderhoorn: There is a difference between `tool0` and `flange`, and it has to do with their intended use.

`tool0` is supposed to coincide with an "all-zeros toolframe", as the Yaskawa controller defines it. It just happens that such a tool has Z+ "forward" (ie: away from the face plate).

`flange` is supposed to be used to attach EEF models in a composite `.xacro`. To make this as straightforward as possible, `flange` should be oriented such that it adheres to the "normal ROS frame orientation" (as defined in REP-103), which has X+ pointing forward. This is a link-local frame, and as such, "forward" (and the other cardinalities) are defined relative to that link.

`tool0` is not intended to attach any EEF model to, nor a toolframe (ie: an additional TF frame used as a toolframe on the ROS side).

If I interpret the left screenshot correctly, `flange` and `tool0` have the same orientation there, being Z+ forward. The GP4 model also has that.

Both need to be fixed to follow conventions.

I'll submit a PR for the GP4 later.

gavanderhoorn:
Looks like the rendering issue is known. See https://github.com/ros-visualization/rviz/issues/1702. Apparently this has already been fixed (https://github.com/ros-visualization/rviz/issues/1703).

---

Edit: but not released yet.

I'm going to ignore the rendering issue.

motoman_hc10_support/urdf/hc10dt_b10_macro.xacro
gavanderhoorn: By moving the `rpy` from `flange` to `tool0`, we'd get this:

image

this would follow convention again.

Would it be OK to push that commit @EricMarcil ?

EricMarcil: Yes, it fine with me.

gavanderhoorn: Resolved in 8e673ac.

gavanderhoorn:
If you could review again @EricMarcil ?

We'll wait for green CI and then merge.

EricMarcil (approved):

EricMarcil:
@gavanderhoorn Looks good.  Thanks for the help on this one.

gavanderhoorn:
CI is *very* slow.

gavanderhoorn:
Thanks @EricMarcil and apologies for the delay merging.

PR #435, opened 2021-12-11
driver: add support for MotoROS read & write Group IO

gavanderhoorn:
As per subject.

Hasn't been tested on HW yet.

---

Edit: and just repeating https://github.com/ros-industrial/motoman/issues/361#issuecomment-709018080 here for future readers:

> While convenience is why we're adding them here, I'm going to repeat my advice (which I always give when talking about ROS, robot drivers and IO): consider using a proper fieldbus for industrial deployments. Having to add IO support to each and every robot driver doesn't scale very well, and will most likely use custom implementations of protocols which do not have the same level of support, integration nor testing as any of the available (software defined) fieldbuses in existence.
> 
> Even a simple wrapper around a modbus/tcp client would be a good option.


akashjinandra:
@gavanderhoorn would you like some help testing on hardware?

gavanderhoorn:
@akashjinandra: sure, the more testing this gets the better.

You should just have to build `motoman_driver` with this PR merged and make sure to have a recent-ish MotoROS installed on the controller (as in: anything `>=` 1.9.2).

Keep in mind that I haven't tested this at all yet, so it could not work, or it could work immediately, or anything in between.

motoman_msgs/srv/WriteGroupIO.srv
```diff
@@ -0,0 +1,21 @@
...
+#
+# This service does not return anything.
+#
+# Addresses are plain, base-10 integers, as used and displayed by the controller
+# (on the teach pendant for instance).
+#
+# Only the following addresses can be written to:
+#
+#  - 27010 and up : Network Inputs (25010 and up on DX100 and FS100)
+#  - 10010 and up : Universal/General Outputs
```
gavanderhoorn: These are of course not the correct addresses, but I don't have the documentation with me right now.

What are the Group IO addresses the MotoROS service supports? @ted-miller @EricMarcil ?

`SimpleMessage.h` has this:

https://github.com/ros-industrial/motoman/blob/01f3e6d95a56cccf5d9f248da8e8999c48604323/motoman_driver/MotoPlus/SimpleMessage.h#L334

ted-miller: DX100 and FS100:
2501 and up
1001 and up

Other controllers:
2701 and up
1001 and up

As for the upper limit, it varies based on the controller model.
https://github.com/ros-industrial/motoman/blob/01f3e6d95a56cccf5d9f248da8e8999c48604323/motoman_driver/MotoPlus/IoServer.h#L61-L268

gavanderhoorn: Ah, right. It's the grouped version of the regular I/O.

> As for the upper limit, it varies based on the controller model.

we can just refer to the controller documentation for that. This ROS service is no different from using the IOs with any other (external) interface.

motoman_driver/src/io_relay.cpp
```diff
@@ -136,6 +141,40 @@ bool MotomanIORelay::readSingleIoCB(
...
+// Service to read a group IO
+bool MotomanIORelay::readGroupIoCB(
+  motoman_msgs::ReadGroupIO::Request &req,
+  motoman_msgs::ReadGroupIO::Response &res)
+{
+  shared_int io_val = -1;
+  std::string err_msg;
+
+  // send message and release mutex as soon as possible
+  this->mutex_.lock();
```
gavanderhoorn, line 192: This is likely not the correct mutex.

See #389.

(same for the other usages for `mutex_` in this PR)

akashjinandra:
Hello, for clarification sake of the "group IO" I have a question about the numbering terminology. For instance the first bit of group "#036" shows up as "10360" when using the single group IO. The only way to write the group for this would be to write to "1036". Does that make sense? I would think with the terminology of "group" I would use "output group" 036, instead of 1036. Please feel free to share your thoughts.

gavanderhoorn:
I'm not sure I follow.

Is your question about addressing of groups (ie: bytes), or are you observing something you believe is incorrect?

---

Edit:

> when using the single group IO.

this is confusing, as we don't have "multi group IO" here.

EricMarcil:
The first digit represent the type of signal:
0: general input
1: general output
2: external input
3: external output (can't be set directly by write I/O)
...

The 3 next digits represent the group number

The last digit is the bit (0-7) of the group.

If you display the IN/OUT on the pendant, you will see the logical address of the signal in the 2nd column.

image

You can also refer to the Concurrent I/O manual for you controller for more details.

motoman_driver/src/io_relay.cpp
gavanderhoorn: Actually, that's incorrect.

This PR updates the `MotomanIORelay` class which uses `mutex_` exclusively for serialising access to its `default_tcp_connection_`. It doesn't use it for anything else.

Only #436 would need to take #389 into account.

akashjinandra:
Thank you for the clarification. I understand now,  I was pointing out that based on your screenshot in order to use the group service on that byte, meaning that group, you would have to put the address as 1001, instead of just "OG" 001 or 10010.

For inputs however you can just read input group or "IG" as is(when using the read group service)

EricMarcil:
All signals are readable.  In the case of global input, since they start with 0, yes you can directly specify 001.
You can read other signals like specific output in the 50010 range that give information about various controller states. (refer to Concurrent I/O manual for details)

gavanderhoorn:
Be careful with using addresses starting with `0`s. Literals starting with a `0` are parsed as octal numbers by many programming languages.

That would make `008` actually represent `10` base-10.

We might want to add that to the `.srv`.

---

Edit: 5409bda warns against using leading zeros with group addresses.

gavanderhoorn:
Tested on our YRC1000. Works. No regressions detected.

@marip8: as this is my own PR, a review is needed before this can be merged.

Similar to #436 and #439, this is mostly copy-paste and search-replace.

gavanderhoorn:
@marip8: could you please take a look?

motoman_driver/include/motoman_driver/simple_message/messages/motoman_read_group_io_message.h
```diff
@@ -0,0 +1,126 @@
...
+   * \brief Initializes message from a simple message
+   *
+   * \param simple message to construct from
+   *
+   * \return true if message successfully initialized, otherwise false
+   */
+  bool init(industrial::simple_message::SimpleMessage & msg);
+
+  /**
+   * \brief Initializes message from a read single io structure
```
marip8:
```suggestion
   * \brief Initializes message from a read group io structure
```

motoman_driver/include/motoman_driver/simple_message/messages/motoman_read_group_io_reply_message.h
```diff
@@ -0,0 +1,130 @@
...
+   * \brief Initializes message from a simple message
+   *
+   * \param simple message to construct from
+   *
+   * \return true if message successfully initialized, otherwise false
+   */
+  bool init(industrial::simple_message::SimpleMessage & msg);
+
+  /**
+   * \brief Initializes message from a read single io reply structure
```
marip8:
```suggestion
   * \brief Initializes message from a read group io reply structure
```

motoman_driver/include/motoman_driver/simple_message/motoman_write_group_io_reply.h
```diff
@@ -0,0 +1,182 @@
...
+  ~WriteGroupIOReply(void);
+
+  /**
+   * \brief Initializes a empty motion control reply
+   *
+   */
+  void init();
+
+  /**
+   * \brief Initializes a complete read single io reply
```
marip8:
```suggestion
   * \brief Initializes a complete read group io reply
```

motoman_driver/src/simple_message/motoman_read_group_io_reply.cpp
```diff
@@ -0,0 +1,161 @@
...
+  this->init();
+}
+ReadGroupIOReply::~ReadGroupIOReply(void)
+{
+}
+
+void ReadGroupIOReply::init()
+{
+  // TODO( ): is '0' a good initial value?
+  this->init(0/*value*/, ReadGroupIOReplyResultCodes::SUCCESS);
```
marip8, line 70: Any real concern here?

gavanderhoorn:
Fixed the conflicts.

motoman_msgs/srv/WriteGroupIO.srv
```diff
@@ -0,0 +1,25 @@
+
+# Write 'value' to the Group IO element at 'address'.
+#
+# This service does not return anything.
```
marip8: Technically it returns a boolean and string

marip8 (commented):
Minor suggestions for comments, but otherwise looks fine

motoman_msgs/srv/WriteGroupIO.srv
gavanderhoorn: Yes, true. I already addressed this for #439 in 8d5a823af05860d40a819242d36318c76ae045ab.

I'll add a similar commit here.

motoman_driver/src/simple_message/motoman_read_group_io_reply.cpp
gavanderhoorn: well, I just don't know whether `0` is a correct default value.

I also wouldn't know any better value, or whether we'd want to use a sentinel of some sort here.

Technically, group IO is max 8 bit if I'm not mistaken, so there are certainly values outside the permissible range.

motoman_msgs/srv/WriteGroupIO.srv
gavanderhoorn: Fixed: 81f7851e.

motoman_msgs/srv/ReadGroupIO.srv
```diff
@@ -0,0 +1,21 @@
...
+#       specifying the wrong address to read.
+#
+# Refer also the Yaskawa Motoman documentation on IO addressing and
+# configuration.
+
+uint32 address
+---
+string message
+bool success
+uint32 value
```
gavanderhoorn: We could consider changing this to `uint8`, to match the width of the IO groups on the controller.

@ted-miller / @EricMarcil: group IO is always 8bit, isn't it?

motoman_msgs/srv/WriteGroupIO.srv
```diff
@@ -0,0 +1,23 @@
...
+#
+# NOTE: many programming languages will parse literals starting with '0' as
+#       octal numbers. Do not add leading zeros to group addresses to avoid
+#       specifying the wrong address to write to.
+
+# Refer also the Yaskawa Motoman documentation on IO addressing and
+# configuration.
+
+uint32 address
+uint32 value
```
gavanderhoorn: Same here.

gavanderhoorn:
Rebased.

@marip8: could you take a look here as well?

gavanderhoorn:
Friendly ping @marip8

marip8 (approved):
I think this looks fine. Do you want to switch the message types from `uint32` to `uint8` per your comments or save that for later?

motoman_msgs/srv/ReadGroupIO.srv
gavanderhoorn: Friendly ping @ted-miller

ted-miller: Yes, a group is always 8 bit. 
There are some operations in Inform and CIO that will use 16 bits. But, these are always considered to use two groups.

gavanderhoorn:
Thanks @marip8.

> I think this looks fine. Do you want to switch the message types from `uint32` to `uint8` per your comments or save that for later?

now that @ted-miller confirmed they're always 8bit, I've changed the `value` field to `uint8` instead of `uint32`.

gavanderhoorn:
CI green.

Thanks for the review @marip8 :+1:

PR #436, opened 2021-12-11
driver: add support for MotoROS SelectTool

### LINKED ISSUE #419: ROS API support for ROS_MSG_MOTO_SELECT_TOOL
## Description
When tool changing in ROS, it is also important to change the set tool on the controller (according to Yaskawa Motoman YRC1000micro manual and pendant display messages **not doing this can cause premature failure of the robot**), however it is unclear whether `ROS_MSG_MOTO_SELECT_TOOL` has been exposed as a service or action server in the ROS API side. There doesn't seem to be an [existing simple_message](https://github.com/ros-industrial/motoman/tree/kinetic-devel/motoman_driver/src/simple_message) for the functionality.

## My environment
ROS Noetic in a docker container with Ubuntu 20.04
Moveit! version 1.1.5
ros-industrial-client with modified motoman driver v1.9.5 ([our own point streamer branch](https://github.com/additiveautomations/motoman/tree/point-streamer))
motoman gp8 robot

## Problem
There seems to be no existing functionality to use `ROS_MSG_MOTO_SELECT_TOOL` and also it is unclear if the low-level control will actually use it even if a new tool is selected with different mass, intertia and centre of gravity.

## Action Taken So Far
Read manuals and looked through which [simple_messages have been defined already](https://github.com/ros-industrial/motoman/tree/kinetic-devel/motoman_driver/src/simple_message).

## Expected behaviour
Set tool through ROS and see the pendant update the selected tool.

## Actual behaviour
The set tool doesn't change even if a tool is changed through ROS.

gavanderhoorn:
As per subject.

Hasn't been tested on HW yet.

### DIFF
--- a/motoman_driver/include/motoman_driver/simple_message/messages/motoman_select_tool_message.h
+++ b/motoman_driver/include/motoman_driver/simple_message/messages/motoman_select_tool_message.h
@@ -90,7 +90,7 @@ public:
    *
    * \return true if message successfully initialized, otherwise false
    */
-  bool init(industrial::simple_message::SimpleMessage & msg);
+  bool init(industrial::simple_message::SimpleMessage & msg) override;
 
   /**
    * \brief Initializes message from a select tool structure
@@ -104,13 +104,13 @@ public:
    * \brief Initializes a new message
    *
    */
-  void init();
+  void init() override;
 
   // Overrides - SimpleSerialize
-  bool load(industrial::byte_array::ByteArray *buffer);
-  bool unload(industrial::byte_array::ByteArray *buffer);
+  bool load(industrial::byte_array::ByteArray *buffer) override;
+  bool unload(industrial::byte_array::ByteArray *buffer) override;
 
-  unsigned int byteLength()
+  unsigned int byteLength() override
   {
     return this->cmd_.byteLength();
   }
--- a/motoman_driver/include/motoman_driver/simple_message/messages/motoman_select_tool_message.h
+++ b/motoman_driver/include/motoman_driver/simple_message/messages/motoman_select_tool_message.h
@@ -1,0 +1,126 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ * Copyright (c) 2021, Delft Robotics Institute
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *  * Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  * Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *  * Neither the name of the Delft Robotics Institute, nor the names
+ *    of its contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * \author G.A. vd. Hoorn (TU Delft Robotics Institute)
+ */
+
+#ifndef MOTOMAN_DRIVER_SIMPLE_MESSAGE_MESSAGES_MOTOMAN_SELECT_TOOL_MESSAGE_H
+#define MOTOMAN_DRIVER_SIMPLE_MESSAGE_MESSAGES_MOTOMAN_SELECT_TOOL_MESSAGE_H
+
+#ifdef ROS
+#include "simple_message/typed_message.h"
+#include "simple_message/shared_types.h"
+#include "motoman_driver/simple_message/motoman_simple_message.h"
+#include "motoman_driver/simple_message/motoman_select_tool.h"
+
+#endif
+
+#ifdef MOTOPLUS
+#include "typed_message.h"           // NOLINT(build/include)
+#include "shared_types.h"            // NOLINT(build/include)
+#include "motoman_simple_message.h"  // NOLINT(build/include)
+#include "motoman_select_tool.h"     // NOLINT(build/include)
+
+#endif
+
+namespace motoman
+{
+namespace simple_message
+{
+namespace misc
+{
+/**
+ * \brief Class encapsulated motoman select tool file message generation methods
+ * (either to or from a industrial::simple_message::SimpleMessage type).
+ *
+ * This message simply wraps the following data type:
+ *   motoman::simple_message::misc::SelectTool
+ * The data portion of this typed message matches SelectTool exactly.
+ *
+ *
+ * THIS CLASS IS NOT THREAD-SAFE
+ *
+ */
+class SelectToolMessage : public industrial::typed_message::TypedMessage
+{
+public:
+  /**
+   * \brief Default constructor
+   *
+   * This method creates an empty message.
+   *
+   */
+  SelectToolMessage(void);
+  /**
+   * \brief Destructor
+   *
+   */
+  ~SelectToolMessage(void);
+  /**
+   * \brief Initializes message from a simple message
+   *
+   * \param simple message to construct from
+   *
+   * \return true if message successfully initialized, otherwise false
+   */
+  bool init(industrial::simple_message::SimpleMessage & msg);
+
+  /**
+   * \brief Initializes message from a select tool structure
+   *
+   * \param cmd select tool data structure
+   *
+   */
+  void init(motoman::simple_message::misc::SelectTool & cmd);
+
+  /**
+   * \brief Initializes a new message
+   *
+   */
+  void init();
+
+  // Overrides - SimpleSerialize
+  bool load(industrial::byte_array::ByteArray *buffer);
+  bool unload(industrial::byte_array::ByteArray *buffer);
+
+  unsigned int byteLength()
+  {
+    return this->cmd_.byteLength();
+  }
+
+  motoman::simple_message::misc::SelectTool cmd_;
+
+private:
+};
+}  // namespace misc
+}  // namespace simple_message
+}  // namespace motoman
+
+#endif  // MOTOMAN_DRIVER_SIMPLE_MESSAGE_MESSAGES_MOTOMAN_SELECT_TOOL_MESSAGE_H
--- a/motoman_driver/include/motoman_driver/simple_message/motoman_select_tool.h
+++ b/motoman_driver/include/motoman_driver/simple_message/motoman_select_tool.h
@@ -182,9 +182,9 @@ public:
   bool operator==(SelectTool &rhs);
 
   // Overrides - SimpleSerialize
-  bool load(industrial::byte_array::ByteArray *buffer);
-  bool unload(industrial::byte_array::ByteArray *buffer);
-  unsigned int byteLength()
+  bool load(industrial::byte_array::ByteArray *buffer) override;
+  bool unload(industrial::byte_array::ByteArray *buffer) override;
+  unsigned int byteLength() override
   {
     return 3 * sizeof(industrial::shared_types::shared_int);
   }
--- a/motoman_driver/include/motoman_driver/simple_message/motoman_select_tool.h
+++ b/motoman_driver/include/motoman_driver/simple_message/motoman_select_tool.h
@@ -179,7 +179,7 @@ public:
    *
    * \return true if equal
    */
-  bool operator==(SelectTool &rhs);
+  bool operator==(const SelectTool &rhs) const;
 
   // Overrides - SimpleSerialize
   bool load(industrial::byte_array::ByteArray *buffer) override;
--- a/motoman_driver/include/motoman_driver/simple_message/motoman_select_tool.h
+++ b/motoman_driver/include/motoman_driver/simple_message/motoman_select_tool.h
@@ -1,0 +1,212 @@
+/*
+ * Software License Agreement (BSD License)
+ *
+ * Copyright (c) 2021, Delft Robotics Institute
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *  * Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  * Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *  * Neither the name of the Delft Robotics Institute, nor the names
+ *    of its contributors may be used to endorse or promote products derived
+ *    from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ * \author G.A. vd. Hoorn (TU Delft Robotics Institute)
+ */
+
+#ifndef MOTOMAN_DRIVER_SIMPLE_MESSAGE_MOTOMAN_SELECT_TOOL_H
+#define MOTOMAN_DRIVER_SIMPLE_MESSAGE_MOTOMAN_SELECT_TOOL_H
+
+#ifdef ROS
+#include "simple_message/simple_serialize.h"
+#include "simple_message/shared_types.h"
+#include "simple_message/log_wrapper.h"
+#endif
+
+#ifdef MOTOPLUS
+#include "simple_serialize.h"  // NOLINT(build/include)
+#include "shared_types.h"      // NOLINT(build/include)
+#include "log_wrapper.h"       // NOLINT(build/include)
+#endif
+
+namespace motoman
+{
+namespace simple_message
+{
+namespace misc
+{
+
+/**
+ * \brief Class encapsulated select tool data. Motoman specific interface
+ * to select a specific tool file on the controller.
+ *
+ * The byte representation of a select tool command is as follows
+ * (in order lowest index to highest). The standard sizes are given,
+ * but can change based on type sizes:
+ *
+ *   member:             type                                      size
+ *   group_number        (industrial::shared_types::shared_int)    4  bytes
+ *   tool_number         (industrial::shared_types::shared_int)    4  bytes
+ *   sequence_number     (industrial::shared_types::shared_int)    4  bytes
+ *
+ * Note: sequence_number is marked as optional by MotoROS. It will be included
+ * in the SimpleMessage service request, but does not have to be set to any
+ * specific value. It can be used as a "tracking number" that will be "echoed
+ * back in the response message" (according to MotoROS source code comments).
+ *
+ * THIS CLASS IS NOT THREAD-SAFE
+ *
+ */
+
+class SelectTool : public industrial::simple_serialize::SimpleSerialize
+{
+public:
+  /**
+   * \brief Default constructor
+   *
+   * This method creates empty data.
+   *
+   */
+  SelectTool(void);
+
+  /**
+   * \brief Destructor
+   *
+   */
+  ~SelectTool(void);
+
+  /**
+   * \brief Initializes a empty select tool command
+   *
+   */
+  void init();
+
+  /**
+   * \brief Initializes a complete select tool command
+   *
+   * Note: sequence_number default value of 0.
+   */
+  void init(industrial::shared_types::shared_int group_number,
+    industrial::shared_types::shared_int tool_number,
+    industrial::shared_types::shared_int sequence_number = 0);
+
+  /**
+   * \brief Sets group number
+   *
+   * \param group_number Controller group the tool file is defined for
+   */
+  void setGroupNumber(industrial::shared_types::shared_int group_number)
+  {
+    this->group_number_ = group_number;
+  }
+
+  /**
+   * \brief Returns the group number
+   *
+   * \return group number
+   */
+  industrial::shared_types::shared_int getGroupNumber()
+  {
+    return this->group_number_;
+  }
+
+  /**
+   * \brief Sets tool number
+   *
+   * \param tool_number Tool file identifier
+   */
+  void setToolNumber(industrial::shared_types::shared_int tool_number)
+  {
+    this->tool_number_ = tool_number;
+  }
+
+  /**
+   * \brief Returns the tool file identifier
+   *
+   * \return tool number
+   */
+  industrial::shared_types::shared_int getToolNumber()
+  {
+    return this->tool_number_;
+  }
+
+  /**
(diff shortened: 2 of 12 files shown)


riv-rbush:
@gavanderhoorn Do you want help testing on hardware?

motoman_msgs/srv/SelectTool.srv
```diff
@@ -0,0 +1,92 @@
...
+#
+# NOTE: this MUST correspond to a group ID currently defined and active on the
+#       controller. The ROS service does not check whether the value specified
+#       here is legal. The MotoROS server will however check this, will refuse
+#       to switch to an illegal value and will report this to the ROS driver.
+#       Callers will be notified by 'success' set to 'false' and an appropriate
+#       error message being provided via the 'message' field of the service
+#       response.
+#
+# legal-values: [0, total_nr_of_groups)
```
gavanderhoorn, line 100: I haven't checked whether this is 0-based or 1-based.

@ted-miller @EricMarcil ?

motoman_msgs/srv/SelectTool.srv
```diff
@@ -0,0 +1,92 @@
...
+int32 group_number
+
+
+# Numeric ID of the tool file to switch to for the specified group.
+#
+# NOTE: see the NOTE for the 'group_number' field above. Illegal values for
+#       this field will result in failed service invocations similar to what
+#       is described for that field.
+#
+# legal-values: [0, total_nr_of_defined_tool_files)
```
gavanderhoorn: Same here.

motoman_msgs/srv/SelectTool.srv
```diff
@@ -0,0 +1,92 @@
...
+#       here is legal. The MotoROS server will however check this, will refuse
+#       to switch to an illegal value and will report this to the ROS driver.
+#       Callers will be notified by 'success' set to 'false' and an appropriate
+#       error message being provided via the 'message' field of the service
+#       response.
+#
+# legal-values: [0, total_nr_of_groups)
+# required: yes (absence-causes-service-failure)
+# default: no-default
+int32 group_number
```
gavanderhoorn: I've kept these fields as signed integers, as that's what MotoROS has here:

https://github.com/ros-industrial/motoman/blob/1509f9b8cf003a1eeda7161b4a14c95b51b5297e/motoman_driver/MotoPlus/SimpleMessage.h#L271-L276

We could opt to spec `uint32` here (edit: in the ROS `.srv` I mean) to prevent users from setting `< 0` here, at least on the ROS API side of things.

motoman_msgs/srv/SelectTool.srv
```diff
@@ -0,0 +1,92 @@
...
+# Numeric ID of the tool file to switch to for the specified group.
+#
+# NOTE: see the NOTE for the 'group_number' field above. Illegal values for
+#       this field will result in failed service invocations similar to what
+#       is described for that field.
+#
+# legal-values: [0, total_nr_of_defined_tool_files)
+# required: yes (absence-causes-service-failure)
+# default: no-default
+int32 tool_number
```
gavanderhoorn: Same here.

motoman_msgs/srv/SelectTool.srv
```diff
@@ -0,0 +1,92 @@
...
+
+# TODO: might want to expose 'sequence_number' here as well
```
gavanderhoorn, line 119: We don't expose `sequence_number` here as it's a SimpleMessage-level piece of information.

motoman_msgs/srv/SelectTool.srv
```diff
@@ -0,0 +1,92 @@
...
+# Changes the tool file used by the controller for (Moto)ROS-controlled motions.
+#
+# NOTE 1: this changes the active tool for (Moto)ROS-controlled motions ONLY.
+#
+#         Regular INFORM programs, jogging or any other (programmed) motions
+#         executed by the controller will NOT use the new tool.
+#
+# NOTE 2: any tool switch initiated by invoking this service WILL however be
+#         taken into account by safety systems such as the FSU and PFL while
+#         executing (Moto)ROS-initiated motions.
```
gavanderhoorn: @EricMarcil: I believe this accurately reflects what you observed during testing?

motoman_msgs/srv/SelectTool.srv
```diff
@@ -0,0 +1,92 @@
...
+#
+# NOTE 3: active tool will be switched AFTER the controller has completed
+#         execution of the last trajectory segment that was sent to the
+#         controller BEFORE this service was called.
+#
+#         The motion queue internal to MotoROS caches a number of segments in
```
gavanderhoorn: @ted-miller @EricMarcil: is there any way we can make this more exact (as in: number of seconds of motion, number of samples, something else)?

motoman_msgs/srv/SelectTool.srv
```diff
@@ -0,0 +1,92 @@
...
+#         Regular INFORM programs, jogging or any other (programmed) motions
+#         executed by the controller will NOT use the new tool.
+#
+# NOTE 2: any tool switch initiated by invoking this service WILL however be
+#         taken into account by safety systems such as the FSU and PFL while
+#         executing (Moto)ROS-initiated motions.
+#
+# NOTE 3: active tool will be switched AFTER the controller has completed
+#         execution of the last trajectory segment that was sent to the
+#         controller BEFORE this service was called.
```
gavanderhoorn: It's a bit complex, but I believe this captures what we discussed in https://github.com/ros-industrial/motoman/pull/440#issuecomment-992651912 and the follow-up comments.

motoman_msgs/srv/SelectTool.srv
```diff
@@ -0,0 +1,92 @@
...
+#         The motion queue internal to MotoROS caches a number of segments in
+#         a FIFO. Only segments received AFTER this service was invoked will
+#         be executed with the new tool. Any segments received before a tool
+#         switch was invoked will use the old tool.
+#
+#         To guarantee motion will be executed using the new tool, applications
+#         should check the 'in_motion' field (part of the RobotStatus message
+#         published on the 'robot_status' topic by the driver) and invoke the
+#         service when the robot has come to a stop (and the motion queue of
+#         MotoROS is empty). Any subsequent trajectories will use the new tool.
```
gavanderhoorn: I don't see any other way to "guarantee" a tool file change is taken into account for any particular motion with the current definition and implementation of the MotoROS service.

gavanderhoorn (commented):
Some in-line comments.

gavanderhoorn:
> @gavanderhoorn Do you want help testing on hardware?

That would be very much appreciated, yes.

Similar to what I wrote in #435 and #439, merge and build should be enough on the ROS side.

If you don't actually need the `mpSetToolNo(..)` part of this (see discussion in #440), any MotoROS version `>=` 1.9.0 should work.

gavanderhoorn:
I've left some in-line comments @EricMarcil and @ted-miller.

If you could take a look that would be appreciated.

@marip8 @steviedale: your thoughts would also be appreciated.

Note though that I really just did a copy-paste and search-replace of the `ReadSingleIO` files, so if you come across something which you feel should be done in a nicer way (or is duplicated unnecessarily), that would be expected. We might consider deduplicating some of the result code -> error message functionality, but this PR focuses on the functionality first.

motoman_msgs/srv/SelectTool.srv
gavanderhoorn: Oh, from [this](https://github.com/ros-industrial/motoman/blob/1509f9b8cf003a1eeda7161b4a14c95b51b5297e/motoman_driver/MotoPlus/SimpleMessage.h#L271-L276) it would appear it's indeed 0-based.

steviedale:
> @marip8 @steviedale: your thoughts would also be appreciated.

Hey @gavanderhoorn, just wanted to give you a heads up that I'm no longer working with Southwest Research Institute so I won't be contributing to ROS-I anymore. But its been a pleasure working with you and the rest of the community and I wish you all the best!

motoman_msgs/srv/SelectTool.srv
```diff
@@ -0,0 +1,92 @@
+
+# Changes the tool file used by the controller for (Moto)ROS-controlled motions.
+#
+# NOTE 1: this changes the active tool for (Moto)ROS-controlled motions ONLY.
+#
+#         Regular INFORM programs, jogging or any other (programmed) motions
```
ted-miller: This note is not valid if #440 gets merged.

motoman_msgs/srv/SelectTool.srv
ted-miller: Sounds accurate to me, based on Eric's report

motoman_msgs/srv/SelectTool.srv
ted-miller: To put another way, "The active tool will not change for any trajectory that has already been submitted to the controller."

gavanderhoorn: @ted-miller wrote:
> To put another way, "The active tool will not change for any trajectory that has already been submitted to the controller."

is that true? Because if IIUC, `motoman_driver` essentially drip-feeds a `JointTrajectoryAction` goal, segment per segment.

MotoROS has an internal buffer, and it accepts new `JointTrajPt`s as long as that buffer is not full. If it is full, it sends a "busy" response to `motoman_driver`, which then keeps trying to queue the current point (ie: the one it couldn't push into the queue).

At any time during this drip-feeding could `SELECT_TOOL` be called.

motoman_msgs/srv/SelectTool.srv
ted-miller: Not necessarily. It's not just the max length of the incremental queue. There could also be pending data that hasn't been broken down yet because the queue is full.

motoman_msgs/srv/SelectTool.srv
gavanderhoorn: True, but see https://github.com/ros-industrial/motoman/pull/440#discussion_r767944176.

I'm not sure, but I believe @EricMarcil would rather we do not add `mpSetToolNo(..)`.

I added this `NOTE` assuming that was what we were going to end up with.

motoman_msgs/srv/SelectTool.srv
ted-miller: I agree

motoman_msgs/srv/SelectTool.srv
```diff
@@ -0,0 +1,92 @@
...
+#
+#         To guarantee motion will be executed using the new tool, applications
+#         should check the 'in_motion' field (part of the RobotStatus message
+#         published on the 'robot_status' topic by the driver) and invoke the
+#         service when the robot has come to a stop (and the motion queue of
+#         MotoROS is empty). Any subsequent trajectories will use the new tool.
+#
+# NOTE 4: as MotoROS currently only executes joint space trajectories, changing
+#         tool file with this service DOES NOT affect TCP poses while executing
+#         trajectories.
```
ted-miller: But the TCP will be honored for things such as safety zones.

motoman_msgs/srv/SelectTool.srv
gavanderhoorn: We could consider publishing a MotoROS specific message (on a ROS topic) which includes information about the MotoROS-internal queue. That could allow clients to call `SELECT_TOOL` earlier (as there would be no need for checking `in_motion`).

However MotoROS currently only returns information about the queue via a SimpleMessage service, which would perhaps be a bit too much overhead for periodic use.

motoman_msgs/srv/SelectTool.srv
gavanderhoorn: Yes. I hoped `NOTE 2` would already clarify this?

I could add a reference to that note here.

motoman_msgs/srv/SelectTool.srv
ted-miller: There is range checking in the code. But yes... those really should be unsigned.

motoman_msgs/srv/SelectTool.srv
ted-miller: 0 - 63.
Even an undefined tool file is 'valid'. It's just zeroed out.

motoman_msgs/srv/SelectTool.srv
ted-miller: > At any time during this drip-feeding could SELECT_TOOL be called.

That is true. Hence, your message below about the guarantee.

motoman_msgs/srv/SelectTool.srv
gavanderhoorn: Could you add a comment with a [suggestion](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/commenting-on-a-pull-request#adding-line-comments-to-a-pull-request) (step 6)? I'll accept and commit it.

motoman_msgs/srv/SelectTool.srv
gavanderhoorn: I'll change the fields in the ROS `.srv` to unsigned. One source of potential problems less.

motoman_msgs/srv/SelectTool.srv
EricMarcil: Yes, tool indexes are 0 based.
Tool #0 actually exists in the controller.

EricMarcil: FYI, largest tool is 63

motoman_msgs/srv/SelectTool.srv
EricMarcil: Yes, this is accurate

motoman_msgs/srv/SelectTool.srv
EricMarcil: I've discussed it with Ted and we both agreed to put it in.  So we can specify that the current jogging tool will also be changed.

motoman_msgs/srv/SelectTool.srv
EricMarcil: I think it is confusing to refer the "TCP poses" here.  How about simply say:  "...DOES NOT affect the robot trajectory."
We could also add:  "However, proper tool selection is important as it is used for motion dynamics and safety calculations."

motoman_msgs/srv/SelectTool.srv
gavanderhoorn: Ok. I've updated the `.srv` docs to reflect this.

motoman_msgs/srv/SelectTool.srv
gavanderhoorn: I've already fixed this in the commits I just pushed.

motoman_msgs/srv/SelectTool.srv
gavanderhoorn: I've used slightly more generic language to clarify this.

Please check whether the current phrasing is OK.

gavanderhoorn:
@rr-robert: just for planning purposes: could you give me an indication when you'd have time to test this on your HW?

gavanderhoorn:
Oh and please note that you *should* actually use the updated MotoROS `.out` from #440.

gavanderhoorn:
Just realised: the `select_tool` service would be one we should/could expose per-group, instead of only a single instance per controller (or: driver instance).

The implementation in this PR only exposes a global one (which works as `SelectTool` requires both a `group_number` and a `tool_number`).

gavanderhoorn:
I've rebased this on `kinetic-devel`, so it now includes #389 which was just merged.

17f9482 has been updated to make use of the new `smpl_msg_conx_mutex_` instead of the incorrect `mutex_` (which is in `industrial_robot_client::JointTrajectoryStreamer`, see also #389).

gavanderhoorn:
@rr-robert: friendly ping?

akashjinandra:
Hello @gavanderhoorn, just tested this on the YRC1000 controller after switching to the 199 .out file , and was able to switch tools and validated that it switched on the pendant.

gavanderhoorn:
Thanks @akashjinandra.

gavanderhoorn:
I've added 1ce68a4 which clarifies changing the tool used for jogging is not supported on FS100 controllers.

gavanderhoorn:
@akashjinandra: have you also tested trying to switch to a tool file outside the defined range (ie: `> 63`)? And for a non-existent group?

gavanderhoorn:
@marip8: it's not the nicest code (as it's copy-paste search-replace), but it is apparently functional, so the ROS-side could now also be reviewed.

motoman_driver/include/motoman_driver/simple_message/messages/motoman_select_tool_message.h
```diff
@@ -0,0 +1,126 @@
...
+/**
+ * \brief Class encapsulated motoman select tool file message generation methods
+ * (either to or from a industrial::simple_message::SimpleMessage type).
+ *
+ * This message simply wraps the following data type:
+ *   motoman::simple_message::misc::SelectTool
+ * The data portion of this typed message matches SelectTool exactly.
+ *
+ *
+ * THIS CLASS IS NOT THREAD-SAFE
```
marip8, line 68: Is this actually true?

motoman_driver/include/motoman_driver/simple_message/messages/motoman_select_tool_message.h
```diff
@@ -0,0 +1,126 @@
...
+  void init();
+
+  // Overrides - SimpleSerialize
+  bool load(industrial::byte_array::ByteArray *buffer);
+  bool unload(industrial::byte_array::ByteArray *buffer);
+
+  unsigned int byteLength()
+  {
+    return this->cmd_.byteLength();
+  }
```
marip8: Would be nice to mark these as `override` for clarity

motoman_driver/include/motoman_driver/simple_message/motoman_select_tool.h
```diff
@@ -0,0 +1,212 @@
...
+
+  /**
+   * \brief Returns the optional sequence number
+   *
+   * \return sequence number
+   */
+  industrial::shared_types::shared_int getSequenceNumber()
+  {
+    return this->sequence_number_;
+  }
```
marip8, line 168: Why not just expose these variables as public members instead of making getters and setters? It would cut out a lot of unnecessary code

motoman_driver/include/motoman_driver/simple_message/motoman_select_tool.h
```diff
@@ -0,0 +1,212 @@
...
+
+  /**
+   * \brief Copies the passed in value
+   *
+   * \param src (value to copy)
+   */
+  void copyFrom(SelectTool &src);
```
marip8, line 175:
```suggestion
  void copyFrom(const SelectTool &src);
```

Why not just make a copy constructor instead?

motoman_driver/include/motoman_driver/simple_message/motoman_select_tool.h
```diff
@@ -0,0 +1,212 @@
...
+
+  /**
+   * \brief == operator implementation
+   *
+   * \return true if equal
+   */
+  bool operator==(SelectTool &rhs);
```
marip8:
```suggestion
  bool operator==(const SelectTool &rhs) const;
```

motoman_driver/src/joint_trajectory_streamer.cpp
```diff
@@ -175,6 +179,43 @@ bool MotomanJointTrajectoryStreamer::enableRobotCB(std_srvs::Trigger::Request& r
...
+
+    // the ROS service was successfully invoked, so return true (even if the
+    // MotoROS service was not successfully invoked)
+    return true;
+  }
+
+  ROS_DEBUG_STREAM("Tool file changed to: " << req.tool_number << ", for group: " << req.group_number);
+
+  // no failure, so no need for an additional message
+  res.success = true;
```
marip8: Simplify

```suggestion
  {
    // SmplMsgConnection is not thread safe, so lock first
    // NOTE: motion_ctrl_ uses the SmplMsgConnection here
    const std::lock_guard<std::mutex> lock{smpl_msg_conx_mutex_};
    res.success = motion_ctrl_.selectToolFile(req.group_number, req.tool_number, err_msg);
  }

  if (!res.success)
  {
    // provide caller with failure indication
    // TODO( ): should we also return the result code?
    std::stringstream message;
    message << "Tool file change failed (grp: " << req.group_number
      << ", tool: " << req.tool_number << "): " << err_msg;
    res.message = message.str();
    ROS_ERROR_STREAM(res.message);
  }
  else
  {
    ROS_DEBUG_STREAM("Tool file changed to: " << req.tool_number << ", for group: " << req.group_number);
  }
  
  // the ROS service was successfully invoked, so return true (even if the
  // MotoROS service was not successfully invoked)
  return true;
```

motoman_msgs/srv/SelectTool.srv
```diff
@@ -0,0 +1,132 @@
+
+# Change the active tool file on the controller.
```
marip8, line 2: I like the level of documentation here, but it might be more accessible to an average user in a wiki page or README

marip8 (commented):
Mostly comments about simplifying code and following const-correctness. Also, would it make sense to push the tool change service into a separate node like the current I/O interface? That would eliminate the need for the connection mutex, but it does seem like the current tool is loosely related to motion control

motoman_driver/include/motoman_driver/simple_message/messages/motoman_select_tool_message.h
gavanderhoorn: there's nothing preventing anyone from using the same instance of this class concurrently from two different threads of execution and suffering from lost-updates.

So I would say it's true.

Whether it's a useful comment on a class like this is debatable.

motoman_driver/include/motoman_driver/simple_message/motoman_select_tool.h
gavanderhoorn: As I wrote earlier: this is all copy-paste and search-and-replace.

We could rewrite it all, but that would mean it's no longer a search-and-replace job.

As to why it was written like this: we'd have to ask the previous developers.

motoman_driver/include/motoman_driver/simple_message/motoman_select_tool.h
gavanderhoorn: Because cargo-cult, and copy-paste, search-and-replace ;)

Perhaps @JeremyZoss can remember.

motoman_msgs/srv/SelectTool.srv
gavanderhoorn: True, but that would increase the distance between the `.srv` and its documentation.

I also expect it's less likely for a ROS developer to lookup a wiki when looking for message documentation than looking at the `.srv` itself, either directly or through `rossrv show`.

Additionally, if `motoman_msgs` ever gets released, the buildfarm would generate HTML documentation based on this "for free".

We could add a reference to a readme though.

motoman_driver/include/motoman_driver/simple_message/motoman_select_tool.h
JeremyZoss: I think most of this is just following the code pattern Shaun used in his original implementation (see [joint_data.h](https://github.com/ros-industrial/industrial_core/commit/3e065210cd8c428dffd94f23dff18fd4b6f79f75#diff-940cfbba5085b9567c89a09d37d661d8c7f5948e29e815904326cff4c8321fb1R129) from 2012), for code consistency.  Some of that design pattern was based on using literally the same code base for ROS and MotoROS `simple_message` classes, which inherits some limitations from MotoROS's C++ implementation.  But that code-sharing is not used in current drivers.  simple_message could probably use some cleanup, but I think it would make sense to tackle that globally (for consistency).  In my opinion, best to keep the style of this PR to match the style of the rest of simple_message.

gavanderhoorn: It's a wonder he was even able to use C++. I don't believe C++ is supported officially by MotoPlus.

gavanderhoorn:
@marip8 wrote:
> would it make sense to push the tool change service into a separate node like the current I/O interface? That would eliminate the need for the connection mutex, but it does seem like the current tool is loosely related to motion control

selecting the tool file is actually intimately tied to "motion" (even if we only execute joint space trajectories at the moment, it still notifies the Yaskawa controller of important information about what is actually mounted on the flange), so from that perspective putting it in `MotomanMotionCtrl` made sense to me.

As to using a separate connection: that would require changes to MotoROS, and quite substantial ones. It would essentially either require a new socket added to the existing `MotionServer` and refactoring the network handling code, or a completely new task, with a new server implementation and refactoring of the `MotionServer` to transplant the affected code paths.

All-in-all, I don't believe it'd be worth it.

motoman_driver/include/motoman_driver/simple_message/motoman_select_tool.h
gavanderhoorn: @JeremyZoss wrote:
>  In my opinion, best to keep the style of this PR to match the style of the rest of simple_message.

would that be acceptable to you @marip8 ?

motoman_driver/include/motoman_driver/simple_message/messages/motoman_select_tool_message.h
marip8: I see what you mean. I was thinking that this class was thread-safe in the sense that it does not hold pointers to objects it doesn't own (unlike the motion control class with its connection pointer).

motoman_driver/include/motoman_driver/simple_message/motoman_select_tool.h
marip8: That's fine with me. Passing arguments by non-const reference unnecessarily is generally not a great idea, so I just wanted to make a note

motoman_msgs/srv/SelectTool.srv
marip8: Good points. I'm fine with leaving it in the `srv` then.

motoman_driver/include/motoman_driver/simple_message/motoman_select_tool.h
marip8: There's nothing technically wrong here, so I'm okay with keeping it in the interest of merging proven and valuable functionality without adding too much additional effort

marip8:
> As to using a separate connection: that would require changes to MotoROS, and quite substantial ones.

I didn't realize there wasn't an abundance of ports available on the controller. It seemed like we just incremented the port number for the IO relay, so I thought we might be able to do the same here if it made sense. 

Since the tool is tied closely with execution, I'm on board with keeping this functionality in the motion control class

motoman_driver/include/motoman_driver/simple_message/motoman_select_tool.h
gavanderhoorn: > Passing arguments by non-const reference unnecessarily is generally not a great idea, so I just wanted to make a note

Yes, agreed.

And please keep doing that. In-depth reviews are very much appreciated.

gavanderhoorn:
> I didn't realize there wasn't an abundance of ports available on the controller. It seemed like we just incremented the port number for the IO relay, so I thought we might be able to do the same here if it made sense.

no, sockets are available. That's not the problem.

Changing the tool frame on the controller however has two sides, and while one of those actions can easily be delegated to a separate task, the other needs to mutate state that would be part of the `MotionController`. So we'd avoid locks on the `motoman_driver` side, only to have to introduce them on the MotoROS side (if we want to do it right).

gavanderhoorn:
I've tested this on our YRC1000 as well.

No regressions detected.

Requests for changing to tool files outside the valid range are refused as they should.

riv-rbush:
@gavanderhoorn I'm sorry for the silence - commercial activities exploded just after I sent that message offering to test it! This is awesome that you guys decided to get on and implement the enhancement. Massive thanks. 

What (if any) testing do you still need on this?

gavanderhoorn:
I believe we've sufficiently tested this and it seems to work.

If you want, you could build the driver with this PR included and try to invoke the service.

gavanderhoorn:
@marip8: I believe I've applied all your suggestions.

Could you take a final look?

CI is *very* slow today, but if you're OK with these changes and CI is green, we'll merge.

marip8 (approved):

gavanderhoorn:
thanks for the review @marip8 :+1:

PR #437, opened 2021-12-12
driver: move message data structure wrappers to 'messages' sub dir

gavanderhoorn:
Noticed while working on #436 and #435: all `_message.(cpp|h)` wrappers should be in their respective `messages/` sub directories.

Some weren't, so move them their.

No functional changes.


gavanderhoorn:
Perhaps @steviedale could (also) take a look?

marip8 (approved):

gavanderhoorn:
Thanks @marip8 :+1:

PR #438, opened 2021-12-12
driver: add missing MotionReply sub codes

gavanderhoorn:
As per subject.

Noticed they were missing while working on #435 and #436.

This will only improve the "human side" of `motoman_driver`. The error (sub) codes were processed correctly before this PR, but the human readable error message just converted them to "unknown" errors.

### DIFF
--- a/motoman_driver/src/simple_message/motoman_motion_reply.cpp
+++ b/motoman_driver/src/simple_message/motoman_motion_reply.cpp
@@ -162,7 +162,9 @@ std::string MotionReply::getSubcodeString(shared_int code)
   case MotionReplySubcodes::NotReady::SKILLSEND:
     return "Waiting on SkillSend";
   case MotionReplySubcodes::NotReady::PFL_ACTIVE:
-    return "PFL is active";
+    return "Power and Force Limiting (PFL) was activated: "
+      "please reset PFL on the robot, re-enable the driver (call the service) "
+      "and send a new trajectory";
   case MotionReplySubcodes::NotReady::INC_MOVE_ERROR:
     return "Incremental move rejected on MotoROS side";
 
--- a/motoman_driver/src/simple_message/motoman_motion_reply.cpp
+++ b/motoman_driver/src/simple_message/motoman_motion_reply.cpp
@@ -134,6 +134,10 @@ std::string MotionReply::getSubcodeString(shared_int code)
     return "Invalid acceleration data";
   case MotionReplySubcodes::Invalid::DATA_INSUFFICIENT:
     return "Insufficient trajectory data.  Must supply valid time, pos, and velocity fields.";
+  case MotionReplySubcodes::Invalid::DATA_TIME:
+    return "Invalid time data: exceeded maximum trajectory duration of 14400 sec (ros-industrial/motoman#244)";
+  case MotionReplySubcodes::Invalid::DATA_TOOLNO:
+    return "Invalid tool file specified";
 
   case MotionReplySubcodes::NotReady::UNSPECIFIED:
     return "Unknown";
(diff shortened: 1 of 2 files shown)


gavanderhoorn:
Perhaps @steviedale could (also) take a look?

motoman_driver/src/simple_message/motoman_motion_reply.cpp
```diff
@@ -157,6 +161,10 @@ std::string MotionReply::getSubcodeString(shared_int code)
     return "Waiting on ROS";
   case MotionReplySubcodes::NotReady::SKILLSEND:
     return "Waiting on SkillSend";
+  case MotionReplySubcodes::NotReady::PFL_ACTIVE:
+    return "PFL is active";
```
marip8: If I remember correctly, PFL stands for power and force limiting and this error means that the driver can't be used when the robot is in PFL mode, correct? If so we might change this message to something more informative like:

```suggestion
    return "Power and force limiting (PFL) mode is active; disable PFL to execute trajectories with the ROS driver";
```

marip8 (dismissed):
These changes look fine to me. I just had one comment about the PFL message we might want to address before merging.

motoman_driver/src/simple_message/motoman_motion_reply.cpp
gavanderhoorn: There are some nuances to this I believe.

Perhaps @EricMarcil can clarify?

EricMarcil: ROS can operate the robot with the PFL function enabled.  This message is when the PFL has detected a force that is above the allowed threshold (like someone pushing on the robot).  So we do not want to 'disable' the PFL but simply 'reset' the PFL, which is normally done by pressing the amber button on the robot B-link.
After that, ROS will need to reinitialize the trajectory mode (ROS_CMD_START_TRAJ_MODE = 200121) before sending a new trajectory from the robot current position.

marip8: So maybe a message like this would be better:

```suggestion
    return "Power and force limiting (PFL) has detected a force on the robot above the allowable threshold. Remove the obstruction and reset the PFL";
```

> After that, ROS will need to reinitialize the trajectory mode (ROS_CMD_START_TRAJ_MODE = 200121) before sending a new trajectory from the robot current position.

Is this something the user needs to do, or is it taken care of by the driver as a part of the trajectory execution action?

gavanderhoorn: New suggestion:

```suggestion
    return "Power and Force Limiting (PFL) was activated: "
      "please reset PFL on the robot, re-enable the driver (call the service) "
      "and send a new trajectory";
```

gavanderhoorn: > > After that, ROS will need to reinitialize the trajectory mode (ROS_CMD_START_TRAJ_MODE = 200121) before sending a new trajectory from the robot current position.
> 
> Is this something the user needs to do, or is it taken care of by the driver as a part of the trajectory execution action?

`motoman_driver` will try to enable trajectory mode upon reception of a new trajectory, but only if the driver has been "enabled" (by calling the `robot_enable` service) before.

It won't reset anything on the controller.

gavanderhoorn: Thanks for the input guys.

I've committed the new message for PFL.

gavanderhoorn:
As I dismissed the review with my last commit, we're going to need another approval before this can be merged.

EricMarcil (approved):
@gavanderhoorn and @marip8 Looks good to me.  
Thanks

gavanderhoorn:
Thanks for the reviews @marip8 and @EricMarcil :+1:

PR #439, opened 2021-12-12
driver: add support for MotoROS read & write M registers

gavanderhoorn:
As per subject.

Hasn't been tested on HW yet.

---

Edit: and just repeating https://github.com/ros-industrial/motoman/issues/361#issuecomment-709018080 here for future readers:

> While convenience is why we're adding them here, I'm going to repeat my advice (which I always give when talking about ROS, robot drivers and IO): consider using a proper fieldbus for industrial deployments. Having to add IO support to each and every robot driver doesn't scale very well, and will most likely use custom implementations of protocols which do not have the same level of support, integration nor testing as any of the available (software defined) fieldbuses in existence.
> 
> Even a simple wrapper around a modbus/tcp client would be a good option.


gavanderhoorn:
I expect conflicts with #435, but we'll resolve those when they happen.

akashjinandra:
@gavanderhoorn would you like some help testing on hardware?

gavanderhoorn:
@akashjinandra: sure, the more testing this gets the better.

You should just have to build `motoman_driver` with this PR merged and make sure to have a recent-ish MotoROS installed on the controller (as in: anything `>=` 1.9.2).

Keep in mind that I haven't tested this at all yet, so it could not work, or it could work immediately, or anything in between.

akashjinandra (dismissed):
Hello @gavanderhoorn I tested this on a YRC1000 controller. Both the read and the write work well. I also tested reading out of range and writing out of range both threw errors, which was expected.

gavanderhoorn:
Thanks @akashjinandra :+1:

motoman_msgs/srv/ReadMRegister.srv
```diff
@@ -0,0 +1,18 @@
...
+# (on the teach pendant for instance).
+#
+# Only the following addresses can be read from:
+#
+#  - 0 to 999
+#
+# Refer also the Yaskawa Motoman documentation on IO addressing and
+# configuration.
+
+uint32 address
```
akashjinandra, line 21: @gavanderhoorn would it make more sense to instead make this a signed int16, techcnially you can't put anything larger than a 16bit number in these m registers. This would also then allow the user to put a negative number in that register as well.

EricMarcil: Yaskawa controller M-Registers are unsigned 16bit numbers.  Registers are mainly used for timers and counters, so you cannot enter negative values.

gavanderhoorn: We could potentially switch to `uint16`s here in the `.srv`.

The main reason why there are `uint32`s used right now is because the MotoROS SimpleMessage structure specs those:

https://github.com/ros-industrial/motoman/blob/7c648e84f623d2c9f8cd640a052efe6ee9c73110/motoman_driver/MotoPlus/SimpleMessage.h#L365-L369

Similar to #436, we could use a narrower type in the `.srv` to prevent users from trying to use values `> 2^16` and then discovering that doesn't work.

akashjinandra: I understand that the m registers technically are supposed to be unsigned 16 bit numbers, we were working with Yaskawa in Europe, and they had us write a negative number in a register essentially by using 2's compliment in the binary form.

gavanderhoorn: Yes, that's how you can do this.

It would not feel correct to me to spec `int16`s in the `.srv` here, as that would make it that much harder to use the full unsigned, 16 bit range for users who need that. It would also not be correct, as the M-registers are 16 bit unsigned.

And additionally, the "trick" you mention is a rather well-known work-around for getting "negative numbers" in an unsigned register/variable/something else.

akashjinandra:
Hello, anything stopping this from getting merged?

gavanderhoorn:
Tested on our YRC1000. Works. No regressions.

akashjinandra:
Hello is there anything I can do to help get this merged?

gavanderhoorn:
I've switched the `value` fields to `uint16`, as that's what M-registers are/support.

gavanderhoorn:
@marip8: could you please also take a look at this one?

gavanderhoorn:
I'll fix the conflicts here after #435 gets merged, as that PR touches the same files.

gavanderhoorn:
Friendly ping @marip8

marip8 (approved):

gavanderhoorn:
Resolved the conflict.

We'll merge if/when CI turns green.

gavanderhoorn:
CI was green, so merging.

Thanks @marip8, @EricMarcil and @ted-miller :+1:

And thank you @akashjinandra for testing.

PR #440, opened 2021-12-13
MotoROS v1.9.9 - Select tool used for jogging

ted-miller:
What ROS_MSG_MOTO_SELECT_TOOL is received, select the tool used for jogging in addition to specifying which tool is used by the motion API.


ted-miller:
@EricMarcil @gavanderhoorn 
This is here if you want it...

motoman_driver/MotoPlus/MotionServer.c
```diff
@@ -1968,14 +1968,29 @@ int Ros_MotionServer_GetDhParameters(Controller* controller, SimpleMsg* replyMsg
 
 int Ros_MotionServer_SetSelectedTool(Controller* controller, SimpleMsg* receiveMsg, SimpleMsg* replyMsg)
 {
+#if !(DX100 || FS100)
```
gavanderhoorn: hm, so this doesn't work on DX100 nor FS100?

motoman_driver/MotoPlus/MotionServer.c
```diff
@@ -1968,14 +1968,29 @@ int Ros_MotionServer_GetDhParameters(Controller* controller, SimpleMsg* replyMsg
...
+
+#if !(DX100 || FS100)
+			//set jogging tool on the pendant
+			setToolData.sRobotNo = controller->ctrlGroups[receiveMsg->body.selectTool.groupNo]->groupId;
+			setToolData.sToolNo = tool;
+			mpSetToolNo(&setToolData, &responseData);
+#endif
+
+			//We don't care if mpSetToolNo fails. It won't affect the actual motion.
 			Ros_SimpleMsg_MotionReply(receiveMsg, ROS_RESULT_SUCCESS, 0, replyMsg, groupNo);
```
gavanderhoorn, line 1995: I'm not sure about this: if the service is supposed to make the controller "change the active tool file", and it fails to do that, it would seem like the caller should be notified of this.

How else would (s)he figure out it didn't work?

motoman_driver/MotoPlus/MotionServer.c
ted-miller: This is a new-ish API that apparently wasn't back-ported to previous controllers. (It's not included in the header files anyway.)

motoman_driver/MotoPlus/MotionServer.c
ted-miller: My thinking is that a set-tool command from ROS should be primarily concerned with how it affects motion being commanded from ROS. That part is going to be successful as long as the tool number is valid. 

We can try to also set the jogging tool. But, the process shouldn't be stopped if that doesn't work for some arbitrary reason.  If the user then goes to pick up the pendant, the active tool number is displayed in the top status bar. So they can see what the current selection is at any time.
image

motoman_driver/MotoPlus/MotionServer.c
gavanderhoorn: For FS100 in `v1.3.0` it's in `inc/mpMotCtrl.h`:

```c
extern int mpMotSetTool(int grpNo, int toolNo);
```
I believe?

ted-miller: moMotSetTool is used for a different motion API. The correct function is `mpSetToolNo`. 
However, I did check this again. Turns out that it is available for DX100, but is not available for FS100. I have updated this in a new commit.

motoman_driver/MotoPlus/MotionServer.c
EricMarcil: I agree with Ted.  
Thought it might seem strange to some, the Yaskawa controller manages the jog tool and the motion tool independently.  This is true for normal (none-ROS) operation.  It is so that  if you set your tool to manually jog the robot and then run some job, after when you want to jog the robot again, the jogging tool is the same as the one that was used before running the job.

motoman_driver/MotoPlus/MotionServer.c
gavanderhoorn: Oh right, apologies. I confused the two. Part of the legacy API for newer controllers it seems.

motoman_driver/MotoPlus/MotionServer.c
gavanderhoorn: Ok, if you both agree then let's do it like this.

Is "motion tool" official terminology? If it is, we could use that to give the service added in #436 a better name.

gavanderhoorn:
Just something which crossed my mind: would it make sense to only allow this service to be invoked when the robot is *not* executing some motion? Or would that be too restrictive / not necessarily safer?

motoman_driver/MotoPlus/MotionServer.c
gavanderhoorn: So just to be clear: is this PR now changing the behaviour from "only setting the motion tool" to "setting the motion tool and the jogging tool"? If so, from what you write @EricMarcil, that doesn't sounds like it would be desirable.

EricMarcil: No, there is no "official" terminology for it.  Manuals just make reference to tool in either case.

gavanderhoorn: > No, there is no "official" terminology for it. Manuals just make reference to tool in either case.

Ok.

Then it would seem prudent to *clearly* document what is going to be updated when users invoke the `SelectTool` service from ROS.

ted-miller:
> would it make sense to only allow this service to be invoked when the robot is not executing some motion?

It is allowable to change the active tool during motion. If you have a gripper that picks something up, it can (will) change the mass properties of the tool. Perhaps you would want to do that in between trajectory requests... but I don't see why you couldn't also do it on-the-fly.

gavanderhoorn:
> It is allowable to change the active tool during motion. If you have a gripper that picks something up, it can (will) change the mass properties of the tool. 

Of course.

> Perhaps you would want to do that in between trajectory requests... but I don't see why you couldn't also do it on-the-fly.

I was just thinking about someone calling this service at an inopportune moment. But we'd have the same problem with other services.

EricMarcil:
Technically, the tool information is part of the incremental data.  So we might want to limited it while a motion is being broken down to incremental data.

EricMarcil:
After checking the code, it should be fine the way it is because we make a local copy of the tool:
`incData.tool = ctrlGroup->tool;`
at the beginning of the function `Ros_MotionServer_JointTrajDataToIncQueue` which decomposes the last motion to smaller increments.  So even if the tool is change while processing the last joint motion, it will not affect it until the next joint motion is received.

motoman_driver/MotoPlus/MotionServer.c
gavanderhoorn: So just to be clear: is this PR now changing the behaviour from "only setting the motion tool" to "setting the motion tool and the jogging tool"? If so, from what you write @EricMarcil, that doesn't sounds like it would be desirable.

gavanderhoorn:
@EricMarcil: I've added a small comment in b90de0f which summarises your analysis (to help our future selves).

motoman_driver/MotoPlus/MotionServer.c
gavanderhoorn: Ok, from https://github.com/ros-industrial/motoman/pull/436#discussion_r768843669 it's clear we do want `mpSetToolNo(..)` here, so closing as *resolved*.

gavanderhoorn (approved):
Tested on our YRC1000. This works for me as well.

gavanderhoorn:
Thanks @ted-miller and @EricMarcil :+1:

gavanderhoorn:
@ted-miller @EricMarcil: will you update the wiki?

ted-miller:
wiki has been updated

riv-rbush:
@gavanderhoorn @ted-miller @EricMarcil 
Successfully switching tools in teach (aka. jogging) mode. Failing in remote mode. Are there permission settings on the controller (YRC1000) to allow tool switching in remote mode?

---
EDIT: I'm assuming in remote mode the motion API is called to change tool and the jogging tool is not changed hence why the Pendant UI remains static.

ted-miller:
Welp... it looks like my testing was inadequate. It seems this API can only be used in teach mode.

image

So, it seems that this PR was pointless.

gavanderhoorn:
I'm pretty sure I've tested this on our YRC1000 here, which was not in teach mode -- I "only" run the driver in play mode.

riv-rbush:
@gavanderhoorn Can you test again to be 100% sure? 

@ted-miller What does `2010` and `2080` relate to? Something we can change locally?

ted-miller:
Those are just error codes. These are the values that will be in the `err_no` field that is returned by the API call.
0x2010 = "Robot is in operation"
0x2080 = "Wrong operation mode"

riv-rbush:
@ted-miller @gavanderhoorn 
I tested with tool interference zones and the FSU robot range function. I confirm it **works in play mode**. FYI: The Pendant tool selection UI does not update. 

---
Thanks for the PR guys, this is an awesome feature for production line systems.

ted-miller:
That is expected. The tool change will take effect for motion-control and safety zones. 

The error codes I mention are specifically regarding the "teach" tool that is displayed on the pendant UI.

riv-ljames:
Hello @ted-miller, do you know if it's possible to get the "Robot range display" 3D view on the pendant to accurately show the tool selected via motoros? Right now we find that it sometimes goes "out of sync" with what the motoros driver thinks is on the end of the robot, which can make debugging very confusing.
I suspect this is an unavoidable side effect of the driver using the mpSetToolNo API in a way which it does not explicitly support... But if you have any suggestions they would be greatly appreciated :pray: 

Thanks.

PR #448, opened 2022-01-25
MotoROS: Clear motion queue after alarm

ted-miller:
If alarm occurs while queue is full and there is pending data to be processed, then subsequent trajectory commands will fail.


ted-miller:
Waiting for confirmation from the user that this resolves the issue

gavanderhoorn:
Could you please update the commit comment to include the cause of the issue this is fixing?

It's not clear from the changes why they fix the issue. Having a description of how this change fixes it would be valuable.

ted-miller:
> It's not clear from the changes why they fix the issue

Well... that's probably because it's total nonsense...

As I attempted to write up a description, it made me think about this in more detail. And now I've started second guessing my original assumptions. 

I need to setup some testing.

ted-miller:
My initial message is definitely wrong. But, the commit does fix a problem. I'm going to scrap this PR, make sure the actual issue gets resolved with the user, and then try another PR.

PR #449, opened 2022-01-26
MotoROS: Improve clearing of internal motion queue during error states

ted-miller:
Note that this does not actually resolve any current user issues. It was discovered while investigating a different issue that is not caused by MotoROS.


gavanderhoorn (dismissed):
Haven't run this on our YRC1000, but the proposed changes look OK.

gavanderhoorn:
Have you tested this on HW @ted-miller ?

ted-miller:
Yes

gavanderhoorn (approved):

gavanderhoorn:
Thanks @ted-miller.

Would you update the wiki?

PR #411, opened 2021-05-27
Added sia30d support package

EricMarcil:
Added sia30d support package

PR #453, opened 2022-02-02
sia30: use materials from resources package

gavanderhoorn:
As per subject.

#411 was submitted before `motoman_resources` was created, so it didn't yet use the `material` definitions from  that package.

No functional changes.


gavanderhoorn:
@EricMarcil: could you take a look, as the original author of the package?

EricMarcil (approved):
That looks fine.  Thanks for the update.

gavanderhoorn:
thanks @EricMarcil :+1:

PR #476, opened 2022-03-14
Add GP110 support

cjue:
This adds a support package for the Motoman GP110 manipulator.

@EricMarcil: As far as I could gather from comparing the datasheets and technical drawings the GP110 is just about identical to the older MH110, apart from the DX200 controller for that one (and a different total weight?). Is this assumption correct?


EricMarcil:
@cjue It's possible that it is very similar to the MH110 (I haven't checked to confirm).  In any event, the attached file has the values for the GP110.

[Robot_Specs_GP110-A0.csv](https://github.com/ros-industrial/motoman/files/8248046/Robot_Specs_GP110-A0.csv)

cjue:
Thank you for the GP110 data, I updated the effort values in the URDF accordingly.

I found my link_6_t mesh was in the wrong place and that the collision meshes were corrupted, this should all be fixed now.

The meshes total about 4MB in size now.

cjue:
I used Blender Decimate to further shrink the biggest links, meshes/ is down to 3 MB now.

@gavanderhoorn Is that size fine with you?

motoman_gp110_support/config/opw_parameters_gp110.yaml
```diff
@@ -0,0 +1,20 @@
...
+#
+opw_kinematics_geometric_parameters:
+    a1:  0.320
+    a2:  -0.235
+    b:   0.000
+    c1:  0.540
+    c2:  0.870
+    c3:  1.020
+    c4:  0.200
+opw_kinematics_joint_offsets: [0.0, 0.0, deg(-90.0), 0.0, 0.0, deg(180.0)]
```
gavanderhoorn, line 19: You've probably checked this, but that `deg(180.0)` at the end is unusual.

Or at least: it's the first time I see it in an OPW param file.

motoman_gp110_support/urdf/gp110_macro.xacro
```diff
@@ -0,0 +1,169 @@
...
+      <origin xyz="0 0 0" rpy="0 0 0" />
+      <axis xyz="-1 0 0" />
+      <limit lower="${radians(-360)}" upper="${radians(360)}" effort="823.76" velocity="${radians(255)}"/>
+    </joint>
+    <!-- end of joint list -->
+
+    <!-- ROS-Industrial 'flange' frame: attachment point for EEF models -->
+    <link name="${prefix}flange"/>
+    <joint name="${prefix}joint_6_t-flange" type="fixed">
+      <origin xyz="0.200 0 0" rpy="${radians(180)} ${radians(-90)} 0"/>
```
gavanderhoorn: Is this correct? `flange` is supposed to have X+ 'forward' (ie: pointing out of the flange link).

Should this rotation be applied to `tool0` instead?

motoman_gp110_support/test/launch_test.xml
```diff
@@ -0,0 +1,51 @@
+<!--
```
gavanderhoorn: please rename this file to include the variant name (ie `launch_test_gp110.xml`).

motoman_gp110_support/package.xml
```diff
@@ -0,0 +1,55 @@
...
+      <b>Specifications</b>
+    </p>
+    <ul>
+      <li>GP110 - Default</li>
+    </ul>
+    <p>
+      Joint limits and maximum joint velocities are based on the information 
+      found in the online documentation files for the robot (GP110_2D.zip
+      and GP110_3D.zip from
+      https://www.yaskawa.eu.com/products/robots/handling-mounting/productdetail/product/gp110_704 )
```
gavanderhoorn: Is there really no document which provides this information? A datasheet / manual?

If you can, please mention document (file)name and revision, as direct links tend to go stale.

motoman_gp110_support/package.xml
```diff
@@ -0,0 +1,55 @@
...
+      unless noted otherwise.
+    </p>
+    <p>
+      Before using any of the configuration files and / or meshes included
+      in this package, be sure to check they are correct for the particular
+      robot model and configuration you intend to use them with.
+    </p>
+  </description>
+  <author>Christian Juelg</author>
+  <maintainer email="christian.juelg@optonic.com">Christian Juelg (Optonic GmbH)</maintainer>
```
gavanderhoorn, line 32: I will call you out if there are ever any issues with this package in the future. Just so you know ;)

motoman_gp110_support/launch/test_gp110.launch
```diff
@@ -0,0 +1,9 @@
+<launch>
+  <include file="$(find motoman_gp110_support)/launch/load_gp110.launch" />
+
+  <node name="joint_state_publisher_gui" pkg="joint_state_publisher_gui" type="joint_state_publisher_gui" />
+
+  <node name="robot_state_publisher" pkg="robot_state_publisher" type="robot_state_publisher" />
+
+  <node name="rviz" pkg="rviz" type="rviz" args="-d $(find industrial_robot_client)/config/robot_state_visualize.rviz" required="true" />
```
gavanderhoorn: please remove the empty lines.

gavanderhoorn (commented):
Firs round of comments.

gavanderhoorn (changes requested):
Changing to `request-changes` to avoid accidentally merging it.

motoman_gp110_support/launch/test_gp110.launch
cjue: Do you mean only the two between the nodes, not the one after the include?

gavanderhoorn: The three nodes you mean?

Yes, that would work.

motoman_gp110_support/package.xml
cjue: `Flyer_Robot_GP110_E_11.2020.pdf` is the closest thing to document name and revision I could find

cjue:
> Is this correct? flange is supposed to have X+ 'forward' (ie: pointing out of the flange link).
> Should this rotation be applied to tool0 instead?

Yes, it appears there is the same mistake in the gp4 macro:
https://github.com/ros-industrial/motoman/blob/kinetic-devel/motoman_gp4_support/urdf/gp4_macro.xacro#L148

The hc10dt_b10 macro does it the right way:
https://github.com/ros-industrial/motoman/blob/kinetic-devel/motoman_hc10_support/urdf/hc10dt_b10_macro.xacro#L156

gavanderhoorn:
Seems that was missed in #433.

Please remove 7e5caf6 from this PR.

We should not touch multiple packages in a PR introducing a new package.

cjue:
> Please remove https://github.com/ros-industrial/motoman/commit/7e5caf65bdaa3e74e45ee27181701c9388caccf0 from this PR.

done, also rebased to the current `kinetic-devel`

motoman_gp110_support/package.xml
```diff
@@ -0,0 +1,55 @@
...
+      <b>Specifications</b>
+    </p>
+    <ul>
+      <li>GP110 - Default</li>
+    </ul>
+    <p>
+      Joint limits and maximum joint velocities are based on the information 
+      found in the online documentation files for the robot (CAD model GP110_SP100_3D.zip
+      and drawing datasheet HW1385705.01 from
+      https://www.yaskawa.eu.com/products/robots/handling-mounting/productdetail/product/gp110_704 )
```
gavanderhoorn:
```suggestion
      and drawing datasheet <em>HW1385705.01</em> from <a href="https://www.yaskawa.eu.com/products/robots/handling-mounting/productdetail/product/gp110_704">here</a>).
```

gavanderhoorn (approved):
I believe this looks OK now.

Thanks for the PR and the follow-up @cjue :+1:

gavanderhoorn:
I'm going to squash-merge this as well, to avoid the fixups (and non-decimated meshes) from ending up in the history.

gavanderhoorn:
Thanks again @cjue

Issue #282, opened 2019-07-05
mh50_support urdf:  incorrect link_3_u joint limits?


cjue:
Hi,

it looks to me like there is an error for one of the joint limits in the MH50 URDF: https://github.com/ros-industrial/motoman/blob/kinetic-devel/motoman_mh50_support/urdf/mh50_macro.xacro#L125
```
<limit lower="-1.39626" upper="3.59538" effort="2745.86" velocity="3.106"/>
```

The specification of the MH-50 II here says the limits are +251/-170°, not +205/-80°:

https://www.motoman.com/en-us/products/robots/industrial/assembly-handling/mh-series/mh50-ii

Is this an error, or was this correct for an earlier MH50 variant?

gavanderhoorn:
Afaik, the MH-50 II and the model supported by `motoman_mh50_support` are not necessarily the same robot.

I'm not entirely sure though, so I'll wait for @EricMarcil to confirm this.

EricMarcil:
The values in the model are correct.  They are defined relative to the previous joint (L-axis).
The specification you are making reference to is the possible orientation of the U-axis relative to the ground which also depends on the L-axis possible position, therefore yielding to a larger range.

MuhongGuo:
@EricMarcil @cjue I think the reason is that it is impossible for the U-axis to reach lower degree than -80 because of physical interference from the motor on L-axis. Please reference to the image below, where the U-axis is at the position of -90 degree and it is already in collision. I've also tried to change the position of L-axis but couldn't find any condition to let U-axis reach the range as per the specs. It's odd that the official website seems providing this unrealistic range.

![image](https://user-images.githubusercontent.com/17403324/68639595-fd5c8900-053f-11ea-9c89-370427adb4f3.png)

EricMarcil:
@MuhongGuo : I wouldn't call it unrealistic, it's just a different reference than what you were expecting.  Like I said, that reference is relative to the ground or the horizontal line.  If the L-axis is at -90 deg and the U-axis -80 deg relative to the L-axis, that puts the U-axis at -170 deg relative to the horizontal.  Marketing people don't always think like engineers...  That being said, the specifications for the new series like the GP series, are represented as you would expect it in reference to the previous link.

MuhongGuo:
> If the L-axis is at -90 deg and the U-axis -80 deg relative to the L-axis, that puts the U-axis at -170 deg relative to the horizontal. 

[...]

PR #477, opened 2022-03-21
Add MH110 support (predecessor of GP110 with DX200 controller)

cjue:
As far as the datasheets tell MH110 and GP110 are identical except for some kg of total weight and DX200 vs YRC1000 controller.

The URDF references the GP110 meshes for the collision models, and package.xml lists motoman_gp110_support as a dependency.

@EricMarcil Could you please check if the joint effort limits are the same for GP110 and MH110? 

I couldn't find any 3D CAD model for the MH110, but it appears to be identical to the GP110.

This MR build upon #476

### DIFF
--- a/motoman_mh110_support/launch/test_mh110.launch
+++ b/motoman_mh110_support/launch/test_mh110.launch
@@ -1,0 +1,9 @@
+<launch>
+  <include file="$(find motoman_mh110_support)/launch/load_mh110.launch" />
+
+  <node name="joint_state_publisher_gui" pkg="joint_state_publisher_gui" type="joint_state_publisher_gui" />
+
+  <node name="robot_state_publisher" pkg="robot_state_publisher" type="robot_state_publisher" />
+
+  <node name="rviz" pkg="rviz" type="rviz" args="-d $(find industrial_robot_client)/config/robot_state_visualize.rviz" required="true" />
+</launch>
--- a/motoman_mh110_support/launch/test_mh110.launch
+++ b/motoman_mh110_support/launch/test_mh110.launch
@@ -2,8 +2,6 @@
   <include file="$(find motoman_mh110_support)/launch/load_mh110.launch" />
 
   <node name="joint_state_publisher_gui" pkg="joint_state_publisher_gui" type="joint_state_publisher_gui" />
-
   <node name="robot_state_publisher" pkg="robot_state_publisher" type="robot_state_publisher" />
-
   <node name="rviz" pkg="rviz" type="rviz" args="-d $(find industrial_robot_client)/config/robot_state_visualize.rviz" required="true" />
 </launch>
--- a/motoman_mh110_support/package.xml
+++ b/motoman_mh110_support/package.xml
@@ -19,8 +19,7 @@
       Visual appearance, joint limits and maximum joint velocities are 
       identical to the successor GP110 model, the only differences being
       joint effort limits and DX200 vs YRC1000 controller connection.
-      Information on the MH110 can be found in <em>HW1482741 (169859-1CD Rev. 2)</em>
-      at the <a href="https://www.motoman.com/en-us/service-training/product-documentation">Yaskawa Motoman product documentation site</a>.
+      Information on the MH110 can be found in <em>HW1482741 (169859-1CD Rev. 2)</em>.
       The GP110 online documentation files (CAD model GP110_SP100_3D.zip and 
       drawing datasheet <em>HW1385705.01</em>) can be found <a href="https://www.yaskawa.eu.com/products/robots/handling-mounting/productdetail/product/gp110_704">here</a>.
     </p>
--- a/motoman_mh110_support/package.xml
+++ b/motoman_mh110_support/package.xml
@@ -19,8 +19,12 @@
       Visual appearance, joint limits and maximum joint velocities are 
       identical to the successor GP110 model, the only differences being
       joint effort limits and DX200 vs YRC1000 controller connection.
+      Information on the MH110 can be found in <em>HW1482741 (169859-1CD Rev. 2)</em>
+      at the <a href="https://www.motoman.com/en-us/service-training/product-documentation">Yaskawa Motoman product documentation site</a>.
       The GP110 online documentation files (CAD model GP110_SP100_3D.zip and 
-      drawing datasheet <em>HW1385705.01</em> can be found <a href="https://www.yaskawa.eu.com/products/robots/handling-mounting/productdetail/product/gp110_704">here</a>).
+      drawing datasheet <em>HW1385705.01</em>) can be found <a href="https://www.yaskawa.eu.com/products/robots/handling-mounting/productdetail/product/gp110_704">here</a>.
+    </p>
+    <p>
       All urdfs are based on the default motion and joint velocity limits, 
       unless noted otherwise.
     </p>
--- a/motoman_mh110_support/package.xml
+++ b/motoman_mh110_support/package.xml
@@ -45,7 +45,10 @@
 
   <exec_depend>joint_state_publisher_gui</exec_depend>
   <exec_depend>motoman_driver</exec_depend>
+
+  <!-- this package reuses the meshes included in motoman_gp110_support -->
   <exec_depend>motoman_gp110_support</exec_depend>
+
   <exec_depend>motoman_resources</exec_depend>
   <exec_depend>robot_state_publisher</exec_depend>
   <exec_depend>rviz</exec_depend>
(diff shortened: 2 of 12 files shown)


EricMarcil:
@cjue The robot 3D model are identical.  There is a slight different for the effort on some axis.  See attached file: 
[Robot_Specs_MH110-A0.csv](https://github.com/ros-industrial/motoman/files/8315394/Robot_Specs_MH110-A0.csv)

cjue:
Thanks, I updated all the values. The support package should be complete then.

motoman_mh110_support/launch/test_mh110.launch
```diff
@@ -0,0 +1,9 @@
+<launch>
+  <include file="$(find motoman_mh110_support)/launch/load_mh110.launch" />
+
+  <node name="joint_state_publisher_gui" pkg="joint_state_publisher_gui" type="joint_state_publisher_gui" />
+
+  <node name="robot_state_publisher" pkg="robot_state_publisher" type="robot_state_publisher" />
+
+  <node name="rviz" pkg="rviz" type="rviz" args="-d $(find industrial_robot_client)/config/robot_state_visualize.rviz" required="true" />
```
gavanderhoorn: Please remove empty lines.

motoman_mh110_support/package.xml
```diff
@@ -0,0 +1,55 @@
...
+    <p>
+      <b>Specifications</b>
+    </p>
+    <ul>
+      <li>MH110 - Default</li>
+    </ul>
+    <p>
+      Joint limits and maximum joint velocities are based on the information 
+      found in the online documentation files for the mostly identical successor GP110 model from
+      https://www.yaskawa.eu.com/products/robots/handling-mounting/productdetail/product/gp110_704 )
```
gavanderhoorn: Same question/suggestion here as in #476.

motoman_mh110_support/test/launch_test.xml
```diff
@@ -0,0 +1,51 @@
+<!--
```
gavanderhoorn: Please also rename this file to include the variant name.

motoman_mh110_support/urdf/mh110_macro.xacro
```diff
@@ -0,0 +1,169 @@
+<?xml version="1.0" ?>
+<robot xmlns:xacro="http://ros.org/wiki/xacro">
+  <xacro:macro name="motoman_mh110" params="prefix">
+    <xacro:include filename="$(find motoman_resources)/urdf/common_materials.xacro"/>
+
+    <!-- link list -->
+    <link name="${prefix}base_link">
+      <visual>
+        <geometry>
+          <mesh filename="package://motoman_gp110_support/meshes/gp110/visual/base_link.stl"/>
```
gavanderhoorn, line 11: Could you please add a comment here explaining why the MH110 package is using files from the GP110 package?

Just one here should suffice. No need to repeat it on the other `mesh` elements.

motoman_mh110_support/urdf/mh110_macro.xacro
```diff
@@ -0,0 +1,169 @@
...
+      <origin xyz="0 0 0" rpy="0 0 0" />
+      <axis xyz="-1 0 0" />
+      <limit lower="${radians(-360)}" upper="${radians(360)}" effort="823.76" velocity="${radians(255)}"/>
+    </joint>
+    <!-- end of joint list -->
+
+    <!-- ROS-Industrial 'flange' frame: attachment point for EEF models -->
+    <link name="${prefix}flange"/>
+    <joint name="${prefix}joint_6_t-flange" type="fixed">
+      <origin xyz="0.200 0 0" rpy="${radians(180)} ${radians(-90)} 0"/>
```
gavanderhoorn: Please check whether the rotation here should be moved to the `tool0` joint instead.

motoman_mh110_support/package.xml
```diff
@@ -0,0 +1,55 @@
...
+  <url type="bugtracker">https://github.com/ros-industrial/motoman/issues</url>
+  <url type="repository">https://github.com/ros-industrial/motoman</url>
+
+  <buildtool_depend>catkin</buildtool_depend>
+
+  <test_depend>roslaunch</test_depend>
+
+  <exec_depend>joint_state_publisher_gui</exec_depend>
+  <exec_depend>motoman_driver</exec_depend>
+  <exec_depend>motoman_gp110_support</exec_depend>
```
gavanderhoorn, line 53: Could you add a comment here explaining why this `exec_depend` exists?

gavanderhoorn (changes requested):
Some in-line comments.

gavanderhoorn:
Just realised the squash-merge in #476 made merging this a bit more difficult.

@cjue: you should be able to drop just about every commit here, and squash e092b34 & fbc8857 together.

cjue:
Rebased onto kinetic-devel after #476 merge.

I updated the package.xml regarding the differences of MH110/GP110. 
Unfortunately I couldn't find any official Yaskawa URLs or downloads for the MH110.

gavanderhoorn:
> Unfortunately I couldn't find any official Yaskawa URLs or downloads for the MH110.

If you search for `motoman MH110`, you'll end up on the [Yaskawa Motoman product documentation](https://www.motoman.com/en-us/service-training/product-documentation) site. That lists `HW1482741` as the MOTOMAN-MH110 INSTRUCTIONS manual (`169859-1CD`), which lists the *Basic Specifications* of this particular robot in section `5.1`:

image

motoman_mh110_support/urdf/mh110_macro.xacro
```diff
@@ -0,0 +1,170 @@
...
+      <origin xyz="0.320 0 0" rpy="0 0 0" />
+      <axis xyz="0 1 0" />
+      <limit lower="${radians(-90)}" upper="${radians(155)}" effort="11645.40" velocity="${radians(110)}"/>
+    </joint>
+    <joint name="${prefix}joint_3_u" type="revolute">
+      <parent link="${prefix}link_2_l"/>
+      <child link="${prefix}link_3_u"/>
+      <origin xyz="0 0 0.870" rpy="0 0 0" />
+      <axis xyz="0 -1 0" />
+      <limit lower="${radians(-80)}" upper="${radians(90)}" effort="4167.83" velocity="${radians(130)}"/>
```
gavanderhoorn, line 121: The limits in `169859-1CD` (see https://github.com/ros-industrial/motoman/pull/477#issuecomment-1075285083) are different: `[-185; 160]`.

@EricMarcil: would you know what causes the difference?

gavanderhoorn: Oh, is that the "limit wrt ground plane vs actual joint limits"?

cjue: Thanks for the link. 

The specifications table uses a weird definition for the U-Axis limits, somehow combining the valid range of motion for the L and U Axis.

EricMarcil: Yes, that would be the case.  The manual limits make reference to the upper arm position relative to the ground which is actually a combination of both L and U axes.  So not useful for the URDF model...

motoman_mh110_support/package.xml
```diff
@@ -0,0 +1,64 @@
...
+    </p>
+    <ul>
+      <li>MH110 - Default</li>
+    </ul>
+    <p>
+      Visual appearance, joint limits and maximum joint velocities are 
+      identical to the successor GP110 model, the only differences being
+      joint effort limits and DX200 vs YRC1000 controller connection.
+      Information on the MH110 can be found in <em>HW1482741 (169859-1CD Rev. 2)</em>
+      at the <a href="https://www.motoman.com/en-us/service-training/product-documentation">Yaskawa Motoman product documentation site</a>.
```
gavanderhoorn:
```suggestion
      Information on the MH110 can be found in <em>HW1482741 (169859-1CD Rev. 2)</em>.
```

gavanderhoorn: It's clear this is a reference to a Yaskawa Motoman released document. That should be sufficient to find it again if needed.

motoman_mh110_support/urdf/mh110_macro.xacro
cjue: This came up before in discussion of the MH50: https://github.com/ros-industrial/motoman/issues/282#issuecomment-510516570

Looks to me as if the MH-series documentation generally used that notation.

gavanderhoorn (approved):
This looks OK now I believe.

Further fixups/corrections should be submitted in follow-up PRs.

gavanderhoorn:
Using squash-merge again here to avoid fixups getting into the repository history.

gavanderhoorn:
Thanks @cjue :+1:

PR #478, opened 2022-03-22
apply flange orientation fix to gp4

cjue:


gavanderhoorn:
I've temporarily disabled actions, as we're pushing quite a few commits to #476 and #477, spawning many runners each time.

gavanderhoorn (approved):
As discovered and discussed in #476.

gavanderhoorn:
thanks @cjue :+1:

PR #396, opened 2021-03-05
Add ar2010 support

Doomerdinger:
Added support for the AR2010.

Requires #378.

PR #480, opened 2022-03-22
Add support for AR2010 (rebased #396)

gavanderhoorn:
Finally had time to rebase this and add some fixups.

Original PR by @Doomerdinger in #396.

Commit provenance retained, rebased on-top of current `kinetic-devel`.

Note: I've not re-oriented the `joint`s, nor the `link`s. As such, this variant's `xacro:macro` deviates from all other models in this repository. I did add a note about this to the package's manifest.

Afaict, this PR is OK to merge like this.


gavanderhoorn:
Thanks again @Doomerdinger :+1:

PR #483, opened 2022-03-24
downgrade scary log message to INFO instead, provide default controller TCP port

cjue:
The `Failed to find topic_list parameter` error message is too aggressive as this is the normal case when there is just one arm connected.

The `Failed to get '~port' parameter: using default (50242)` warning can be avoided by just adding this default parameter explicitly to io_relay.launch.

Thanks to @gavanderhoorn for your feedback.


gavanderhoorn:
Closing as I believe #484 and #485 are improvements on the change(s) proposed here.

gavanderhoorn:
Thanks for submitting the PR @cjue. That's always appreciated :+1:

PR #484, opened 2022-03-25
Driver: add default port for IO server to launch file (alternative to #483)

gavanderhoorn:
Altered the implementation proposed in #483 slightly to not hard-code the default value, but expose it as a parameter.

This is a slight divergence from the other nodes in `motoman_driver` (as they don't support configurable port numbers), but the `io_relay` is a more recent addition, and does support this.

### DIFF
--- a/motoman_driver/launch/io_relay.launch
+++ b/motoman_driver/launch/io_relay.launch
@@ -5,6 +5,9 @@
   <!-- IP of robot (or PC running simulation) -->
   <arg name="robot_ip" doc="IP of controller" />
 
+  <!-- TCP port the IO server is listening on -->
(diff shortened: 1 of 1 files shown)


gavanderhoorn:
As mentioned in the commit comment: thanks @cjue for the initial work (in #483).

I preferred to use separate commits (and PRs) for the two disjoint changes proposed in that PR.

gavanderhoorn:
Let me know if this is an acceptable alternative @cjue.

cjue:
Looks good to me.

For everyone else, the motivation for this and #483 is to avoid the `Failed to get '~port' parameter: using default (50242)` warning that was previously always shown.

gavanderhoorn:
Merged without additional review, as I'd already reviewed #483 and @cjue vetted my adaptation.

PR #485, opened 2022-03-25
Improve WARN and INFO msgs when topic_list parameter isn't found

### LINKED ISSUE #113: driver: unneeded ROS_ERROR in robot_state_interface?
This may seem trivial, but I'm trying to eliminate as many ROS_ERROR messages as possible in my application, where remote ROS-newbies are using my software stack, and red error lines result in frequent issues back to me.  If people agree with my assessment below, I can create a patch, though it's a simple issue for anyone to change in a couple of minutes.

So, the ([Usuage Tutorial](http://wiki.ros.org/motoman_driver/Tutorials/Usage)) for motoman driver discusses needing the _controller_joint_names_ parameter (which works for the industrial-robot-sim and industrial-robot-client), but when using the motoman launch files, the motoman driver complains twice about a missing _topic_list_.   It looks like robot_state_interface and joint_trajectory_interface properly handles a missing _topic_list_ parameter (getJointGroup() fails) by using _controller_joint_names_, so then why the need to throw an ERROR in getJointGroup()?

I do see that the multi-group (non-version0) motoman_driver_joint_trajectory_action node requires the _topic_list_ parameter, but it's not even handling the return from getJointGroup().  So, the better solution would be to NOT put an error out to the user in getJointGroup() if _topic_list_ does not exist, but catch this and put an error out in motoman_driver_joint_trajectory_action for people that really do require this parameter for multigroup setups.

gavanderhoorn:
This fixes #113 -- and basically implements the suggestion there.

Original commit proposed by @cjue in #483. The version here is a slightly expanded version.

I also wanted to use separate commits and PRs for this change (and the other change in #483, which is now in #484).

The result of this change is that callers of `getJointGroups(..)` get to decide what to do/print/log when it can't find the `topic_list` parameter: print a warning or do something else.

Example output with these changes:

```
[ INFO] [...] [/joint_state]: Failed to find 'topic_list' parameter
[ INFO] [...] [/joint_state]: Expecting/assuming single motion-group controller configuration
[ INFO] [...] [/joint_state]: Adding joint_1_s to list parameter
[ INFO] [...] [/joint_state]: Adding joint_2_l to list parameter
...
```

As the commit comment explains though: the `WARN` is retained for the case where the multi-group version of the JTA is started, but the multi-group configuration cannot be found. That's still not necessarily an `ERROR`, as that node *can* be used for such cases, but it's definitely abnormal, and warrants a `WARN`.

### DIFF
--- a/motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
+++ b/motoman_driver/src/industrial_robot_client/joint_trajectory_action.cpp
@@ -60,7 +60,15 @@ JointTrajectoryAction::JointTrajectoryAction() :
   pn.param("constraints/goal_threshold", goal_threshold_, DEFAULT_GOAL_THRESHOLD_);
 
   std::map<int, RobotGroup> robot_groups;
-  getJointGroups("topic_list", robot_groups);
+  if (!getJointGroups("topic_list", robot_groups))
+  {
+    // this is a WARN as this class is the multi-group version of the regular JTA,
+    // and we're actually expecting to find the 'topic_list' parameter, as using
+    // this multi-group version with a single-group system is unnecessary and also
+    // doesn't make much sense.
+    // It probably also won't work.
+    ROS_WARN("Expecting/assuming single motion-group controller configuration");
+  }
 
   for (size_t i = 0; i < robot_groups.size(); i++)
   {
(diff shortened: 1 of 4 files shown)


gavanderhoorn:
Let me know if this is an acceptable alternative @cjue.

cjue:
Looks good to me, much more comprehensive than my initial proposed change.

gavanderhoorn:
Merging without additional review as I'd already reviewed #483 and @cjue vetted this one.

PR #452, opened 2022-02-01
Add gp200r support 

drolleigh:
Created a support package for gp200r including a urdf that has been compared to gp200r pendant read-outs of a real robot in unique poses. The translation was accurate under 1.0 mm and rotation was accurate under 1.0 degree.

gp200r_pendant_urdf_compare

PR #489, opened 2022-03-30
gp200r: add missing dependency on resources pkg

gavanderhoorn:
As per subject.

Seems I missed this in the review of #452.


marip8 (approved):

gavanderhoorn:
Thanks @marip8 :+1:

PR #506, opened 2022-06-10
compare joint names, reject joint_state updates with differing names

### LINKED ISSUE #326: motoman_driver accepting joint_states that are not from the robot arm controller
On my robot I have a gripper that is publishing its joint_states on the topic /my_gripper/joint_states and this information is relayed down the the /joint_states topic. I have recently found that I am randomly receiving errors from the industrial_robot_client stating:
`isWithinRange::Key vectors are not similar`
and my trajectory being rejected. 

I did some digging and found that in [joint_trajectory_interface.cpp](https://github.com/ros-industrial/motoman/blob/26d739bec0ed307557bdc5be07f0f30a04ac55a2/motoman_driver/src/industrial_robot_client/joint_trajectory_interface.cpp#L722) in the jointStateCB the software blindly accepts the incoming message on the /joint_states topic as the robots `cur_joint_pos_`. I added a simple check `if(this->all_joint_names_.size() != msg->position.size())`
then just return from the callback without updating the robots `cur_joint_pos_` and it fixed my errors. 

I am happy to put in a pull request with a fix for this error but wanted to get your guy's feedback on how you would like to ensure the received JointState message is for the robot arm we are connected to.
Should we check to make sure that all of the names in the `all_joint_names_` variable are included in the `msg->name` list? Should we check to make sure that the JointState message includes some minimum level of information (i.e. the name, position and velocity vectors all have the same length)? Anything else you would like it to check?
Another option would be to launch the node in a namespace and we would get /motoman/joint_states but users would be required to relay that topic down to /joint_states for the PlanningSceneMonitor and RobotStatePublisher.

cjue:
fixes https://github.com/ros-industrial/motoman/issues/326

based on https://github.com/gavanderhoorn/motoman/commit/e369ddc88e4a04adf847d9415ae94a0103ab9fd8 but also added the check to the multi-arm callback.

@gavanderhoorn in https://github.com/ros-industrial/motoman/issues/326#issuecomment-627420401
> I'm still not happy with having to do that here in this driver, but it does help make things more robust, so a PR with that change ported to current kinetic-devel would be something we can merge.

Note: this has not been tested on multi-arm setups.

### DIFF
--- a/motoman_driver/src/industrial_robot_client/joint_trajectory_interface.cpp
+++ b/motoman_driver/src/industrial_robot_client/joint_trajectory_interface.cpp
@@ -518,7 +518,6 @@ bool JointTrajectoryInterface::calc_velocity(const motoman_msgs::DynamicJointsGr
     return true;
   }
 
-  robot_groups_[pt.group_number].get_joint_names();
   for (size_t i = 0; i < robot_groups_[pt.group_number].get_joint_names().size(); ++i)
   {
     const std::string &jnt_name = robot_groups_[pt.group_number].get_joint_names()[i];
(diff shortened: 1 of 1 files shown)


cjue:
closed this PR by accident, re-opened it now

PR #518, opened 2022-08-24
Update MH5: Add Short & Long Variants (based on pull request #380)

cjue:
This is an updated version of pull request #380 by @acbuynak. 

I rebased everything to current kinetic-devel and made some additions:
* fix MH5 joint 6 offset: 80mm per datasheet and spec, same as for the MH5L variant
* add opw configs
* use materials from resources package
* use joint_state_publisher_gui for testing 

The most important difference is in the definition of joint_6_t for the MH5. According to the specs posted by @EricMarcil in https://github.com/ros-industrial/motoman/issues/377#issuecomment-742556746 the offset should be `0.080`, not `0.0865´. The old value was already present in the kinetic-devel version of motoman_mh5_support, but to my understanding it has always been wrong.

@acbuynak I could not open a clean pull request with my changes to https://github.com/OSU-AIMS/motoman/tree/kinetic-devel-update-mh5 because there was an extra "merge kinetic-devel" commit in there. Feel free to add my commits to your PR, we can then close this one.


EricMarcil:
The 0.0865 was the correct value for the original MH5 models on NX100 and DX100: MH00005.  They are the models with the upper arm casting being straight.
Then there was an optimization and they made the upper arm taper toward the end and the distance to the flange was reduce to 0.080.  Those are the models that normally have a letter after the 5 in the model identification: MH0005S, MH0005F.
Looking back at the package and its history, the package was made for the newer version in 2014 and the current package.xml does identify the correct model that the package supports.  However the flange offset in the urdf seems to have been wrong from the beginning, originally at 0.051 and then revised in 2017 to 0.0865 (maybe that person had an old MH5 model). 
All that being said, I'm confirming that given the mesh model, this package is for the MH5S or MH5F models and the correct value is 0.080.

acbuynak:
Thanks for bringing this one back! Happy to see it used. I'll take a closer look at this in the coming week.
@EricMarcil - Thanks for the model verification.

cjue:
> Thanks for bringing this one back! Happy to see it used. I'll take a closer look at this in the coming week.

Do you have feedback on the updated support package?

gavanderhoorn:
Just to get it confirmed (again): @EricMarcil: the change of the `joint_6_t` offset is correct for the variant modelled by the support package?

EricMarcil:
Yes, the changes for joint_6_t is correct for the models specified in the package.xml:
      <li>MH5  - Short Variants: MH5F(FS100),  MH5SII(DX200),  MH5S(DX100)</li>
      <li>MH5L - Long Variants:  MH5LF(FS100), MH5SLII(DX200), MH5SL(DX100)</li>

gavanderhoorn:
I've cleaned the commit history up a bit and added some more changes, see https://github.com/ros-industrial/motoman/compare/kinetic-devel...update_mh5_acbuynak.

It's mostly @acbuynak's commits (but squashed, to prevent fixups ending up in the history) and @cjue's additions plus one of my own.

As I can't push to either of the branches these PRs originated from (both this one and #380), I'll probably open a new one.

gavanderhoorn:
Closing as we now have #620.

Thanks again @cjue for your work :+1: 

And apologies for leaving it open for this long :disappointed:

PR #526, opened 2022-11-18
Renaming yrcmlib to correct yrcmLib filename extension.

fgrcar:
Because of the wrong filename extension, the MotoPlus IDE didn't link the library file into the output file and therefore the robot displayed the missing symbols error when executing the app. From my point of view, this is a terrible bug on which I spent days. Finally I was managed to resolve it when the local Yaskawa programmer pointed it out.


gavanderhoorn:
While technically correct, Windows is not case-sensitive, so I'm confused as to how this could be a problem.

It could be the compiler/linker used is case sensitive though.

fgrcar:
Yeah, I don't know, but until I changed it, even the IDE didn't show the library file under the library files' section. I realized that the `out` file is exactly that smaller in size as is the size of the library file, so I exactly knew what is wrong but wasn't able to fix it at the first place. Anyway, after the filename extension correction, it worked as expected.

ted-miller:
Can you please clarify whether you are using MotoPlusIDE or Visual Studio?

This is MPIDE:
image

ted-miller (approved):
Regardless of whether windows _should_ care about case, it is technically correct that the L should be capitalized.

gavanderhoorn (approved):

gavanderhoorn:
Thanks @fgrcar :+1:

fgrcar:
@ted-miller: MotoPlus IDE, yes, as I wrote in my first comment.
Thank you guys for such quick response, I'm glad to help.

PR #534, opened 2023-03-21
Fix ES200-120 joint S limit

marip8:
Changes per PR title and commit message


EricMarcil (approved):
It should have been -180.  It was probably a typo on the original commit.

PR #535, opened 2023-03-29
GP35L Support Package

EricMarcil:
Add GP35L Support Package


motoman_gp35l_support/config/opw_parameters_gp35l.yaml
gavanderhoorn, line 1: @EricMarcil: could you clarify where this OPW config file came from?

Did you verify it yourself?

gavanderhoorn (commented):
Looks OK. One inline question.

@EricMarcil: seeing as this appears to be an `L` variant: are there any other variants of the GP35? If so, the support package name should drop the `l`.

gavanderhoorn:
Something else I noticed:

image

the black motors and cable and white Yaskawa lettering are a nice touch, but as you can see, the black is so *black*, there is no detail visible at all like this.

Notice also the rendering artefacts on other parts of the geometry: could it be the meshes have been decimated/compressed too much?

EricMarcil:
@gavanderhoorn Changed the black to dark gray.  Also fixed the color of a few faces that got messed up by the decimated.  It does look as if the model was a bit too decimated but that's how my original source file was.  So we'll go with that for now.

ted-miller (approved):

cjue:
> seeing as this appears to be an `L` variant: are there any other variants of the GP35? If so, the support package name should drop the `l`.

Judging from this [overview](https://www.yaskawa.eu.com/Global%20Assets/Downloads/Brochures_Catalogues/Robotics/MOTOMAN_Robots/GP-Series/Yas_GP_Serie_Gesamt_E.pdf) there is no "GP35" besides the GP35L, so package name should be fine.

EricMarcil:
> seeing as this appears to be an L variant: are there any other variants of the GP35? If so, the support package name should drop the l.

On that note, most of the models ending by an L are hybrid robot with longer reach and have nothing to do with a base model with the same number.  The number in the model simply indicates the resulting payload of the hybrid model.  The GP35L is actually closer to a GP50 but with an extension in the upper arm to give it more reach but reduced payload to 35 kg.  If we were to include it in the GP50 support package, then I think most people would not find it.  So, I think that having a package named GP35L is the best solution.

PR #536, opened 2023-04-17
Fix for "within range" error

qweewqrty:
While using Motoman GP12 with ROS I regularly got problem: "Trajectory start position doesn't match current robot position". I found, that, after complete moveit trajectory, robot position have small fluctuation (within 0.005 rad) for couple of seconds. If moveit generate plan while position changing - error occured. I checked motoman code, that generated error message, and found that it belong to FS100 controller. But my GP12 had YRC1000. I disabled that check and my robot worked fine for hours. Seems like this check really needed for FS100, not YRC1000.
I wrote fix for that [roblem in this pull request.

### DIFF
--- a/motoman_driver/include/motoman_driver/joint_trajectory_streamer.h
+++ b/motoman_driver/include/motoman_driver/joint_trajectory_streamer.h
@@ -54,6 +54,8 @@ using industrial_robot_client::joint_trajectory_streamer::JointTrajectoryStreame
 using industrial::simple_message::SimpleMessage;
 using industrial::smpl_msg_connection::SmplMsgConnection;
 
+enum class MotomanControllerType { DX100, DX200, FS100, YRC1000 };
+
 /**
  * \brief Message handler that streams joint trajectories to the robot controller.
  *        Contains FS100-specific motion control commands.
@@ -134,6 +136,7 @@ public:
 
 protected:
   int robot_id_;
+  MotomanControllerType robot_controller_type_;
   MotomanMotionCtrl motion_ctrl_;
 
   // used to enforce serialisation of all access to the single, shared SmplMsgConnection.
--- a/motoman_driver/launch/robot_interface_streaming.launch
+++ b/motoman_driver/launch/robot_interface_streaming.launch
@@ -21,12 +21,16 @@
 	<!-- robot_ip: IP-address of the robot's socket-messaging server -->
 	<arg name="robot_ip" doc="IP of controller" />
 
+	<!-- robot_controller: type of robot controller used with robot -->
+	<arg name="robot_controller" doc="type of robot controller" />
+
 	<!-- Load the byte-swapping versions of Fanuc nodes if required -->
 	<arg name="use_bswap" doc="If true, robot driver will byte-swap all incoming and outgoing data" />
 
 	<!-- copy the specified parameters to the Parameter Server, for 
 	     use by nodes below -->
 	<param name="robot_ip_address" type="str" value="$(arg robot_ip)" />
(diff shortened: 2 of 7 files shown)


gavanderhoorn:
Thanks for the PR. That's always appreciated.

However:

>  I checked motoman code, that generated error message, and found that it belong to FS100 controller

could you clarify what made you conclude it "belong[s] to FS100 controller"?

ted-miller:
> I checked motoman code, that generated error message, and found that it belong to FS100 controller.

The comment in the code is misleading. This check should be performed for all controller models. 

I recommend monitoring the `in_motion` flag of the [RobotStatus](https://github.com/ros-industrial/industrial_core/blob/2b7ef6505d3a4742a35a33345a30b92d5c5e9080/industrial_msgs/msg/RobotStatus.msg#L25) message. This should indicate whether it is safe for MoveIt to plan the next trajectory

qweewqrty:
> could you clarify what made you conclude it "belong[s] to FS100 controller"?

Comment in the code said, that this check needed by FS100 controller.

qweewqrty:
> I recommend monitoring the `in_motion` flag of the [RobotStatus](https://github.com/ros-industrial/industrial_core/blob/2b7ef6505d3a4742a35a33345a30b92d5c5e9080/industrial_msgs/msg/RobotStatus.msg#L25) message. This should indicate whether it is safe for MoveIt to plan the next trajectory

Thanks, I'll check this way ASAP, and, if problem will persist, return here =)

gavanderhoorn:
> > could you clarify what made you conclude it "belong[s] to FS100 controller"?
> 
> Comment in the code said, that this check needed by FS100 controller.

Please always link to code or text when you make references to it. That makes our lives much easier. Otherwise we have to search.

EricMarcil:
Just to clarify why that code is there.  The MotoRos driver moves the robot by increments, so if you start moving the robot at the wrong place, it will end-up at the wrong place.  Without that start check, if you grabbed the wrong robot position (for example while the robot is still moving), plan a trajectory, and send it to the controller, then the robot could potentially move to an unexpected position.  We've had a few cases where we increase a bit the tolerence on the start motion, but do NOT completely remove it.

qweewqrty:
> > > could you clarify what made you conclude it "belong[s] to FS100 controller"?
> > 
> > 
> > Comment in the code said, that this check needed by FS100 controller.
> 
> Please always link to code or text when you make references to it. That makes our lives much easier. Otherwise we have to search.

Will do!

qweewqrty:
> 

It felt like changing tolerance didn't do anything. I set it up to 10 rad and still got an "within range" error. And when I thought, that check didn't important for my controller - that had some kind of sense. Now I think, I'll do some more debugging, to understand, what's going on.

gavanderhoorn:
> It felt like changing tolerance didn't do anything. I set it up to 10 rad and still got an "within range" error. And when I thought, that check didn't important for my controller - that had some kind of sense. Now I think, I'll do some more debugging, to understand, what's going on.

are you saying you always get the error "Trajectory start position doesn't match current robot position"?

If that's the case, *that* would be something to diagnose.

But the changes proposed in your PR are not a correct solution.

If you feel you are getting that error while "everything should just work", please open an issue instead, so we can try and find out what's going on with your system.

gavanderhoorn:
With that said, I'm going to close the PR for now.

To diagnose any problems with the traj-start-pos checking logic, please open an issue.

If it turns out we *do* need to fix something, we could re-open this PR.

@qweewqrty: in any case, thank you for submitting the PR. That is always appreciated.

PR #500, opened 2022-05-24
Update GP7 support package (OPW params, missing links, collision mesh update)

cjue:
* add opw file
* fix link2 collision mesh
* add flange link
* update reference documentation
* minor fix for velocity limits
* rename and refactor test

PR #493, opened 2022-04-13
Update GP8 support package (OPW params, missing links, collision mesh update) 

cjue:
* add opw file
* fix link2 collision mesh
* add flange link
* update reference documentation
* minor fix for velocity limits

PR #553, opened 2023-06-15
Motoman GP70L Support

EricMarcil:
Add support package for Yaskawa/Motoman GP70L robot


motoman_gp70l_support/config/opw_parameters_gp70l.yaml
gavanderhoorn, line 1: Could you clarify whether this OPW config was created by you @EricMarcil?

Did you also verify it?

gavanderhoorn (commented):
One in-line question.

Something I noticed: the meshes in this PR have the same 'issue' with black parts lacking in contrast as in #535:

image

the cable is much nicer here though.

motoman_gp70l_support/config/opw_parameters_gp70l.yaml
EricMarcil: @gavanderhoorn OPW were added by @cjue on PR #493 and #500.  I simply used those as a base and updated the values.  They have not been verified.
I'm not sure if that it something we want to standardize or not.  We can remove the file, if you think it is better not to have them.

EricMarcil:
@gavanderhoorn I started importing model data from our MotoSim software.  That already have color, smaller models and model for cables.
I'll try to adjust the color of the motors...

motoman_gp70l_support/config/opw_parameters_gp70l.yaml
gavanderhoorn: I welcome the addition of OPW configuration files. I just want to make sure we have validated the configuration and are sure the values are correct.

EricMarcil:
@gavanderhoorn I fixed the motor color by making it dark gray.  It took a while because every time I tried making changes with blender I would get strange results.  I finally realized that the .dae was actually a text file and edited the color directly in it.
image

ted-miller (approved):

PR #566, opened 2023-09-02
Add GP20HL Support Package

acbuynak:
Add support package for Yaskawa Motoman GP20HL robot. 

If requested, I can draft a GP20 URDF model but I would not be able to verify on hardware. The GP20 model appears to be less common and maybe only distributed in EU?

This support package and URDF have been tested with real hardware and the joint directions verified. All joint motion in RViz match physical motion.


gavanderhoorn:
This might have been duplicated in #604.

gavanderhoorn:
@acbuynak: thank you for opening the PR and for making the effort to share the model.

As mentioned by @EricMarcil in https://github.com/ros-industrial/motoman/pull/604#issuecomment-2627375051:

> No, it is a new one that I made. I must have missed the #566 that was made by @acbuynak.
I used the MotoSim model to generate the visual meshes in .dae which keeps the colors of the orignial model. So, I would prefer to keep #604.

We'll close this one in favour of #604 then.

I'll compare the files in this PR against those in #604, just to see if there are any discrepancies.

motoman_gp20_support/urdf/gp20_hl_macro.xacro
```diff
@@ -0,0 +1,173 @@
...
+    <!-- end of link list -->
+
+    <!-- joint list -->
+    <!-- Joints are specified such that the driver rotates the joints about the positive Z axis. -->
+    <joint name="${prefix}joint_1_s" type="revolute">
+      <parent link="${prefix}base_link"/>
+      <child link="${prefix}link_1_s"/>
+      <origin xyz="0 0 0.540" rpy="0 0 0" />
+      <axis xyz="0 0 1" />
+      <limit lower="-${radians(180)}" upper="${radians(180)}" effort="212.66" velocity="${radians(125)}"/>
```
gavanderhoorn, line 108: All of the velocity limits in this file appear to be different from the model in #604.

@EricMarcil: would you have any idea why that could be?

EricMarcil: @gavanderhoorn  No, I do not. I've double checked the values on #604 with two separate sources: the manual and the parameter data from MotoSim, and the values on #604 are the correct ones.

acbuynak: Looked over my project history and don't see any reason why the 125 deg/sec velocity was set. I agree that the datasheet I pulled down today reports 180 deg/sec that @EricMarcil used in his PR. 

I'd recommend closing this PR. Thanks for the review @gavanderhoorn

gavanderhoorn: Thank you both @acbuynak and @EricMarcil.

@acbuynak: thank *you* for submitting the PR.

Given the state of things I'll close this one now.

PR #541, opened 2023-04-25
Add support for HC10DTP

EricMarcil:
Added support for HC10DTP (1-06VXCP10-B0*)

PR #573, opened 2023-11-17
Add support for HC10DTP - rebase

cjue:
Same as https://github.com/ros-industrial/motoman/pull/541 by @EricMarcil  but based on kinetic-devel with a linear history


ted-miller (approved):

PR #582, opened 2024-01-16
GP215 and GP250 Support Package

EricMarcil:
Add motoman_gp215-250_support which contains the GP215 and GP250 variant.


gavanderhoorn:
Is the `GP250` a variant of the `GP215`? If so, it would be better to name the package `motoman_gp215_support` and be clear about the fact it also supports the `GP250` variant in the manifest.

That would be how we do it for other variants.

ROS package names are not allowed to have dashes in them, so the current name would not be legal.

gavanderhoorn (changes requested):
Marking as *changes requested* until we resolve the naming issue

EricMarcil:
> Is the `GP250` a variant of the `GP215`? If so, it would be better to name the package `motoman_gp215_support` and be clear about the fact it also supports the `GP250` variant in the manifest.
> 
> That would be how we do it for other variants.
> 
> ROS package names are not allowed to have dashes in them, so the current name would not be legal.

It is not clear of the GP215 is a variant of the GP250 or the other way around. The main different is that the GP215 has a longer upper arm and a reduces payload of 215.
Other package have variant like GP25 and GP25-12, so it is easier to think about looking for the GP25-12 variant in the GP25 package. 
But in the case where the names (GP215 vs GP250) don't make reference to the original model name, users will not think of looking for the GP250 model in the GP215 package. So, I thought including both reference in the package name would make it esier. However, I was not aware that package should not contain '-' characters. We can change that for an underscore '_'.

gavanderhoorn:
> But in the case where the names (GP215 vs GP250) don't make reference to the original model name, users will not think of looking for the GP250 model in the GP215 package. So, I thought including both reference in the package name would make it esier.

thing is, with the dash, the `-250` looks like a variant of the `GP215`, if we follow the naming scheme.

With the underscore, it's even more unclear imo.

My suggestion would be we create a separate package for the `GP250`, in which we refer to the meshes of the `GP215` wherever possible.

That way we increase discoverability of the `GP250`, while still not duplicating everything.

We'd basically split out the files you made for the `250` and put them in their own package, refer to `motoman_gp215_support/meshes` in the `_macro.xacro` for all links that are identical to those in the `215` and make the `215` package an `exec_depend` of the `250` package (in other words: anyone who installs/uses the `250` package will automatically also get the `215` package).

---

Edit: if you can let me know whether that'd be acceptable, I can quickly split the package into two.

EricMarcil:
@gavanderhoorn I don't have a problem with what you are suggesting. I'm not familiar on the exec_depend and how exactly to set that up.
The other option would be to simply duplicate also the meshes. Size and download speed is not much of an issue now a day compare to 15 years ago, when we started creating the first packages.

motoman_gp215_250_support/package.xml
```diff
@@ -0,0 +1,60 @@
...
+      <li>GP215 - Default (YR-1-06VX215-A00, YR-1-06VX215-A20)</li>
+	  <li>GP250 - Default (YR-1-06VX250-A00, YR-1-06VX250-A20)</li>
+   </ul>
+    <p>
+      Joint limits and maximum joint velocities are based on the information 
+      found in:
+      <em>YASKAWA MOTOMAN-GP215,GP250 INSTRUCTIONS (HW2480629)</em>
+      version <em>194804-1CD, rev 1</em>.
+	  <em>YASKAWA MOTOMAN-GP215,GP250,GP280 INSTRUCTIONS (HW1484400)</em>
+      version <em>182639-1CD, rev 2</em>.
```
gavanderhoorn: @EricMarcil: could you clarify which of these documents you used?

Both, for both robots, or the first for the 215 and the second for the 250? Something else?

EricMarcil: I used both for both robots. The information affecting URDF model is identical in both.
The difference is that 182639-1CD is for the standard model -A00 and the 194804-1CD is for the -A20 which is the Dust-Proof/Water-Proof version with higher IP rating.

gavanderhoorn:
@EricMarcil: I've split out the GP250 from the GP215 package. The former depends on the latter for the meshes.

I feel this is the best compromise, given their relationship.

> The other option would be to simply duplicate also the meshes. Size and download speed is not much of an issue now a day compare to 15 years ago, when we started creating the first packages.

true, but if by spending a little extra effort we can avoid it, we might as well just do it.

gavanderhoorn:
We'll squash-merge all of this in the end, so as to avoid having all these fixups end up in the repository history.

EricMarcil:
@gavanderhoorn Thanks for splitting this up. Everything looks good.

gavanderhoorn (approved):

gavanderhoorn:
Ok.

We have green CI and @EricMarcil's approval, so (squash) merging.

gavanderhoorn:
Thanks @EricMarcil :+1:

PR #583, opened 2024-01-16
Add support for GP225 robot

EricMarcil:
Add support for Yaskawa Motoman GP225 robot


motoman_gp225_support/package.xml
```diff
@@ -0,0 +1,58 @@
...
+    <p>ROS-Industrial support for the Motoman GP225.</p>
+    <p>
+      This package contains configuration data, 3D models and launch files
+      for Motoman GP225 series of manipulators..
+    </p>
+    <p>
+      <b>Specifications</b>
+    </p>
+    <ul>
+      <li>GP225 - Default</li>
```
gavanderhoorn: Similar to #607, the `xacro:macro` appears to model what *Table 5-1: Basic Specifications (YR-1-06VX225-A00)* labels as *Not-equipped*:

image

Should we update this as well:

```suggestion
      <li>GP225 - Not-equipped</li>
```

?

motoman_gp225_support/package.xml
```diff
@@ -0,0 +1,58 @@
...
+    <p>
+      Before using any of the configuration files and / or meshes included
+      in this package, be sure to check they are correct for the particular
+      robot model and configuration you intend to use them with.
+    </p>
+    <p>
+      NOTE: the base model xacro macro has all joints oriented such that they
+      rotate around the Z+ axis. As a consequence, link frames in this model
+      do not follow REP-103 wrt frame orientation.
+    </p>
```
gavanderhoorn: This does not appear to be true for this package.

Copy-pasta?

motoman_gp225_support/meshes/gp225/visual/link_6_t.dae
gavanderhoorn, line 1: This mesh gets rendered with an unexpected greenish washed out colour for me:

rviz_screenshot_2025_01_31-14_07_04

Should this not be black?

Meshes in general look like they might be downsampled a bit too much.

gavanderhoorn (changes requested):

gavanderhoorn:
Friendly ping @EricMarcil

motoman_gp225_support/package.xml
EricMarcil: You're right, that paragraph should be removed. It probably came from using another model as a base to build this one.

motoman_gp225_support/package.xml
EricMarcil: After further reflection, I don't think we should start doing equipped variant. The equipped variant has extra brackets and  cables running outside of the arm which is not included in the 3D model and I don't have the 3D models for that optional package. It is mainly for SpotWelding application which I don't think is used much among ROS users.

motoman_gp225_support/meshes/gp225/visual/link_6_t.dae
EricMarcil: The yellow-greenish color is correct (it is some more resistant coating that is done on the flange). But there should be that much shadow, the original model surface lost it flatness probably too much downsampling. I found higher resolution of the same part on a different model but more details. I'll see if I can improve using that model tomorrow.

cjue: I just added an opw config to my fork of this PR here:
https://github.com/isys-vision/motoman/tree/GP225

That branch is rebased to kinetic-devel as well.

One peculiar thing that I noticed:

The support structure that connects from link2 to link1 does not move as it would in reality, causing it (and the collision model of that link) to "protrude" in some joint angles, where actually it would stay connected to link1. 

I guess this can be ignored similarly to the cables attachments of many robots that are not realistically modeled, I just noticed it when checking the visual&collision meshes.
Screenshot_20250223_035643_gp225_link2_attachment

https://www.yaskawa.eu.com/Global%20Assets/Downloads/Brochures_Catalogues/Robotics/MOTOMAN_Robots/GP-Series/Flyer_Robot_GP225_E_11.2020.pdf

EricMarcil: @cjue I pushed a new commit in which the side balancer is moves using the urdf mimic option. It is not perfect, the real equation is non-linear (sin, cos elements) and the mimic function only support a linear function.  So, I had to linearize the function over the L-axis range. It gives an error of about +/- 2 degrees (about 20 mm) at the end of the motion range.

I also included it in the collision model, but those model will collide with each other, so maybe it would be better remove them.

EricMarcil: image

cjue: Wow, nice work!

I don't think the overlap of the two parts poses an issue, and the result is much better than the old version that could "poke out" at certain J2 angles.

One note for actual usage in MoveIt: it might be necessary to explicitly disable collision detection between link_1_s/link_2_s and both new links link_1_s_2 and link_2_s_2, and possible collisions of the two new links with each other (5 combinations total).

cjue: I updated my fork of this PR to include your new commit, with the following additions
* add OPW config
* based on current ros-industrial/motoman kinetic-devel, while your branch is based on an older version
* remove trailing spaces in the urdf

https://github.com/isys-vision/motoman/tree/GP225

motoman_gp225_support/urdf/gp225_macro.xacro
```diff
@@ -0,0 +1,212 @@
...
+        <axis xyz="0 0 1" />
+        <limit lower="${radians(-180)}" upper="${radians(180)}" effort="7845.3" velocity="${radians(120)}"/>
+    </joint>
+    <joint name="${prefix}joint_1_s_2" type="revolute">
+        <parent link="${prefix}link_1_s"/>
+        <child link="${prefix}link_1_s_2"/>
+        <origin xyz="0.3989 0 -0.2576" rpy="0 0 0" />
+        <axis xyz="0 1 0" />
+        <limit lower="${radians(-60)}" upper="${radians(76)}" effort="11522.8" velocity="${radians(97)}"/>
+        <mimic joint="${prefix}joint_2_l" multiplier="0.7634"/>
```
cjue: trailing space

motoman_gp225_support/urdf/gp225_macro.xacro
```diff
@@ -0,0 +1,212 @@
...
+        <axis xyz="0 1 0" />
+        <limit lower="${radians(-60)}" upper="${radians(76)}" effort="11522.8" velocity="${radians(97)}"/>
+    </joint>
+    <joint name="${prefix}joint_2_l_2" type="revolute">
+        <parent link="${prefix}link_2_l"/>
+        <child link="${prefix}link_2_l_2"/>
+        <origin xyz="0 0 0.720" rpy="0 0 0" />
+        <axis xyz="0 1 0" />
+        <limit lower="${radians(-60)}" upper="${radians(76)}" effort="11522.8" velocity="${radians(97)}"/>
+        <mimic joint="${prefix}joint_2_l" multiplier="-0.2366"/>
```
cjue: trailing space

motoman_gp225_support/meshes/gp225/visual/link_6_t.dae
cjue: I opened a second PR in case it helps regarding the rebase issue:
https://github.com/ros-industrial/motoman/pull/619

EricMarcil: @cjue 

> One note for actual usage in MoveIt: it might be necessary to explicitly disable collision detection between link_1_s/link_2_s and both new links link_1_s_2 and link_2_s_2, and possible collisions of the two new links with each other (5 combinations total).

That why I was wondering if we shouldn't simply remove  link_1_s_2 and link_2_s_2 from the collision model. But I guess end-users can make that decision.

EricMarcil: > I opened a second PR in case it helps regarding the rebase issue: #619

I'm not sure of the rebase issue you are mentionning, but I'm fine with using the #619 since it already has you commits on it.

cjue: @EricMarcil 

> That why I was wondering if we shouldn't simply remove link_1_s_2 and link_2_s_2 from the collision model. But I guess end-users can make that decision.

Now that I had another look at it: the collision model of link_2_l_2 appears to be completely inside of that of link_2_s, which is quite large because it also contains the cabling in that area. link_1_s_2 on the other hand is not fully redundant.

My proposal would therefore be to remove just link_2_l_2 from the collision model, which would reduce the extra collisions that might need to be manually disabled to just 2 instead of 5.

I added a commit for that to my fork.

gavanderhoorn:
Ok, so I took a look and I believe we've arrived at something we can merge.

@cjue: thanks for your updates and for #619. As @EricMarcil already had a PR and to not make things more complex than they need to be I've cleaned up the history a bit here and included the relevant commits from #619.

(note btw: 32628f57 technically broke the `xacro:macro`, as empty `collision` elements are not supported)

I'll wait for CI to finish and then we can merge.

gavanderhoorn (approved):

gavanderhoorn:
As this PR contains commits by multiple contributors we'll use a regular merge commit.

gavanderhoorn:
Thanks @EricMarcil and thanks again @cjue :+1:

PR #597, opened 2024-08-15
HC20DTP and HC30PL Update

EricMarcil:
The HC20 package was updated with the HC20DTP and HC30PL variant. HC20 visual meshes were changed from .stl to .dae format to improve rendering.


gavanderhoorn:
High-level comment: could you please use multiple commits in future PRs when changing multiple things and adding support for new variants/series?

motoman_hc20_support/package.xml
```diff
@@ -2,13 +2,13 @@
 <?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="2">
   <name>motoman_hc20_support</name>
-  <version>0.1.0</version>
+  <version>1.0.0</version>
```
gavanderhoorn: We'll address this in a later PR, but we can't bump the version of this single package like this.

Catkin/Bloom limits things such that every package in a repository must have the same version.

Please revert this change and keep `version` at `0.1.0`.

motoman_hc20_support/package.xml
```diff
@@ -2,13 +2,13 @@
...
   <description>
     <p>ROS-Industrial support for the Motoman HC20 (and variants).</p>
     <p>
       This package contains configuration data, 3D models and launch files
       for Motoman HC20 manipulators. This currently includes the base model
-      which also supports the DT and XP variants.
+      which also supports the DT, XP, DTP and HC30PL variants.
```
gavanderhoorn: we normally reserve this kind of phrasing for variants which don't get their own `xacro:macro`.

The DTP and 30PL do have their own `xacro:macro`, so I would recommend we change this to reflect that.

EricMarcil:
@gavanderhoorn Yes, I will try do split them up in the future.

motoman_hc20_support/package.xml
```diff
@@ -17,11 +17,17 @@
       <li>HC20 - Default</li>
       <li>HC20DT - Default</li>
       <li>HC20XP - Default</li>
+      <li>HC20DTP</li>
+      <li>HC30PL</li>
```
gavanderhoorn, line 21: Seeing as we noticed there are actual variants of variants in other recently merged support packages: is there nothing we can state here as to which specific specification were used?

motoman_hc20_support/urdf/hc20dtp_macro.xacro
```diff
@@ -0,0 +1,159 @@
...
+    </joint>
+    <!-- end of joint list -->
+
+    <!-- ROS base_link to Robot Manufacturer World Coordinates transform -->
+    <link name="${prefix}base" />
+    <joint name="${prefix}base_link-base" type="fixed">
+      <origin xyz="0 0 0.380" rpy="0 0 0"/>
+      <parent link="${prefix}base_link"/>
+      <child link="${prefix}base"/>
+    </joint>
```
gavanderhoorn, line 167: should we perhaps take this opportunity to add `flange` and `base` here?

Same comment for the 30PL below.

motoman_hc20_support/urdf/hc30pl_macro.xacro
```diff
@@ -0,0 +1,163 @@
...
+      <origin xyz="0 0 0.820" rpy="0 0 0" />
+      <axis xyz="0 -1 0" />
+      <limit lower="${radians(-67)}" upper="${radians(247)}" effort="642.88" velocity="${radians(120)}"/>
+    </joint>
+    <joint name="${prefix}joint_4_r" type="revolute">
+      <parent link="${prefix}link_3_u"/>
+      <child link="${prefix}link_4_r"/>
+      <origin xyz="0.880 0 0" rpy="0 0 0" />
+      <axis xyz="-1 0 0" />
+      <limit lower="${radians(-15)}" upper="${radians(15)}" effort="368.48" velocity="${radians(112)}"/>
```
gavanderhoorn: Is this correct? 15 degrees only?

motoman_hc20_support/urdf/hc30pl_macro.xacro
```diff
@@ -0,0 +1,163 @@
...
+    </joint>
+    <joint name="${prefix}joint_5_b" type="revolute">
+      <parent link="${prefix}link_4_r"/>
+      <child link="${prefix}link_5_b"/>
+      <origin xyz="0 0 0" rpy="0 0 0" />
+      <axis xyz="0 -1 0" />
+      <limit lower="${radians(-15)}" upper="${radians(15)}" effort="141.12" velocity="${radians(132)}"/>
```
gavanderhoorn: same here

motoman_hc20_support/urdf/hc30pl_macro.xacro
```diff
@@ -0,0 +1,163 @@
...
+          <mesh filename="package://motoman_hc20_support/meshes/hc20/collision/link_4_r.stl"/>
+        </geometry>
+      </collision>
+    </link>
+    <link name="${prefix}link_5_b">
+      <visual>
+        <geometry>
+          <mesh filename="package://motoman_hc20_support/meshes/hc20dtp/visual/link_5_b.dae"/>
+        </geometry>
+        <origin rpy="0 1.57075 0"/>
```
gavanderhoorn, line 77: why do these Collada meshes have a different orientation?

It would be nice to have the meshes aligned already, so as to simplify the `xacro:macro`.

Same comment for the `collision` of `link_5_b` and both meshes for `link_6_t`.

gavanderhoorn (changes requested):
Marking as *requested changes* while we discuss my comments.

motoman_hc20_support/urdf/hc20dtp_macro.xacro
EricMarcil: Yes, that is a good suggestion

motoman_hc20_support/urdf/hc30pl_macro.xacro
EricMarcil: The HC30PL is an HC20 for Palletizing, so the wrist is always kept down.  This is what allows to manipulate 30 kg instead of 20 kg. But I'm realizing that the 15 deg in the manual is relative to the ground, not the previous link. For the R-axis, I believe that the 15 deg is correct because it doesn't combine with other axes.

motoman_hc20_support/urdf/hc30pl_macro.xacro
EricMarcil: The HC30PL is an HC20 for Palletizing, so the wrist is always kept down. This is what allows to manipulate 30 kg instead of 20 kg. But I'm realizing that the 15 deg in the manual is relative to the ground, not the previous link. The B-axis range probably needs to be reviewed to allow more motion, because the L and U axes are working in the same plane. I don't think there is any way in a URDF to put at motion restriction on the B-axis relative to the ground or by considering the other axes current position.

motoman_hc20_support/urdf/hc30pl_macro.xacro
EricMarcil: Because the HC30PL cannot have the B-axis in the horizontal position, so its zero position is with the wrist pointing down. I didn't want add an extra collada mesh just to turn it 90 degrees when it can be done here inthe xacro.

motoman_hc20_support/package.xml
EricMarcil: Ok, I didn't realize that it was impacting other things. We can revert.

motoman_hc20_support/urdf/hc30pl_macro.xacro
gavanderhoorn: hm. As we don't have any way to model these kinds of constraints (`joint` limits are always relative to parent `joint`s/`link`s), I'm wondering what the best course of action would be.

motoman_hc20_support/urdf/hc30pl_macro.xacro
gavanderhoorn: > I don't think there is any way in a URDF to put at motion restriction on the B-axis relative to the ground or by considering the other axes current position.

no, there currently isn't. See my previous comment.

motoman_hc20_support/urdf/hc30pl_macro.xacro
gavanderhoorn: Ok, that makes sense. Thanks for clarifying. We could perhaps add your comment to the `xacro:macro` as a comment.

motoman_hc20_support/urdf/hc20dtp_macro.xacro
gavanderhoorn: @EricMarcil: would you be able to to add those `link`s?

motoman_hc20_support/package.xml
EricMarcil: Reverted to 0.1.0

motoman_hc20_support/package.xml
EricMarcil: CHanged to:
This package contains configuration data, 3D models and launch files for Motoman HC20, HC20DTP and HC30PL manipulators. The HC20 base model also includes the DT, XP variants.

motoman_hc20_support/package.xml
EricMarcil: For those robots, there are only one version specified in the referenced manuals for each model.

motoman_hc20_support/urdf/hc20dtp_macro.xacro
EricMarcil: Added for all HC20, HC20DTP and HC30PL

motoman_hc20_support/urdf/hc30pl_macro.xacro
EricMarcil: The B-axis range was increased back to +/-180 deg. I added a note about it in the package.xml:
```
       NOTE: The HC30PL is the palletizing variant of the HC20 where the 
      wrist must always point toward the ground +/-15 degrees, which allows 
      a greater payload of 30 kg. There is no way of building this constraint
      in the urdf model. Therefore, this constraint will need to be handled 
      explicitly by the motion planner.
```

cjue:
In our repo we had added (identical) opw configs for HC20 DT and HC20 DTP:

https://github.com/isys-vision/motoman/commit/f2f4c61176f787a7a907aeead063e6859dfc1e55

We also renamed the hc20 files to hc20dt, which you probably don't want to copy:

https://github.com/isys-vision/motoman/commit/f10f3e612f21da1a93672537decd523a99db8fcb

EricMarcil:
@cjue I've added the opw_parameters files based on your files. Can you please double check the one for the hc30pl that I added. I believe that the -90 offset for the 5th joint is the only change required.

cjue:
> @cjue I've added the opw_parameters files based on your files. Can you please double check the one for the hc30pl that I added. I believe that the -90 offset for the 5th joint is the only change required.

MoveIt accepts the opw config for the HC30 PL, I could move the robot and the OPW correctly corresponds to the URDF file.

About the HC30 urdf file:

> The B-axis range probably needs to be reviewed to allow more motion, because the L and U axes are working in the same plane

Can you explain that a bit more? Currently the URDF in the PR allows +/-15° for joint_4_r, matching the [datasheet](https://www.yaskawa.eu.com/Global%20Assets/Downloads/Brochures_Catalogues/Robotics/MOTOMAN_Robots/HC30PL/Flyer_Cobot_HC30PL_E.pdf), while allowing +/- 180° for joint_5_b, where the datasheet would also dictate +/-15°.

Won't the controller abort any trajectory that moves outside the +/- 15 ° range for J5?

EricMarcil:
The +/- 15 degrees is relative to the floor.
This is an example where the B-axis is within its 15 degrees limits relative to the floor, but the B-axis position relative to the previous joint is around 217 deg

image

But you are correct, I'll need to correct the B-axis range. I forgot to account for the 90 deg offset. So the range is -90 to 270.

EricMarcil:
I've updated the hc30pl based on the limits from the parameter files. So the new values will be correct.

The user will need to insure that the robot flange remains +/- 15 degrees relative to the ground.

I haven't tested in MotoRos but normally the controller will give an alram 4420 SPECIAL SOFTLIMIT [R1: MAX B] at execution time when the robot is commanded to move outside of the 15 deg position.

cjue:
Sounds reasonable :+1: 

It's quite confusing that the datasheet differs this much from the URDF and the parameter files without specifying that the +/- 15° constraint is relative to the floor.

motoman_hc20_support/urdf/hc30pl_macro.xacro
gavanderhoorn: 'the wrist' means both `joint_4_r` and `joint_5_b`, correct?

gavanderhoorn: Assuming *yes*, I've added 83a611a308563f715264dddf42333bb76d8fe012.

motoman_hc20_support/config/opw_parameters_hc20dtp.yaml
```diff
@@ -0,0 +1,20 @@
...
+# Ortho-parallel Basis and a Spherical Wrist" by Mathias Brandstötter, Arthur
+# Angerer, and Michael Hofbaur (Proceedings of the Austrian Robotics Workshop
+# 2014, 22-23 May, 2014, Linz, Austria).
+#
+# The moveit_opw_kinematics_plugin package provides such a solver.
+#
+opw_kinematics_geometric_parameters:
+    a1:  0.000
+    a2:  0.000
+    b:   0.000
```
gavanderhoorn, line 14: @EricMarcil @cjue: before e7a2ee7d this read `-0.0`. I'm assuming that was copy-pasta and removed the `-` sign.

motoman_hc20_support/package.xml
```diff
@@ -7,26 +7,43 @@
...
     </p>
     <p>
-      All urdfs are based on the default motion and joint velocity limits, 
+      All urdfs are based on the default motion and joint velocity limits,
       unless noted otherwise.
     </p>
+    <p>
+      NOTE: the HC30PL is the palletizing variant of the HC20 where the wrist
+      joints ('joint_4_r' and 'joint_5_b') are limited to +/-15 degrees with
+      respect to the ground plane normal.
```
cjue: I understand this constraint to mostly limit the wrist / tool0 / flange orientation as a combination of joint_4_r and joint_5_b.

```suggestion
      NOTE: the HC30PL is the palletizing variant of the HC20 where the wrist
      joints ('joint_4_r' and 'joint_5_b') are constrained to orient the flange to 
      within +/-15 degrees with respect to the ground plane normal.
```

gavanderhoorn: given the zero pose of those two joints is aligned with the ground plane normal, would there be a difference?

I'm also not sure we can orient a surface in the way you describe. We'd either have to align the two normals -- ground plane and flange normal -- or say something about the flange surface having to be orthogonal to the ground plane normal, which isn't really clear.

cjue: You're right, I clarified that the `flange normal` is to be aligned to the ground plane normal.

```suggestion
      NOTE: the HC30PL is the palletizing variant of the HC20 where the wrist
      joints ('joint_4_r' and 'joint_5_b') are constrained to orient the flange normal 
      to within +/-15 degrees with respect to the ground plane normal.
```

gavanderhoorn: as I wrote, as the zero pose of the joints already causes the flange normal to be aligned with the ground plane normal, I believe it'd be ok to say something about the joint limits -- which is easier to understand I believe than surface normals.

But not a native speaker, so perhaps @ted-miller and/or @EricMarcil could take a look?

ted-miller: I'm ok with cjue's wording.

EricMarcil: I think it is ok. Otherwise, it would be: the 'joint_4_r' and the combination of 'joint_2_l', 'joint_3_u' and 'joint_5_b' are constrained to orient the flange normal to within +/-15 degrees with respect to the ground plane normal.

gavanderhoorn: Ok. We'll merge it then.

gavanderhoorn (approved):

gavanderhoorn:
Thank you @EricMarcil for the updates and thanks @cjue for the additional comments and contributions :+1:

PR #600, opened 2024-09-24
Switch to JSP GUI in all `test_<model>.launch` files

EricMarcil:
Changed the test_<model>.launch files to use the "joint_state_publisher_gui" instead of the "joint_state_publisher" so that the joint slide controls are displayed with RViz.

### DIFF
--- a/motoman_ar2010_support/package.xml
+++ b/motoman_ar2010_support/package.xml
@@ -47,7 +47,7 @@
 
   <test_depend>roslaunch</test_depend>
 
-  <exec_depend>joint_state_publisher</exec_depend>
+  <exec_depend>joint_state_publisher_gui</exec_depend>
   <exec_depend>motoman_driver</exec_depend>
   <exec_depend>motoman_resources</exec_depend>
   <exec_depend>robot_state_publisher</exec_depend>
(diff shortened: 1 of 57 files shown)


gavanderhoorn (changes requested):
It won't make CI green for Kinetic or Melodic, but if we change this, we must update the `package.xml` to state the dependency on `joint_state_publisher_gui` (instead of the current `joint_state_publisher`).

Like in:

https://github.com/ros-industrial/motoman/blob/433be15f6a27cb9a5715cc3887b1cd15ff33a8cb/motoman_gp110_support/package.xml#L44

EricMarcil:
@gavanderhoorn Updated the package,xml files dependency

gavanderhoorn:
I'll take another look at this one after the PRs contributing more robot support packages have been merged.

gavanderhoorn:
@EricMarcil: I've merged your two commits into one (bcd4d892a6ad458e6e35efa71a08c598e0f92677 was a fixup of the real fix) and corrected some minor issues.

We'll wait for green CI and then merge.

gavanderhoorn (approved):

gavanderhoorn:
Thanks @EricMarcil :+1:

PR #604, opened 2024-10-09
Add GP20HL robot support package

EricMarcil:
Add GP20HL robot support package


motoman_gp20hl_support/meshes/gp20hl/visual/link_6_t.dae
gavanderhoorn, line 1: this file appears to be 'empty' (as in: there's no mesh data defined here).

motoman_gp20hl_support/config/opw_parameters_gp20hl.yaml
gavanderhoorn, line 1: Have these values been verified?

gavanderhoorn (changes requested):
`link_6_t.dae` needs to be fixed.

gavanderhoorn:
@EricMarcil: I just noticed #566.

Is this a copy of that PR?

motoman_gp20hl_support/config/opw_parameters_gp20hl.yaml
EricMarcil: No, there are based on the definition from [opw_kinematics_tutorial.rst](https://github.com/moveit/moveit_tutorials/blob/master/doc/opw_kinematics/opw_kinematics_tutorial.rst)

And reviewing them, the a2 value should be negative.

It is untested. We could remove if you prefer.

gavanderhoorn: Adding OPW parameters is a good thing, but we should make sure the values are correct.

EricMarcil:
No, it

> @EricMarcil: I just noticed #566.
> 
> Is this a copy of that PR?

No, it is a new one that I made. I must have missed the #566 that was made by [acbuynak](https://github.com/acbuynak).
I used the MotoSim model to generate the visual meshes in .dae which keeps the colors of the orignial model. So, I would prefer to keep this one.

motoman_gp20hl_support/config/opw_parameters_gp20hl.yaml
gavanderhoorn: > And reviewing them, the a2 value should be negative.

Fixed in b195458539fdd8353d03047547e8beca9328dc41.

Clarified status in 99c26e1fd6161027bef64c0c39cf4810ded4cd38.

motoman_gp20hl_support/meshes/gp20hl/visual/link_6_t.dae
EricMarcil: Oops, the model must not have been selected when I exported it and I didn't noticed. 
I just committed an update of the model.

gavanderhoorn (approved):

gavanderhoorn:
Thanks @EricMarcil :+1:

PR #606, opened 2025-01-24
GP8L robot support package

EricMarcil:
Add the support package for the Yaskawa GP8L robot


motoman_gp8l_support/config/opw_parameters_gp8l.yaml
gavanderhoorn, line 1: Have these values been verified?

EricMarcil: No, there are based on the definition from [opw_kinematics_tutorial.rst](https://github.com/moveit/moveit_tutorials/blob/master/doc/opw_kinematics/opw_kinematics_tutorial.rst)
But untested. We could remove if you prefer.

gavanderhoorn: Adding OPW parameters is a good thing, but we should make sure the values are correct.

gavanderhoorn: Clarified their status in b91d6897f65bb37575c6187925328c29bc0527e3

cjue:
The OPW config matches the URDF, I could use it with MoveIt to move the robot.

gavanderhoorn:
> The OPW config matches the URDF, I could use it with MoveIt to move the robot.

I've effectively 'removed' b91d6897f65bb37575c6187925328c29bc0527e3 again.

Thanks for the confirmation @cjue.

gavanderhoorn (approved):
Waiting on CI, but approving already.

gavanderhoorn:
Thanks @EricMarcil :+1:

PR #607, opened 2025-01-28
GP165R robot support package

EricMarcil:
Add the support package for the Yaskawa GP165R robot


motoman_gp165r_support/package.xml
```diff
@@ -0,0 +1,56 @@
...
+    <p>ROS-Industrial support for the Motoman GP165R (and variants).</p>
+    <p>
+      This package contains configuration data, 3D models and launch files
+      for Motoman GP165R series of manipulators.
+    </p>
+    <p>
+      <b>Specifications</b>
+    </p>
+    <ul>
+      <li>GP165R - Default</li>
```
gavanderhoorn: From *Table 5-1: Basic Specifications (YR-1-06VR165-A00)* in *186044-1CD*, it seems this support package models the *Not-equipped* version of the 165R:

image

There doesn't appear to be a *Default* version of that robot.

```suggestion
      <li>GP165R - Not equipped</li>
```

If that's not correct, we should at least mention which column *Table 5-1* was used for the joint limits.

gavanderhoorn (changes requested):

gavanderhoorn:
Other comment: this model appears to have a balancer unit. As we're not modelling the actual motion of that, I would suggest removing the piston.

When the arm rotates back, the piston starts intersecting with the arm. This can make motion planners consider that a collision -- unless it has been disabled on purpose.

Just removing the piston would avoid that completely I believe.

motoman_gp165r_support/package.xml
EricMarcil: @gavanderhoorn That a good point. Maybe I can add gp165r-equiped.xacro and indicate that people can swap it out if they have the equiped version.

gavanderhoorn: Hm. We could add another file. It's another variant, just like the HC10 and HC20 variants essentially.

Might be the most user-friendly thing to do. Otherwise users would have to change these limits themselves.

EricMarcil:
> Other comment: this model appears to have a balancer unit. As we're not modelling the actual motion of that, I would suggest removing the piston.
> 
> When the arm rotates back, the piston starts intersecting with the arm. This can make motion planners consider that a collision -- unless it has been disabled on purpose.
> 
> Just removing the piston would avoid that completely I believe.

@gavanderhoorn I looked at removing just the piston, it doesn't change much, because we then use a convex hull on the model to create the collision model which will collide with the L-axis anyway.  

image

The orange outline is the model without the piston, so not much of a difference.

motoman_gp165r_support/package.xml
EricMarcil: Added a new commit adding information about the 'Equipped' version.  Added a gp165R_macro_equipped.xacro but did not doubled up all the launch and test files. Instead a put comments to switch the file reference in the gp165r.xacro.

gavanderhoorn:
Thanks for checking.

I agree about the convex hull, but not all planning tools use it.

Having the 'detailed' mesh also not intersecting with the arm covers all cases and doesn't really reduce visual fidelity.

EricMarcil:
@gavanderhoorn I've updated the link_1_s and removed the piston as per your request.

motoman_gp165r_support/package.xml
gavanderhoorn: I've gone ahead and made the *equipped* version a proper variant.

That led to some duplication of files, but I believe that's better UX than users having to edit files (which would be impossible if/when we actually release these packages through the buildfarm).

motoman_gp165r_support/urdf/readme.txt
```diff
@@ -0,0 +1,8 @@
+This model comes in two versions 'Not-equipped' and 'Equipped'. The 
+equipped version has reduced motion range to account for restrictions
+from extra hardware and cables routed outside of the robot.
+
+Also note that the 'Equipped' version has a reduce payload of 150 kg
+and lower allowable moment and inertia for the wrist joints. Please
```
gavanderhoorn:
```suggestion
and lower allowable moment of inertia for the wrist joints. Please 
```

gavanderhoorn (approved):
Waiting for green CI, but approving already.

gavanderhoorn:
Thanks @EricMarcil :+1:

EricMarcil:
@gavanderhoorn Thanks for all the recent reviews and merge!

PR #608, opened 2025-01-30
Add GP25-12 model to motoman_gp25_support

EricMarcil:
Added the Yaskawa Motoman GP25-12 robot model to the motoman_gp25_support as a variant. Updated the GP25 model and made small corrections.


gavanderhoorn:
@EricMarcil: I've added a couple of commits fixing some minor issues.

> Updated the GP25 model and made small corrections.

could you please use multiple commits when changing properties of other models in a PR?

That way we can keep things logically separated and make it easier to review (and revert, if necessary).

motoman_gp25_support/config/opw_parameters_gp25.yaml
gavanderhoorn, line 1: Have these values been verified?

(same for the other OPW `.yaml` below)

motoman_gp25_support/launch/test_gp25_12.launch
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" ?>
+<launch>
+  <include file="$(find motoman_gp25_support)/launch/load_gp25_12.launch" />
+  <node name="joint_state_publisher_gui" pkg="joint_state_publisher_gui" type="joint_state_publisher_gui" >
+    <param name="use_gui" value="true" />
+  </node>
```
gavanderhoorn:
```suggestion
  <node name="joint_state_publisher_gui" pkg="joint_state_publisher_gui" type="joint_state_publisher_gui"/>
```

`use_gui` is no longer needed.

motoman_gp25_support/config/opw_parameters_gp25.yaml
EricMarcil: No, there are based on the definition from [opw_kinematics_tutorial.rst](https://github.com/moveit/moveit_tutorials/blob/master/doc/opw_kinematics/opw_kinematics_tutorial.rst)
But untested. We could remove if you prefer.

gavanderhoorn: Adding OPW parameters is a good thing, but we should make sure the values are correct.

gavanderhoorn: Clarified status in fd61175bcec832b2117f7554c48426902a55d64f

gavanderhoorn:
@EricMarcil: please avoid adding merges with `kinetic-devel`. I know Github makes it seem things are 'out of sync', but that doesn't block this PR from getting merged.

Adding these merge commits will only make history more complex, by adding all sorts of additional merge relationships between `kinetic-devel` and this PR's branch.

I'll see if I can remove that commit. If not, I'll ask you to do it.

---

Edit: as expected I can't force-push to your branch.

Could you check you have no local changes staged/uncommitted and run the following in a terminal:

```shell
git checkout GP25-12
git reset --hard HEAD~1

# 4c17f11 should have 'disappeared' here

git push origin GP25-12 --force-with-lease
```

this should update the PR and remove the unnecessary merge commit.

Please also do this for #604, #606 and #607 (make sure to use the appropriate branch names for those PRs of course).

EricMarcil:
@gavanderhoorn : Done for this branch. Did it have the result you wanted?
I was trying to make sure that my local branch had the commits that you had added, and the web interface was showing differences with a Sync button that I used.

gavanderhoorn:
Yep, perfect. Thanks.

EricMarcil:
Ok, https://github.com/ros-industrial/motoman/pull/604, https://github.com/ros-industrial/motoman/pull/606 and https://github.com/ros-industrial/motoman/pull/607 should be done now.

gavanderhoorn (approved):

gavanderhoorn:
Thanks @EricMarcil :+1:

PR #628, opened 2025-03-20
driver: Fixes `JointTrajPtFullEx::init` method

briancairl:
`setMultiJointTrajPointData(...)` was being passed the class member, not the argument, so this call had no effect when adding traj points.

Workaround is to just call `setMultiJointTrajPointData(...)` outside of `init(...)`


gavanderhoorn (approved):
Can't test this with real hw right now, but this seems like it would be the correct change to make.

gavanderhoorn:
Thanks for the PR @briancairl :+1:

briancairl:
@gavanderhoorn np. Fwiw, tested on my local setup 😊

PR #631, opened 2025-04-01
add opw config for GP180

cjue:


gavanderhoorn:
Can you confirm you've tested this with real hw / moveit?

cjue:
> Can you confirm you've tested this with real hw / moveit?

Yes, this config has been tested with Moveit and real hw.

cjue:
it's also identical to the one for the GP180_120 except for c3.

gavanderhoorn:
> it's also identical to the one for the GP180_120 except for c2.

I assume you mean `c3`?

but thanks for confirming.

gavanderhoorn (approved):

gavanderhoorn:
Thanks @cjue :+1:

cjue:
> > it's also identical to the one for the GP180_120 except for c2.
> 
> I assume you mean `c3`?
> 
> but thanks for confirming.

yes, that's  an unfortunate typo.

PR #632, opened 2025-04-15
Store MoveIt configurations in sub directory

gavanderhoorn:
They're essentially unmaintained, but we don't need to delete them.

Catkin will still be able to find them.

No functional changes.


gavanderhoorn:
Note: this is the first PR going into `noetic-devel`.

cjue:
Looks good :+1:

marip8 (approved):
Looks good to me

PR #633, opened 2025-04-15
Remove `.urdf`s from packages (`.xacro`s should be used instead)

gavanderhoorn:
We're not keeping these `.urdf`s up-to-date, so changes to the `.xacro` might not reach users of these `.urdf`s.

Converting `.xacro` to `.urdf` is relatively trivial, and if really needed, can be done on request.

Note: this branches from `move_moveit_pkgs`, so includes those commits. Ignore those.

The main commit in this PR is 11a1c30040a48e9f3daf933303d2e3c64b9da6b7 (diff [here](https://github.com/ros-industrial/motoman/pull/633/commits/11a1c30040a48e9f3daf933303d2e3c64b9da6b7)).

I'll rebase after #632 is merged.


cjue:
good idea :+1:

marip8 (approved):
Looks good to me

PR #634, opened 2025-04-15
ci: switch to 24.04 runner

gavanderhoorn:
As per title.

The `20.04` runner is being deprecated, following the EOL of Focal.


marip8 (approved):
You could also update this to `ubuntu-latest` since ICI runs in a docker image. Then we may not need to update this again for a while

gavanderhoorn:
That's a good point.

I don't like tracking moving targets, but I did bump to `24.04`.

gavanderhoorn:
Thanks @marip8 :+1:

PR #635, opened 2025-04-19
Updates to buildscripts and package manifests

gavanderhoorn:
Quite a few changes here (but nothing functional):

- updates to maintainers in all packages
- added missing dependencies
- clarified licenses in all packages (changed to SPDX identifiers)
- synced version nrs everywhere

and in general made `catkin_lint` much happier (although some things had to be ignored as it is sometimes a bit opinionated), so much so we can now actually add it as a job to the CI config.


gavanderhoorn:
@EricMarcil: could you check you'd be OK with the updated maintainer entry I used? Instead of your work email, I've changed it all to `techsupport@`. Noticed you started using that in recent support packages, so thought I'd use that everywhere.

EricMarcil (approved):
Yes, it is better to make reference to techsupport e-mail as my e-mail could eventually be no longer be valid in the eventuality that I leave Yaskawa.
I also reviewed the updates on older models like MH12, which are good.
I'll let @marip8 review the changes that or more specific to the ROS side of the driver.
Thanks for all the work.

marip8 (approved):
Looks good to me

gavanderhoorn:
Thanks for the reviews :+1:

PR #639, opened 2025-04-21
Tabs to spaces in `.xacro`, `.launch` and `roslaunch` tests

gavanderhoorn:
As per title.

No functional changes.

We had a mix of spaces and/or tabs used for indentation in these files, which was really annoying, especially since we also have mixed 2 and 4 space indents.

Used separate commits to make reviewing slightly more convenient (easier to compare similar changes to similar files).


gavanderhoorn:
This will likely conflict with #636. I'll fix and rebase once that gets merged.

marip8 (approved):

PR #414, opened 2021-06-24
Migrate CSDA10F support and config packages from experimental

Danfoa:
This PR addresses the migration of the CSDA10F support and config packages from the `motoman_experimental` repo, which were originally introduced in that repository in this [PR](https://github.com/ros-industrial/motoman_experimental/pull/43).

As mentioned in that PR, the files have already matured and have been used by several teams that have forked the [invite-robotics](https://github.com/Danfoa/invite-robotics) repository  (which is the origin of these files), without any reported issue. 

I personally worked for almost two years with the files with the CSDA10F and all the errors I detected were corrected in the files that are now in the Motoman experimental repo.

PR #646, opened 2025-07-04
Migrate CSDA10F package from experimental repository (integrate into SDA10F support pkg)

gavanderhoorn:
As per title.

This migrates all packages related to the CSDA10F from `motoman_experimental` to `motoman`, with the commit history that touches those packages.

This obviously leads to duplicate commits (as many touch multiple packages) and changed hashes, but there is no way to maintain the original commits while also importing them into this repository. At least this way the full history is present here as well. PR and issue references have been updated where necessary to still refer to `motoman_experimental`.

No functional changes have been made to any of the packages. This is essentially a straight import of the artefacts from `motoman_experimental` with some (minor) updates to the package manifests (such as references to `motoman_experimental` and the wiki).


gavanderhoorn:
This is a replacement for #414, which didn't include the commit history. That PR also migrated the MoveIt configuration, but as we're not hosting those here anymore, I decided to not include it in this PR.

@EricMarcil: could you perhaps take a look at the joint limits? The original values ([here](https://github.com/ros-industrial/motoman_experimental/blob/04b183e887beaff76c22713a5525b0753796fac1/motoman_csda10f_support/urdf/arm_macro.xacro#L119-L167) and [here](https://github.com/ros-industrial/motoman_experimental/blob/04b183e887beaff76c22713a5525b0753796fac1/motoman_csda10f_support/urdf/common_torso_macro.xacro#L60-L74)) didn't seem correct to me for some joints.

Specifying everything in degrees now (using `${radians(..)}`) makes comparing to the documentation a bit easier.

EricMarcil:
The values seem to be fine. But I agree that it might be better to put the reference in degrees using ${radians(..)}.

These are the values from the all.prm for that robot:
image

Both arms are the same except that the S, E, R, T axes direction is reversed.

The base is:
image

gavanderhoorn:
I've used the max acc/dec torque to populate the `effort` fields, but I noticed there's no data for `B1` @EricMarcil?

EricMarcil:
> I've used the max acc/dec torque to populate the `effort` fields, but I noticed there's no data for `B1` @EricMarcil?

I've double-checked and the values are not available in the data from the MotoSim. There might be valid values from the actual controller all.prm but I don't have access to a FS100 controller.  So I would leave it out for now. 

If someone is actually still using the arm, then they can send us the all.prm from the controller and I'll review the values.

gavanderhoorn:
@EricMarcil / @ted-miller: could you perhaps take a look here as well? As it's 'my' PR I can't approve it myself.

EricMarcil:
I just noticed that we already have the SDA10F package. The CSDA10F is essentially the same except that the B-T joint are slightly different with a 20 mm difference. 
image
The model in the front is the CSDA10F and the on in the back SDA10F.

The SDA10F package structure seems cleaner. The CSDA10F has strange origin xyz values.

gavanderhoorn:
so would your suggestion be to add the `C` variant to the base `SDA10F` support package?

are there no other differences whatsoever?

EricMarcil:
I'll need to review the specifications, there might be some small changes on range and speed. I'll try to look into it, next week.

gavanderhoorn:
Friendly ping @EricMarcil :)

EricMarcil:
The data is almost the same, the S-axis motion range is +/-180 deg instead of +/-170 deg, and the Flange offset is 155 mm instead of 175 mm.

image

image

So, I would recommend merging the two models into the same package.

EricMarcil:
Do we have anyone who requested the CSDA10F? I'm fairly busy recently, so that would be a low priority to me, unless someone needs it.

gavanderhoorn:
> Do we have anyone who requested the CSDA10F? 

No requests per se.

I'm merely migrating the contributions by community members (see [commits](https://github.com/ros-industrial/motoman/pull/646/commits)) from `ros-industrial/motoman_experimental` to this repository.

> The data is almost the same, [..]

what I'll do is add some commits that merge the new package into the existing one.

That way the commit history (and thus contributions) is retained, but we do not end up with (almost) duplicated packages and files.

gavanderhoorn:
@EricMarcil: I've followed your advice and have merged the CSDA10F into the SDA10F support package.

I've kept all of the original commits, but added new ones to move the relevant files to their new location.

In the end I reused the complete structure of `motoman_sda10f`, but updated:

- `S` position limits (`+-170` degrees instead of `+-180` degrees)
- `B`->`T` transform (`175`mm vs `150`mm)

generally this works, but of course the `link_b` mesh is now incorrect:

rviz_screenshot_2025_08_27-12_15_42

could you perhaps provide an updated mesh for that link?

motoman_sda10f_support/package.xml
```diff
@@ -7,18 +7,24 @@
...
-      for Motoman SDA10F manipulators.
+      for Motoman SDA10F manipulators. This currently includes the base
+      model and the CSDA10F variant (ie: clean standard).
     </p>
     <p>
       <b>Specifications</b>
     </p>
     <ul>
       <li>SDA10F - Default</li>
+      <li>CSDA10F (YR-CSDA010F-A10) - Default</li>
```
gavanderhoorn, line 18: added the specific model I looked up the specs for in the documentation.

motoman_sda10f_support/urdf/csda10f_arm_macro.xacro
```diff
@@ -0,0 +1,165 @@
...
+      </collision>
+    </link>
+    <link name="${prefix}tool0" />
+
+    <joint name="${prefix}joint_1_s" type="revolute">
+      <parent link="${parent}"/>
+      <child link="${prefix}link_1_s"/>
+      <origin xyz="0.100 ${reflect*0.265} 0.0" rpy="${radians(90)} 0.0 ${(reflect-1)*radians(90)}"/>
+      <axis xyz="0 0 ${-reflect}" />
+      <limit lower="${radians(-170)}" upper="${radians(170)}" effort="268.36" velocity="${radians(170)}" />
```
gavanderhoorn, line 114: @EricMarcil: could you please verify?

The `ALL.PRM` values you posted (https://github.com/ros-industrial/motoman/pull/646#issuecomment-3046257622) have the velocity limit at `170`deg/s.

`HW1481698` however states this (page 32, section 5.1):

image

the other values seem to match, but this one does not.

motoman_sda10f_support/urdf/csda10f_arm_macro.xacro
```diff
@@ -0,0 +1,165 @@
...
+        <geometry>
+          <mesh filename="package://motoman_sda10f_support/meshes/sda10f/collision/link_r.stl"/>
+        </geometry>
+      </collision>
+    </link>
+    <link name="${prefix}link_6_b">
+      <visual>
+        <origin rpy="0 0 0" xyz="0 0 0"/>
+        <geometry>
+          <mesh filename="package://motoman_sda10f_support/meshes/sda10f/visual/link_b.stl"/>
```
gavanderhoorn, line 83: we'll have to use a mesh specific to the CSDA10F here (same for the `collision` mesh).

EricMarcil:
> could you perhaps provide an updated mesh for that link?

Yes, it is actually the T-axis that is different, not the B-axis

motoman_sda10f_support/package.xml
EricMarcil: That's correct.

motoman_sda10f_support/urdf/csda10f_arm_macro.xacro
EricMarcil: Look again at the referred post above. The range is +/-170 deg and the speed is 130 deg/s. So the values are matching.

motoman_sda10f_support/urdf/csda10f_arm_macro.xacro
EricMarcil: Yes. but it is the link_t.stl that is different.

EricMarcil:
Doesn't look like I can push to you branch. Here are the files:
[csda10f.zip](https://github.com/user-attachments/files/22011619/csda10f.zip)
[csda10f_arm_macro.zip](https://github.com/user-attachments/files/22011666/csda10f_arm_macro.zip)

motoman_sda10f_support/urdf/csda10f_arm_macro.xacro
jimmy-mcelwain: @EricMarcil the B1 axis looks correct, but it does look like the S1 axis has a speed of 170 which differs from the manual, right? 

Edit: I now see that is just the FS100 version, but that confuses me for a different reason

gavanderhoorn: IIUC, your first comment showed the CSDA10F parameters:

image

I'm not entirely sure how to interpret what the manual states though, is the value for `S` missing for FS100?

motoman_sda10f_support/urdf/csda10f_arm_macro.xacro
gavanderhoorn: I just assumed the `B` mesh would be different.

motoman_sda10f_support/urdf/csda10f_arm_macro.xacro
jimmy-mcelwain: [This brochure](https://www.yaskawa.eu.com/Global%20Assets/Downloads/Brochures_Catalogues/Robotics/MOTOMAN_Robots/CSDA10F/Flyer_Robot_CSDA10F_E_05.2017.pdf) from YEU has `170 deg/s` as the `S` speed.

It seems like the torso axis has different names between the `CSDA10F` and `CSDA10D`. `B1` for `CSDA10F` and `S1` for `CSDA10D`. So they are both missing the `S` speed in the manual. Maybe an oversight?
CSDA

EricMarcil: yes, the S-axis speed is missing in the manual. The S1 makes reference to the station 1. On the DX200 version, the torso was configured as a Station. In the FS100 version, it is was changed to Base1 and Base 2 which physically maps to the same torso motor.

gavanderhoorn:
> Doesn't look like I can push to you branch. Here are the files: [csda10f.zip](https://github.com/user-attachments/files/22011619/csda10f.zip) [csda10f_arm_macro.zip](https://github.com/user-attachments/files/22011666/csda10f_arm_macro.zip)

Fixed:

rviz_screenshot_2025_09_02-17_23_00

gavanderhoorn:
It's not the cleanest package, nor have I touched any of the existing files in this support package, but the CSDA10F is now added as a variant of the SDA10F to `motoman_sda10f_support` while preserving provenance and attribution.

EricMarcil (approved):
All good!

EricMarcil:
@gavanderhoorn Thanks for working on this. It is really appreciated. 
I'll let you do the final merge since I'm not sure if you want to squash or not.

gavanderhoorn:
As I wrote earlier, it's not perfect, but let's merge and we can address further issues in follow-up PRs.

Thanks for the review @EricMarcil :+1:

gavanderhoorn:
> I'm not sure if you want to squash or not.

in this case we don't want to squash, as we want to keep the provenance and history clear, as this is (at least partially) a migration from `motoman_experimental`.

PR #643, opened 2025-06-02
Add GP20 (GP25-20) to the motoman_gp25_support package 

EricMarcil:
Adds the GP20 variant to the motoman_gp25_support packages as the model name GP25-20 (which is the configure model type 1-06VX25-20-A00)

PR #647, opened 2025-07-10
Add UP50N-35 model

EricMarcil:
This package contains configuration data, 3D models and launch files for Motoman UP50N-35 manipulators.
This model is for an NX100 controller which is does not support MotoPlus and cannot be used with the motoman_driver.


motoman_up50n_support/package.xml
```diff
@@ -0,0 +1,59 @@
...
+  <url type="bugtracker">https://github.com/ros-industrial/motoman/issues</url>
+  <url type="repository">https://github.com/ros-industrial/motoman</url>
+
+  <buildtool_depend>catkin</buildtool_depend>
+
+  <test_depend>roslaunch</test_depend>
+
+  <exec_depend>industrial_robot_client</exec_depend>
+  <exec_depend>joint_state_publisher_gui</exec_depend>
+  <exec_depend>motoman_driver</exec_depend>
```
gavanderhoorn:
```suggestion
```

the driver does not support it, so let's remove it from the dependencies.

motoman_up50n_support/test/launch_test.xml
```diff
@@ -0,0 +1,30 @@
...
+  </group>
+
+  <group ns="robot_state_visualize_up50n_35">
+    <group ns="dx100" >
+      <include file="$(find motoman_up50n_support)/launch/robot_state_visualize_up50n_35.launch">
+        <arg name="robot_ip"   value="$(arg req_arg)" />
+        <arg name="controller" value="$(arg dx100)"/>
+      </include>
+    </group>
+  </group>
```
gavanderhoorn:
```suggestion
```

driver is not supported here, so no need to try and test those files.

motoman_up50n_support/urdf/up50n_35_macro.xacro
```diff
@@ -0,0 +1,170 @@
...
+    <link name="${prefix}base_link">
+        <visual>
+            <geometry>
+                <mesh filename="package://motoman_up50n_support/meshes/up50n_35/visual/base_link.dae"/>
+            </geometry>
+            <xacro:material_yaskawa_blue/>
+        </visual>
+        <collision>
+            <geometry>
+                <mesh filename="package://motoman_up50n_support/meshes/up50n_35/visual/base_link.dae"/>
```
gavanderhoorn: should we not have collision models here instead?

motoman_up50n_support/launch/load_up50n_35.launch
```diff
@@ -0,0 +1,3 @@
+<launch>
```
gavanderhoorn, line 2:
```suggestion
<?xml version="1.0" ?>
<launch>
```

gavanderhoorn (commented):
first round of comments

motoman_up50n_support/config/joint_names_up50n_35.yaml
gavanderhoorn, line 1: I would also remove this file, as there is no driver to use it.

motoman_up50n_support/urdf/up50n_35_macro.xacro
EricMarcil: The model is composed of simple geometry and are very small, so I figured that it could be used for collision as is. But I could generate collision models if you think it is better.

gavanderhoorn: It's probably almost superstition at this point, but I've experienced problems with motion planners not loading Collada meshes properly and then ending up with 'empty' collision models (and EEFs crashing into walls as a result).

From that time on I try to always use STL for collision models. I'll take a look at the Collada meshes.

EricMarcil: Ok, I can quickly export to stl

EricMarcil:
@gavanderhoorn Updated suggestion and add collision model. Thanks for the review.

motoman_up50n_support/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+find_package(catkin REQUIRED)
+
+catkin_package()
+
+if (CATKIN_ENABLE_TESTING)
+  find_package(roslaunch REQUIRED)
+  roslaunch_add_file_check(test/launch_test.xml)
+endif()
+
+install(DIRECTORY config launch meshes urdf
```
gavanderhoorn:
```suggestion
install(DIRECTORY launch meshes urdf
```

motoman_up50n_support/package.xml
```diff
@@ -0,0 +1,58 @@
...
+  <description>
+    <p>ROS-Industrial support for the Motoman UP50N (and variants).</p>
+    <p>
+      This package contains configuration data, 3D models and launch files
+      for Motoman UP50N-35 manipulators.
+    <p>
+      This model is for an NX100 controller which is does not support
+      MotoPlus and cannot be used with the motoman_driver.
+    </p>
+    </p>
```
gavanderhoorn:
```suggestion
    <p>
      This package contains configuration data, 3D models and launch files
      for Motoman UP50N-35 manipulators.
    </p>
    <p>
      This model is for an NX100 controller which is does not support
      MotoPlus and cannot be used with the motoman_driver.
    </p>
```

motoman_up50n_support/package.xml
```diff
@@ -0,0 +1,58 @@
...
+    <p>
+      <b>Specifications</b>
+    </p>
+    <ul>
+      <li>UP50N-35 - Default</li>
```
gavanderhoorn: would this have a more exact model nr, like we started listing in #643?

motoman_up50n_support/test/launch_test.xml
```diff
@@ -0,0 +1,12 @@
+<launch>
+  <arg name="req_arg" value="default"/>
+  <arg name="dx100" value="dx100"/>
+
```
gavanderhoorn: these are also no longer needed

```suggestion
```

gavanderhoorn (approved):
Motor housings are a bit dark (ie: almost black) but other than that this seems OK.

gavanderhoorn:
I'll squash-merge this, as we don't need the fixups in the commit history.

gavanderhoorn:
Thanks @EricMarcil :+1: 

@matthias-mayr: have you used the model in the meantime?

PR #654, opened 2026-05-21
Add GP400 support package

EricMarcil:
Add GP400 support package


jimmy-mcelwain:
I talked about this with @EricMarcil offline, but I believe that this is the first URDF on this repo that has a closed loop chain as part of the main kinematic structure. I am interested in if this is the best way to achieve that with URDF, since I know it is meant for serial kinematic chains. 

Once this gets merged, I'll port it to [motoman_ros2_support_packages](https://github.com/Yaskawa-Global/motoman_ros2_support_package)

